Issue #264, opened 2015-11-07
Arbitrary version comparison


benwbrum:
The University of Texas needs the ability to compare a page version not only with the previous version, but with any arbitrary version.

PR #265, opened 2015-11-07
Implemented #264 by adding a form.  Further UI work needed

benwbrum:
Currently the page version screen lets users view a version and see both the state of the page at that version and the differences with the previous version.  The University of Texas wants to be able to see the differences between one version with any other version, not just the previous one.

I've implemented the base functionality, but we need to make this more visually appealing.  Ideally, the select list for the comparison version would appear at the header of the `diff` view, where I have added the text *Form goes here*.  This would serve the two purposes of 1) showing the user which version was being compared, and 2) allowing them to select a different version to compare against.

So the final behavior should be:
* Users may select from the existing list on the left column and see that version compared against the previous version, as before.
* Users may select a different version from the existing list on the left column and see that version compared against is previous version, as before.
* Users may choose a different comparison version from a select list, which will compare the currently selected version against the comparison version of the page.

I'm open to other ideas on how to accomplish this in a visually appealing manner.

### DIFF
--- a/app/assets/stylesheets/components/shared.scss
+++ b/app/assets/stylesheets/components/shared.scss
@@ -962,11 +962,24 @@ button, .button, .dropdown dd a {
   table-layout: fixed;
   border-collapse: separate;
   th {
+    padding: 10px 0;
     line-height: 30px;
     color: $bodyFgFaded;
-    padding: 0 0 5px 10px;
-    vertical-align: bottom;
-    &:first-child { width: 25%; }
+    &:first-child {
+      width: 25%;
+      padding-left: 10px;
+    }
+    form {
+      position: relative;
+      padding-left: 115px;
+      label {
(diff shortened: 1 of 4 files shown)


kolking:
@benwbrum I was thinking about removing the versions list at the left and making both version columns to have that select list at the top, but then just realized that the behavior you've initially proposed is better. Most likely, the very common use case here is just to compare a version with the previous one to see what has been changed at this iteration, so they need to click only once (in the list at the left), and then, if they really need to compare with another version (which extends the common use case), they use the version select at the top.

benwbrum:
That is also my conclusion -- showing the selected version and comparing it with the previous version will serve 90% of use cases, and all of the casual use cases.  This request will be deliberate, done by sophisticated users (i.e. professors grading student contributions).

PR #315, opened 2016-05-16
Ui design

benwbrum:
This should address most of the uploaded issues reported by @atomrab, fixing the stuck-in-'processing' problem, fixing the empty-work-with-the-name-of-the-file problem, and adding more logging so we can diagnose the pages-from-X-appear-in-Y problem once it's deployed.

Can you review and merge, @saracarl?  I only opened a pull request on master, so we'll need to merge into the fromthepage.com branch manually.

### DIFF
--- a/lib/tasks/ingestor.rake
+++ b/lib/tasks/ingestor.rake
@@ -68,7 +68,7 @@ namespace :fromthepage do
     print "unzip_tree(#{temp_dir})\n"
     ls = Dir.glob(File.join(temp_dir, "*"))
     ls.each do |path|
-      print "unzip_tree handling #{path}\n"
+      print "\tunzip_tree considering #{path}\n"
       if Dir.exist? path
         print "Found directory #{path}\n"
         unzip_tree(path) #recurse
@@ -77,7 +77,7 @@ namespace :fromthepage do
           print "Found zipfile #{path}\n"
           #unzip and recur
           destination = File.join(File.dirname(path), File.basename(path).sub(File.extname(path),''))
-          print "Calling unzip_file(#{path}, #{destination})"
+          print "Calling unzip_file(#{path}, #{destination})\n"
           ImageHelper.unzip_file(path, destination)
           unzip_tree(destination)  # recurse
         end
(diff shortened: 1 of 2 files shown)


lib/tasks/ingestor.rake
```diff
@@ -124,38 +124,50 @@ namespace :fromthepage do
...
-      print "ingest_tree handling #{path})\n"
+      print "ingest_tree considering #{path})\n"
       if Dir.exist? path
         print "Found directory #{path}\n"
         ingest_tree(document_upload, path) #recurse
       end
     end    
     # now process this directory if it contains image files
-    image_files = Dir.glob(File.join(temp_dir, "*{"+IMAGE_FILE_EXTENSIONS.join(',')+"}"))
+    image_files = Dir.glob(File.join(temp_dir, "*.{"+IMAGE_FILE_EXTENSIONS.join(',')+"}"))
```
benwbrum: This was the source of the empty file.  Because we were looking for "image files" in a directory before processing it, and because we were defining "image files" as files that end with `jpg`, and because the zip file UT uploaded happened to have a name ending with `jpg`, we got a false positive on the root directory for the upload.

I think adding the period should fix this.

lib/tasks/ingestor.rake
```diff
@@ -124,38 +124,50 @@ namespace :fromthepage do
...
+    print "\tconvert_to_work collection = #{document_upload.collection.title}\n"
+    print "\tconvert_to_work title = #{File.basename(path).ljust(3,'.')}\n"
+#    binding.pry if path == "/tmp/fromthepage_uploads/16/terrell-papers-jpg"
     User.current_user=document_upload.user
     
     work = Work.new
     work.owner = document_upload.user
     work.collection = document_upload.collection
-    work.title = File.basename(path)
+    work.title = File.basename(path).ljust(3,'.')
```
benwbrum: I believe that this was the source of the "processing" bug.  At the very end of the collection ingestion, the depth-first-search considers the root directory.  Because of the ends-with-jpg bug, it was trying to ingest the root temp directory, `/tmp/fromthepage_uploads/16`.  It turns out that we require works to have titles of at least three characters, so validation was failing.

Adding `ljust` pads out a work title with periods until it's at least three characters.  That fixed the `processing` bug, since `ingest_tree` actually completes.  Now that we've fixed the ends-in-jpg bug as well, this should not be needed, but I didn't see any reason to remove it.

PR #389, opened 2016-09-24
Contributor mailer

saracarl:


saracarl:
Here's the question -- when I want to set this up for a particular customer/collection owner, how am I going to do that?  This mailer uses hardcoded ADMIN_EMAILS for the "to".  I'd like to either set it to the collection owner's email address OR pass it in via the rake task to this mailer.

app/mailers/admin_mailer.rb
```diff
@@ -0,0 +1,26 @@
...
+  before_filter :add_inline_attachments!
+
+  default from: "FromThePage <support@fromthepage.com>"
+  layout "admin_mailer"
+  
+  def contributor_stats(collection_id, start_date, end_date)
+
+    new_contributors(collection_id, start_date, end_date)
+
+    mail from: SENDING_EMAIL_ADDRESS, to: ADMIN_EMAILS, subject: "New Transcription Information "
```
saracarl: hard coded ADMIN_EMAILS here.  Can we pull the owner from the collection_id?

lib/tasks/contributor_stats.rake
```diff
@@ -0,0 +1,10 @@
+namespace :fromthepage do
+  desc "transcription stats from the previous 24 hours"
+  task :contributor_stats, [:collection_id] => :environment do |t,args|
```
saracarl: How hard would it be for the rake task to take one (or more) email addresses and use them for the "to" in the mailer?  That's most flexible.

app/assets/stylesheets/datepicker/demo.css
```diff
@@ -0,0 +1,234 @@
+body
```
saracarl: Is this css file actually being used anywhere?  If not, let's not include it.

lib/tasks/contributor_stats.rake
tbl73: It's very easy to set it to take one email; I think the "or more" isn't bad, but I'm investigating.

app/assets/stylesheets/datepicker/demo.css
tbl73: It doesn't look like it is, so I just removed it from my environment.

app/mailers/admin_mailer.rb
tbl73: You probably can - do you want to send it to the collection owner or pass it in the rake task?

Issue #403, opened 2016-09-29
creating categories during subject-category assignment


saracarl:
When you create new subjects, they have to be assigned to a category.  If you have no categories, you can't assign them.  So a brand new collection owner would get confused here -- we need to ability to create a category directly from the dropdown  OR we should show the "uncategorized" category in the dropdown from the get go.

http://97.107.141.53/transcribe/assign_categories?page_id=1

benwbrum:
When we create a new collection we should automatically create 2 categories:
People
Places

The export code looks for these categories and treats them specially, and
they're good defaults.

Ben

tbl73:
Okay, I set these up to create automatically.  I had to change the validation scope on the category, so that it looks at the collection_id as well as the parent_id.  Pushed up in the categories branch.

PR #418, opened 2016-10-06
issue #403 - categories branch

tbl73:
Add default categories of People and Places when the collection is created.


saracarl:
Verified fix.

Issue #348, opened 2016-08-18
500 Error on Start A Project


eturnator:
This is for the UT installation: Start A Project is giving a 500 error upon click.

saracarl:
We don't think this is happening anymore.  @atomrab -- is this still an issue?

atomrab:
@saracarl -- yes, this does seem to be an issue still. If I'm logged in and using the owner dashboard, I see two tabs: "Your works", with my documents, which is the default; and "Start a Project". If I click on the "Start a Project" tab, I get a page with the 500 message. The same happens if I click on the link to "Start a Project" that appears under each of the individual documents listed in "Your works".

saracarl:
@atomrab Just sent you and your team an email about a new version of FtP.  I'd like to test this issue against that code and if it still happens we will address it. I suspect it's caused by the "multiple owners" not being fully fleshed out, and @tbl73 is working another fix (not included in that release, unfortunately) around that that I have high hopes will finally put all of those related bugs to bed.

atomrab:
Perry just installed the new version (looks good!), and this is still happening.

saracarl:
Just looked at the logs, and this seems to be the problem:

I, [2016-10-13T09:49:55.840702 #29272]  INFO -- : Completed 500 Internal Server Error in 429ms
F, [2016-10-13T09:49:55.847039 #29272] FATAL -- : 
ActionView::Template::Error (undefined method `encoding' for nil:NilClass):
    33:       ul.omeka-sites
    34:         -@sc_collections.each do |sc_collection|
    35:           li.omeka-site
    36:             h5.omeka-site_title =link_to sc_collection.label, {:controller => 'sc_collections', :action => 'explore', :at_id => CGI::escape(sc_collection.at_id)}
    37:     =form_tag({ controller: 'sc_collections', action: 'explore_manifest' }) do |f|
    38:       table.form
    39:           td
  app/views/dashboard/startproject.html.slim:36:in `block in _app_views_dashboard_startproject_html_slim__1581545557777498371_62563380'
  app/views/dashboard/startproject.html.slim:34:in `_app_views_dashboard_startproject_html_slim__1581545557777498371_62563380'

We'll look into it.

saracarl:
We think this is a problem if you have never imported a IIIF collection (which UT hasn't/won't...).  We should be able to reproduce by using our old export of UT's database (which has no IIIF collections/works).

Relevant code:
dashboard_controller.rb line 47 & 48:
[...]

PR #445, opened 2016-10-14
issue #348

tbl73:
Added migration to delete bad ScCollection data to resolve issue #348 on UT's server.


saracarl:
tested and reviewed.

Issue #422, opened 2016-10-07
change "editor" and "transcriber" references in the UI to "contributor"


saracarl:
We've struggled with what to call someone who transcribes a work.  A "transcriber" is accurate, but discounts the level of work many transcribers are actually doing (some transcribers are actually professional editors working on their own projects).  Using "editor" for a volunteer/undergrad transcriber runs the risk of offending professional editors who bring 

We need to go through the UI and document everywhere we are using those terms, think about whether they could be successfully renamed "contributor", and change them if it's appropriate.  If it's not appropriate or questionable, we should discuss it here.  (Possible hack is to just search for those terms in the views.)

I also just noticed that's what github calls people "involved in an issue" -- coder, reviewer, tester, weigher-in -- are all called "contributors".

saracarl:
Change the following from editor to collaborator:
1)  "Editor Dashboard" --> "Collaborator Dashboard"


work settings: http://fromthepage.com/work/edit?work_id=192
2) Allow editors to edit page titles

3) Restrict editors  

4)  Allowed editors

5)  "Add new scribe" ->  "Add new collaborator"

6)  "No allowed scribes selected" -> "No allowed collaborators selected"

FAQ:

What is an editor? The editor and transcriber roles are currently the same.

What is a transcriber? Transcribers transcribe or translate pages in a work/collection. They can also edit page titles if the work owner has given them permission to do so.

Who can transcribe a work? Anyone with a FromThePage account can transcribe any work, unless editors have been restricted by the owner of the work.

7)  get rid of "What is a transcriber?..."

8)  change "what is an editor"? to "What is a collaborator?" and add the above description from transcriber.

tbl73:
What was the conclusion on the collection overview and collection statistics page with top transcribers/top editors?  Did we settle on Top Transcribers and Top Collaborators rather than Top Transcribers and Top Editors?

tbl73:
Changing editors to collaborators in those instances as well.

saracarl:
On Collection Overview, get rid of the whole "top transcribers", "top editors", and "top indexers".

Can we change the bottom of the Collection Statistics page to have a header:
Collaborators
And then 3 columns labelled:
Transcribing
Editing
Indexing

tbl73:
Just pushed those changes up in wording branch.

Issue #410, opened 2016-10-03
"mentions" is unclear on statistics tab


saracarl:
What does "mentions" mean in the statistics tab?  A work with 2 pages transcribed by 2 users, 2 subjects, and a couple of revisions has 2 "mentions" but I don't understand what they actually are.

Once we figure out what they are, we need to either explain them or take them away.

@suwiding fyi.

tbl73:
As best I can tell, "mentions" are created when someone links a subject in a transcription.   It's not clear to me what the difference is between "subjects" and "mentions" is, though.  From a little testing, they appear to track together.

saracarl:
So what if you link the same subject multiple times in the work?  Maybe the
mention count goes up?  (It's very common for the same person to appear
multiple times in a work, and get linked to the article about that person,
so this would make sense to me.)

tbl73:
They do go up, so that's the difference.  I think we might need a different word than mentions, though, because I didn't understand that either.

tbl73:
Oh, and while we're talking about this, it took me a while to realize that "comments" were the same thing as notes, though I don't know if anyone other than me is confused by that.

tbl73:
Change "mentions" to "references", and change "comments" to "notes".

saracarl:
reviewed.

Issue #406, opened 2016-10-02
faq for IA wrong


saracarl:
Get rid of this bit of the Internet archive import doc:   an OAI (Open Archives Initiative)

saracarl:
reviewed.

PR #449, opened 2016-10-17
Wording, issue #406, #410, #422

tbl73:
Multiple wording changes to the UI.


saracarl:
Needs a bit more work on #422

saracarl:
I like the "transcribing" "editing" and "indexing" a lot better here.

Issue #447, opened 2016-10-17
Status of page import from internet archive using OCR


tbl73:
When you import a work from the internet archive and use OCR, the work shows up as transcribed in the collection view, but not in the page view (because of the revisions).  To make the workflow less confusing, we need to change the "transcribed" wording for OCR-imports to "corrected" - if you look at a page, rather than saying the page hasn't been transcribed yet, say the page hasn't been corrected yet.

tbl73:
I changed this wording, but I just realized that this also means that the first time a person does a correction on the OCR text, it's going to be considered a transcription deed, rather than edit deed.  Is that what we want?  I could change the import so that OCR text ends up at revision 2 (so that deeds are edit deeds).  Sara and I had talked about that this morning, and the only thing that's problematic with that is that you can't really see what needs to be corrected at a glance, but I could change the UI so that the new pages show up as needing correction.  Just trying to figure out what the right workflow is here.  @benwbrum , @saracarl , what do you think?

saracarl:
If we change the import to set it to revision 2, I have 2 concerns:
1)  What does the revision history look like?  (no real changes between 1 and 2?)
2)  How do you tell that a page has or hasn't been corrected yet?

#2 is the most important bit here -- and maybe we can't accomplish it with this change, but anything we can do that implies "this page here, it needs correcting" (and vice versa) would be good.

saracarl:
Ah, your note says:  " the UI could be set up to show you the pages that haven't yet been corrected similar to the way new transcriptions show up. "  Should that actually read "couldn't"?  

And if it couldn't, then we should not set it to revision 2.

benwbrum:
Should we be creating a new "Corrected" deed or "OCR Corrected" deed?  It
seems like this ought to be a totally separate process.

tbl73:
I was actually thinking that I might be able to set it in the UI to show up saying "This page needs to be corrected", regardless of the revision.  When you use OCR in the import, the page status is set to "raw_ocr", and that doesn't change until someone actually saves a change to the page (i.e., it's true for revision 0 and revision 1 when those are done automatically).  I could use that status to display "this page needs to be corrected", much the way we do "this page needs to be transcribed", so that the user could tell by looking at the pages whether they have been corrected or not.

However, it's true that the revision history looks kind of strange.  I think Ben might be right about needing to have another kind of deed for this - it does seem like a separate sort of process.

saracarl:
I love the first idea here.

[...]

Issue #411, opened 2016-10-04
"Pages transcribed" in statistics tab off


saracarl:
Not sure what this is currently pulling from, but we've started transcribing 2 pages and they are not showing up in the "pages transcribed" statistics in the statistics tab.  May be related to #213 .

@suwiding

tbl73:
This seems to be pulling the info okay on my test server, so a couple of questions:   Is anything showing up in the statistic?  Is there more than one work?  Who is the owner of the collection?

tbl73:
Is this related to #365, which hasn't been merged into production yet?

saracarl:
@benwbrum and I brainstormed some ideas for fixing this.  

1)  one approach is turning off the revisioning during the import
wonder if we start with revision 0 or 1
if we start with 1, we could make a revision 0 for the IA work

2)  OR take a look at the code that determines whether a page edit is a transcription or not
Rather than driving that logic off of page version, we could look at the previous version's source_text and see whether it was blank.

3)  It may also be worth investigating why we are doing multiple versions for IA import?  Maybe it's the titling based on OCR?  Perhaps that's done in a second pass?
If we could keep it from generating the 2nd version, it'd both fix the deeds and whatever funkiness I imagine shows up on the version/diff view

tbl73:
I was digging further into this, trying to figure out where exactly the multiple versions are coming from.  Pages do start from revision 0, and they're set up (in the model) so that a new page version is created after each save (which makes sense).   I think what's happening is that the page is being saved twice during the conversion process, and so it's getting both version 0 (which is expected) and version 1 (which is not).  I think the relevant saves are in  line 84 and line 87 of the convert_to_work method in the ia_work model.  
So I don't know how exactly you could stop it doing two versions with the current setup unless you don't need both saves. (Though I agree, the version/diff view looks a little weird now, too.)

tbl73:
See #447 for some related issues.

tbl73:
Pushed up code in ia-stats.

Issue #451, opened 2016-10-20
OCR Correction deeds


tbl73:
We need to create a new deed type related to OCR correction.  Works imported from the internet archive which use OCR are currently creating transcription and edit deeds, when really they're a separate OCR correction process.  When a work is imported, it is set to revision 0.  The following revisions should have ocr_correction deeds (rather than transcription or edit deeds).
@benwbrum @saracarl - do you think we need to have an initial correction deed and separate correction edit deeds (following the model of transcription/translation), or just one type of ocr_correction deed?  It seems to me that they're really all one type of deed (ocr_corrected), but I wanted to double-check.

PR #452, opened 2016-10-20
Issues #411 and #447

tbl73:
Fixed transcription/edit statistics for works imported from internet archive.  Changed UI/wording for works that need OCR correction when imported.


saracarl:
On a collection page as owner, for a collection imported from IA  (http://localhost:3000/collection/show?collection_id=6) the "Recent Edits" activity on the right hand side still says "transcribed" even though the first action should be "corrected".

recentedits

tbl73:
That information comes from the deeds, so I think this needs to wait until the OCR correction deeds have been created, issue #451.

PR #464, opened 2016-11-14
Localtime

tbl73:
Change all the time displays in the UI to show as the localtime based on the user's computer time zone (rather than UTC).


tbl73:
In app/views/export/tei.html.erb and app/views/omeka_sites/_items.html.slim, I didn't add localtime because there are only dates displayed, not times, so I didn't think it would make much difference.  (Though I guess the date might be wrong if it was done late at night.)  

I don't see anything in app/views/collection/_contributors_body.html.slim that doesn't already have localtime included, unless I'm missing something?  I wrote the code originally using localtime, though, so I didn't make any changes to that file in this commit.

Issue #460, opened 2016-11-07
"create work" action on owner dashboard doesn't work


saracarl:
It just gives a yellow warning icon.

We may need to remove this -- the way to start a work at this point is to "start a project".  If we keep it, it needs to work!

saracarl:
put it here instead:

http://localhost:3000/collection/new_work?collection_id=16

tbl73:
Need to be able to create an empty new work for users who are uploading individual images.  Move into the "Start a Project" tab for consistency.  Remove from the Action menu.
Also, in the "Add a page" option in the Pages tab for a work, the modal currently has "Save & New Work" as an option, which should just be "Save".

saracarl:
So I spoke with Ben about this particular bit more:

Also, in the "Add a page" option in the Pages tab for a work, the modal currently has "Save & New Work" as an option, which should just be "Save".

And his view is that building a work via individual pages won't ever be done via the web UI, but would be a pretty common thing using the mobile UI.  

Mobile is addressed in #458  I suggest we back out the page view changes and address them in that feature/issue.  (Let's note them there.)

tbl73:
Closed with pull request #468

PR #468, opened 2016-11-28
Issue #460

tbl73:
Removed create work from Action menu, added the ability to create an empty work to the Start a Project tab.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -44,6 +44,7 @@ class DashboardController < ApplicationController
   #other methods in AddWorkHelper
   def startproject
     @work = Work.new
+    @work.collection = @collection
     @document_upload = DocumentUpload.new
     @document_upload.collection=@collection
     @omeka_items = OmekaItem.all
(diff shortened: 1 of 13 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -43,6 +43,7 @@ def index
   #Owner Dashboard - start project
   #other methods in AddWorkHelper
   def startproject
+    @work = Work.new
```
saracarl: I don't understand why we need a new work created here.  Isn't it done under the create_work under the add_work_helper?

app/controllers/page_controller.rb
```diff
@@ -77,10 +77,10 @@ def create
         process_uploaded_file(@page, @page.base_image)
       end
 
-      if subaction == 'save_and_new'
-        ajax_redirect_to({ :controller => 'dashboard', :action => 'owner', :anchor => 'create-work' })
-      else
+      if subaction == 'save_and_next'
```
saracarl: So (unfortunately) we need to undo the removal of "save and new".  

It doesn't make sense in the web UI, but it does in the mobile UI where you are taking pictures of i.e. letters page by page and adding them to an existing work.  When you reach the end of a letter you want to create a new work for the next one.  

(I'm sorry, that's a product management research problem, not a code problem.)

app/controllers/dashboard_controller.rb
saracarl: OK.  Makes sense.

Issue #413, opened 2016-10-04
NYBG: "export as xhtml" should export whole collection as a zip file


saracarl:
Instead of working work-by-work.

@suwiding

tbl73:
This depends on #494.

saracarl:
When exporting individually (click export xhtml, view web page, right click save-as), the header looks like:
<html><head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8"></head>
<body>
<h1 class="work-title">aivazian_sept1894</h1>
<div class="export-metadata">FromThePage export of aivazian_sept1894 from Brumfield Test Collection made on 2017-01-11 10:25:46 -0600.</div>

<hr>
<h2 class="divider">Page Transcripts</h2>

Within the zip it looks like:


<h1 class="work-title">aivazian_sept1894</h1>
<div class="export-metadata">FromThePage export of aivazian_sept1894 from Brumfield Test Collection made on 2017-01-11 10:25:06 -0600.</div>

<hr />
<h2 class="divider">Page Transcripts</h2>


We're missing this:

<html><head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8"></head>
<body>

(Also missing the closing body and html)

Does it matter?  Will this work for NYBG's purposes?

@benwbrum -- need your approval before we merge this in.

Issue #494, opened 2016-12-13
Export does not seem to provide valid XHTML


tbl73:
If you click Export as XHTML, a view of the data shows up.  If you save that page, it saves as HTML, not XHTML.  If you change the extension to xhtml, the file won't open.  

I'm pretty sure some of the problem is because that view isn't inheriting from layouts/application.html.slim - presumably it shouldn't, because the CSS shouldn't be included, but there's also no doctype declaration for the page then.

I ran the file (as xhtml) through an online validator, and the results are below:

Validation Output: 8 Errors

Error Line 3, Column 1: no document type declaration; implying "<!DOCTYPE html SYSTEM>"
<html><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8">…
✉
The checked page did not contain a document type ("DOCTYPE") declaration. The Validator has tried to validate with a fallback DTD, but this is quite likely to be incorrect and will generate a large number of incorrect error messages. It is highly recommended that you insert the proper DOCTYPE declaration in your document -- instructions for doing this are given above -- and it is necessary to have this declaration before the page can be declared to be valid.

Error Line 3, Column 86: end tag for "meta" omitted, but OMITTAG NO was specified
…tent="text/html; charset=UTF-8"></head><body><h1 class="work-title">[Letter to…
✉
You may have neglected to close an element, or perhaps you meant to "self-close" an element, that is, ending it with "/>" instead of ">".

Info Line 3, Column 13: start tag was here
<html><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8">…
Error Line 3, Column 86: end tag for "head" which is not finished
…tent="text/html; charset=UTF-8"></head><body><h1 class="work-title">[Letter to…
✉
Most likely, you nested tags and closed them in the wrong order. For example <p><em>...</p> is not acceptable, as <em> must be closed before <p>. Acceptable nesting is: <p><em>...</em></p>

Another possibility is that you used an element which requires a child element that you did not include. Hence the parent element is "not finished", not complete. For instance, in HTML the <head> element must contain a <title> child element, lists require appropriate list items (<ul> and <ol> require <li>; <dl> requires <dt> and <dd>), and so on.

Error Line 6, Column 5: end tag for "hr" omitted, but OMITTAG NO was specified
<hr>
✉
You may have neglected to close an element, or perhaps you meant to "self-close" an element, that is, ending it with "/>" instead of ">".

Info Line 6, Column 1: start tag was here
<hr>
[...]

PR #498, opened 2016-12-19
Export entire collection as zip, issue #413

tbl73:


tbl73:
The headers have been changed, per #494, and the files within the zipfile are now validating correctly.

Issue #501, opened 2017-01-05
Option to disable subject feature


saracarl:
We need the ability to turn off subject linking. 

-- available on a collection basis
-- I'd do it as a checkbox under "footnotes" and before "manage works" on the collection settings page
-- It should hide the "subjects" tab for a collection
-- It should hide the "autolink" button on the transcription screen
-- hide the following text on the transcription screen:
"Autolink" will suggest subjects certain words could be linked to or you can use double braces to link subjects. [[Jane Doe]] will link the text "Jane Doe" to the subject Jane Doe, while [[Jane Doe|Jane]] will link the text "Jane" to the subject Jane Doe. We recommend that linking be left to an editor after the initial transcription is made.

View helper that takes the xml text and turns it into a href anchor tag.  The view helper should check and see if subjects are enabled.  If they are not, then do not create the link in the view.  

Note that if you have subjects originally, then disable them, you could still get to the "normal" subject page by typing in the URL for it.  We've decided that is OK.  (We are not checking the setting on display of the subject page.)

Feel free to break this into multiple work items/issues.

saracarl:
Also need:

disable the Subject Categories on the collection main page
here at the bottom for example:
http://fromthepage.com/JuliaBrumfield

Hide top indexers list
Hide Disable percent indexed in work progress bar

Collection statistics page:
        Disable subjects
        Disable mentions
        Disable pages indexed
        Disable top indexers and percent indexed (as above)

Work read page:
        Disable category navigation on side panel
    Work contents page:
        Suppress “annotate” button by unindexed pages

 Work help Page:
        Move help text contents from page_block into template
        Disable indexing instructions within template.

tbl73:
Since this is such a lot of little pieces, here's a list of all the different things that were changed and which file they were changed in.  It includes all the things listed above plus a couple of additional pieces.

-  Collection settings - "disable subject indexing" checkbox - app/views/collection/edit.html.slim
-  Migration to add subjects_disabled column so this can be set at the collection level - db/migrate/20170111143351_add_subjects_disabled_to_collections.rb

 When subjects are disabled: 
- Collection overview 
     - remove subject categories
     - Remove % indexed from work stats and progress bar (color changed) - app/views/collection/show.html.slim & - app/models/collection.rb
- Collection - hide subjects tab - app/views/shared/_collection_tabs.html.slim
- Collection - statistics: remove Subjects, References, Pages Indexed
 	- Last 7 days stats - remove references, new subjects
[...]

PR #519, opened 2017-01-13
Disable subject indexing, #501

tbl73:


saracarl:
Transcription conventions OK, the subject indexing instructions are a different block of text.

Issue #495, opened 2016-12-14
Autolinking finds only first instance


tbl73:
The code to autolink subjects only finds the first instance of an entity - needs to be changed to find multiple entities if they exist.  (See the abstract_xml_controller.)

saracarl:
for a collection, pull out all verbatim text from page-subject links.
search in source text field for the verbatim
currently "stop" once we find the first match for verbatim
guessing that's done via a search & replace (i.e. .sub).  

Turn line 33-? into a while loop:
give us the index of the first match
do our linking 
then loop again, but start at the index of this text starting at the index +1 

(check how "index" works.)

Alternative:  do all this using very clever regex and gsub.  Challenge is to make sure stuff isn't double linked.  
Option 1)  brute force: substitute verbatim with subject link, then a "cleanup" phase -- look for 4 [[[[ and substitute/cleanup in those instances.
Option 2)  Regex that uses look-aheads to make sure that what it's looking for in the text doesn't have square brackets.

tbl73:
I ended up using a regex gsub with a block that can pass the matched text to a function that tests if it's already a link before doing the substitution.  In the autolinking branch.

PR #531, opened 2017-01-24
Autolinking #495

tbl73:


app/controllers/abstract_xml_controller.rb
```diff
@@ -7,7 +7,11 @@ module AbstractXmlController
   # belongs here.
   ##############################################
 
+  #constant - words to ignore when autolinking
+  STOPWORDS = ["Mrs", "Mrs.", "Mr.", "Mr", "Dr.", "Dr"]
```
saracarl: We need to add "Miss" -- it shows up all over 

We also need to add "he", "she", and "it"

tbl73:
I'm in the process of adding the stop words, but I need more information on
what you're referring to with the "case-changing scenario".

Issue #461, opened 2016-11-09
Create a guest user to allow transcription before a user account is created.


saracarl:
https://github.com/plataformatec/devise/wiki/How-To:-Create-a-guest-user

PR #536, opened 2017-02-03
Guest user account Issue #461

tbl73:


app/controllers/application_controller.rb
```diff
@@ -213,6 +236,14 @@ def after_sign_in_path_for(resource)
     end
   end
 
+#destroy guest user session if a user signs out, then redirect to root path
+  def after_sign_out_path_for(resource)
+    if session[:guest_user_id]
+      session[:guest_user_id] = nil
```
saracarl: Does this mean we don't transfer the deeds to the permanent guest user if they sign out?

app/views/article/show.html.slim
```diff
@@ -7,7 +7,8 @@
     -if description.present?
       .big ==description
     -else
-      i =link_to 'Edit the description in the settings tab.', { :controller => 'article', :action => 'edit', :article_id => @article.id }
```
saracarl: Let's not display this if the user is a guest.

app/views/shared/_page_tabs.html.slim
```diff
@@ -27,16 +27,27 @@
...
-    :name => 'Translate',
-    :selected => 6,
+    :name => 'Transcribe',
+    :selected => 3,
     :options => {\
       :controller => 'transcribe',
-      :action => 'translate',
+      :action => 'guest',
     },
   }]
```
saracarl: The previous block doesn't quite make sense to me.  If no current_user and the work is not restricted:

formerly:  show translate
now:  show transcribe  ??

I suspect it's fine, but I need a bit of a walk-through because it doesn't "make sense" to me.

lib/tasks/guest_cleanup.rake
```diff
@@ -0,0 +1,47 @@
+namespace :fromthepage do
```
saracarl: Can we make this take a variable amount of time we set in the rake task?  (I'm thinking a week is too big of a window.)  Let's do "more than X days old" and pass X to the rake task.

app/controllers/application_controller.rb
tbl73: No, this just destroys the guest session; it doesn't do anything to the temporary guest account that was created in the database.  (Actually, it really just removes the "guest_user_id" from the session variable, which it what it checks for to tell that a user is a guest.)

app/views/article/show.html.slim
tbl73: Do you not want to display the description, or articles at all?

lib/tasks/guest_cleanup.rake
tbl73: Yep, I'll change it to take that as a variable.

app/views/article/show.html.slim
saracarl: A guest should not be able to edit the description.  Transcription and translation is really all we want them to be able to do.

saracarl:
If a page is not transcribed at all, we display a "This page is not transcribed, please help transcribe this page" message.  The "help transcribe" is linked to the "transcribe" screen, and if you are not logged in that takes you to the sign in page.  This needs to take you to the guest user transcribe page if you are not logged in.

The message shows up in the "read work" view (http://localhost:3000/display/read_work?work_id=52) as well as the individual page views (http://localhost:3000/display/display_page?page_id=348) and needs to work in both places.

saracarl:
I'm getting a spec test failure now:

Failures:

  1) disable subject linking checks work level subject items
     Failure/Error: expect(page).not_to have_content("Categories")
       expected not to find text "Categories" in "DashboardCollectionsFAQSigned In AsMargaretOwner DashboardCollaborator DashboardYour ProfileSign OutCS PierceMS_844ReadAboutContentsVersionsSettingsPagesHelp1This page is not transcribed, please help transcribe this pageTranslation2This page is not transcribed, please help transcribe this pageTranslation3This page is not transcribed, please help transcribe this pageTranslation4This page is not transcribed, please help transcribe this pageTranslation5This page is not transcribed, please help transcribe this pageTranslation← 1 2 3 4 →Records 1 – 5 of 16SearchCategories PeoplePlaces © 2017 FromThePage. All rights reserved.AboutTerms & ConditionsPrivacy PolicyContact Us"
     # ./spec/features/disable_subjects_spec.rb:61:in `block (2 levels) in <top (required)>'

Finished in 2 minutes 5.3 seconds (files took 2.18 seconds to load)
65 examples, 1 failure

Failed examples:

rspec ./spec/features/disable_subjects_spec.rb:52 # disable subject linking checks work level subject items

saracarl:
So I'm thinking I was supposed to create a permanent guest user and didn't, and that might be why I'm seeing this, but in my deeds, for a user that transcribed (or rather OCR corrected) but didn't create their own user I'm seeing:

/user/profile?user_id=36 corrected the OCR transcription of Untitled Page 3 page in Test collection

That "/user/profile?user_id=36" I'm assuming is my temporary guest user, but it isn't pretty....  What's supposed to happen here, and how can we fix it?

saracarl:
While testing other PRs, but using my database with a guest user (non converted) in it, I ran across this:

http://localhost:3000/collection/edit?collection_id=6
undefined method `empty?' for nil:NilClass
Extracted source (around line #39):

    @owners = [@main_owner] + @collection.owners
    @nonowners = User.order(:display_name) - @owners
-->   @nonowners.each { |user| user.display_name = user.login if user.display_name.empty? }
    @works_not_in_collection = current_user.owner_works - @collection.works
  end

I suspect it is related to the non-cleaned up or converted user.

saracarl:
Hmm.  I deleted user_id 36 from my DB, thinking it was the one causing problems -- here's what it looked liked:

2.2.0 :003 >   User.find(user_id=36)
  User Load (0.8ms)  SELECT  `users`.* FROM `users`  WHERE `users`.`id` = 36 LIMIT 1
 => #<User id: 36, login: nil, display_name: nil, print_name: nil, email: "guest_148677745010@example.com", owner: false, admin: false, encrypted_password: "", password_salt: "", created_at: "2017-02-11 01:44:10", updated_at: "2017-02-11 01:44:10", remember_token: nil, remember_token_expires_at: nil, location: nil, website: nil, about: nil, reset_password_token: nil, reset_password_sent_at: nil, remember_created_at: nil, sign_in_count: 0, current_sign_in_at: nil, last_sign_in_at: nil, current_sign_in_ip: nil, last_sign_in_ip: nil, account_type: nil, paid_date: nil, guest: true> 
2.2.0 :004 > User.delete(user_id=36)
  SQL (4.8ms)  DELETE FROM `users` WHERE `users`.`id` = 36
 => 1 
2.2.0 :005 > 

And the last error still occurred.  So either the other "converted" user (name "Sara Guest" most likely user_id=35) is causing the problem or it's something else.  Still worrisome.

tbl73:
I'm not sure why you'd be getting an error on this spec test - the code
isn't really related, but maybe something got overwritten? I'll have to
take a look.

tbl73:
I'm pretty sure this is related to the undeleted user - it probably needed
to be cleaned up with the rake task rather than deleted directly.  I'll
check on that Monday as well.

tbl73:
You should be seeing "Guest" (without having to do anything) - I'm a little
concerned about why you aren't.  I'll check it out on Monday.

saracarl:
I went back and tested with a copy of the production database, and both an abandoned guest account and a claimed one are working as designed, so I'm not sure what was going on with my earlier testing.

tbl73:
That's good to hear!  I was concerned that you weren't seeing "Guest" for
the abandoned account - display name problems would affect a lot of
different pages.

saracarl:
We decided this error:

> _http://localhost:3000/collection/edit?collection_id=6
> undefined method `empty?' for nil:NilClass
> Extracted source (around line #39):
> 
> @owners = [@main_owner] + @collection.owners
> @nonowners = User.order(:display_name) - @owners
> 
> --> @nonowners.each { |user| user.display_name = user.login if user.display_name.empty? }
> @works_not_in_collection = current_user.owner_works - @collection.works
> end_

Was caused by having a guest user without a display name, and running a branch without the guest user code that handles and returns a "guest" for the display name.  (So no worries there...)

Issue #538, opened 2017-02-04
Collection footnote should show up on every page in a collection


saracarl:
Footnote not showing up on indiviual pages. -- only on collection home page

PR #546, opened 2017-02-08
#538, #543 - Footer visible on each page.

tbl73:


saracarl:
Looks good visually, spec tests pass.

Issue #517, opened 2017-01-12
Collections overview - %Indexed


tbl73:
The progress bar with % indexed on the collections overview page isn't actually registering the number of indexed pages - it just says 0% regardless of how many pages have been indexed.

tbl73:
This was basically just a small typo, so i added it in the autolinking branch.  One note, though; the percent indexed doesn't update unless the work or a page within it is saved.  So once this code is pushed up, there will probably need to be some cleanup so that they show correctly.

tbl73:
We'll add a migration that calls recalculate work statistics for each work.

tbl73:
Turns out that fixing the typo does not seem to have totally resolved the problem for works with more information, so I'll look at it again.

PR #547, opened 2017-02-08
#517 - Migration to fix %indexed statistic

tbl73:


saracarl:
Had to update the test environment to get the spec tests to pass, but checked in that file and they are all good.  

Verified visually as well.

Issue #489, opened 2016-12-12
add "login as" option for administrators


saracarl:
There are gems that do this for you that work with devise.

Good for support and demos.

saracarl:
please research & then we'll discuss.

tbl73:
Noting the gems/instructions I've found as possibilities.
https://github.com/rcook/user_impersonate2
https://github.com/ankane/pretender
https://github.com/flyerhzm/switch_user
Not a gem, but a method of doing this: https://github.com/plataformatec/devise/wiki/How-To:-Sign-in-as-another-user-if-you-are-an-admin

PR #548, opened 2017-02-08
#489 Ability for admin to login as a user

tbl73:


tbl73:
I replaced the default masquerades controller (very similar to what I did with the registrations controller for guest user accounts) and added authorization in, so a user can no longer URL hack.  I also updated the tests.  This change does require a change in the routes file as well, so we'll have to figure out the best way to manage that.

Issue #552, opened 2017-02-13
Deed list on owner-user homepage


saracarl:
A filtered deed list for all collections/works owned by this user.

Issue #550, opened 2017-02-13
Owner-user homepage should list collections first


saracarl:
List Collections.

PR #561, opened 2017-02-22
#550 and #552, Owner homepage

tbl73:


saracarl:
This is weird.  If you test with a backup of production, this collection:
http://localhost:3000/collection/show?collection_id=13
Is owned by this user:
http://localhost:3000/user/profile?user_id=214
but the user profile is very much a contributor, not an owner.

saracarl:
tests pass, but would like to understand the cause of the earlier issue before we merge this in.

tbl73:
Well, oddly, if you look at that user's information, she isn't actually set as an owner, even though she owns that collection.  The profile checks for the owner flag, so that would explain why she has a "user" profile rather than an "owner" profile.  I don't know how or why she owns the collection without being flagged as an owner, though.

saracarl:
Possibly we turned her "off" as an owner at some point after they finished
their project.

I'd say "working as designed"!

Sara

Issue #527, opened 2017-01-20
"Create Collection" should redirect to start project


benwbrum:
Currently the "Create Collection" action takes the user to the collection settings screen.  This isn't unreasonable, but people starting projects from scratch generally have files in hand ready to upload.  We should redirect them to the Start a Project screen instead.

PR #571, opened 2017-03-03
#527 - create collection redirect

tbl73:


saracarl:
reviewed and tested, looks good.

Issue #558, opened 2017-02-15
sign in screen needs a reminder


saracarl:
People try to log in using their email address instead of their user ID.
How easy is it to let someone use their email to log in?  That's the idea (user ID OR email).  

Failing that, can we put some text on the sign in screen that says "userid NOT email address"?  

I get support questions on this regularly.  It's likely our biggest usability issue.

PR #572, opened 2017-03-03
#558 Allow user login by email or login id

tbl73:


saracarl:
tested & ui reviewed;  good.

Issue #570, opened 2017-03-02
IIIF Manifest duplicate imports


tbl73:
When you import an IIIF manifest via a link, the import process is creating two works - one in a collection that you select, and one in a collection of its own.

saracarl:
@jbhoward-dublin this was deployed on fromthepage.com today.  You shouldn't see any more duplicates when importing via a iiif manifest.

jbhoward-dublin:
Has this been implemented on the fromthepage.com site? I have imported two manifests as works to add to an existing collection, but they continue to also appear as new Collections (this with the Luke Wadding Collection)

saracarl:
We'll investigate, @jbhoward-dublin and see why this is still happening.  (Hopefully it's just a deployment issue.)

saracarl:
Fixed.  (Changes were merged, but not pushed back out to github.  Sigh.)

PR #573, opened 2017-03-03
#570 - Duplicate IIIF imports

tbl73:


saracarl:
reviewed and tested.  looks good.

PR #577, opened 2017-03-06
Delete child associations and clean up orphaned data

tbl73:


tbl73:
The ui-design version of config/routes.rb is the correct one - this branch was created before that had been merged in.

db/migrate/20170301150018_delete_orphaned_data.rb
```diff
@@ -0,0 +1,48 @@
...
+    Work.all.each {|w| w.destroy unless w.collection}
+    
+    #delete pages for which there are no works
+    Page.all.each {|p| p.destroy unless p.work}
+    
+    #delete document sets for which there is no collection
+    DocumentSet.all.each {|d| d.destroy unless d.collection}
+
+    #delete notes for which there is no collection
+    Note.all.each {|n| n.destroy unless n.collection}
```
saracarl: Notes are on a collection, rather than a work or page?  I wonder if this could go wrong?  

 # associations
  belongs_to :user
  belongs_to :page
  belongs_to :work
  belongs_to :collection

Any chance a note would have a work or a page, but no collection?  Any chance the collection would be wrong?

db/migrate/20170301150018_delete_orphaned_data.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    #delete omeka items for which there is no work
+    OmekaItem.all.each {|i| i.destroy unless i.work}
+
+    #delete sc manifests for which there is no work
+    ScManifest.all.each {|s| s.destroy unless s.work}
+
+    #delete sections for which there is no work
+    Section.all.each {|s| s.destroy unless s.work}
+
```
saracarl: Sections are more closely associated with pages, rather than works.  I wonder if the unless should be s.page?

db/migrate/20170301150018_delete_orphaned_data.rb
saracarl: Thinking more.  People delete works, pages, and collections and any of the above could leave behind orphaned notes.  I think this needs to be fleshed out more to get the logic right.  Maybe the order is important too -- start with the highest things in the hierarchy (collections => works & document sets => pages => notes) so that the "parent" of a particular thing is gone by the time we check for it?

saracarl: OK, I see that this migration is mostly using the hierarchy concept.  I think we should only thing through the notes bit and likely move to the end and check for the pages rather than collections.

db/migrate/20170301150018_delete_orphaned_data.rb
saracarl: And move it downward (close to notes)

db/migrate/20170301150018_delete_orphaned_data.rb
tbl73: My logic was that a note should be deleted if it belongs to a collection which doesn't exist, because you can't have works or pages (or anything, really) on a collection that doesn't exist.

saracarl: Checking the production db shows
No notes orphaned by page
No notes orphaned by work

this is fine.

db/migrate/20170301150018_delete_orphaned_data.rb
saracarl: WorkID is the only association Section has.  Also, there are no orphaned sections, so this shouldn't do anything.

Issue #518, opened 2017-01-12
surface progress bar on export screen


saracarl:
Customers want to export complete works.  We should let them see completeness on the export screen.

Issue #506, opened 2017-01-08
pages marked blank should not display the "help transcribe" text


saracarl:
If marked blank, do not display the "This page is not transcribed, please help transcribe this page" message in the work view.

for example:
http://fromthepage.com/display/read_work?work_id=266

Issue #502, opened 2017-01-05
Mark document as complete to get to 100%.


saracarl:
remove the document state drop down and add a “save and complete” button next to the current “save” button. We'll then have 3 states – 0% (not started), 50% (started/some transcription), 100% (“save and done” clicked).

saracarl:
We discussed this with the customer, and settled on a checkbox on each page view that is "needs review".  This checkbox should replace the status drop down in terms of layout and formatting.

The status screen (collection homepage listing works & status -- http://fromthepage.com/JuliaBrumfield ) should then reflect pages by percentage based on 3 states:

1)  has any transcription + "needs review" not checked = transcribed; green in status bar
2)  has any transcription + "needs review" checked = does not count as transcribed + orange in status bar
3) has no transcription = does not count as transcribed

Open question: how do we mark blank pages?  I'm wondering about a checkbox for that.

saracarl:
@benwbrum : need your input on blank page indicators.

saracarl:
4 states:  transcribed, indexed (or not), needs review, or nothing.  Figure percentage of each one -- no overlaps (if something is transcribed and needs review, then it is in the "needs review" bucket/percentage, rather than the "transcribed" bucket.  Same for indexed.)

States should be listed on the green status bar in the following order from left to right:
indexed, transcribed, needs review, nothing.

needs review should be done in yellow.

saracarl:
Remove status dropdown.  Add 2 checkboxes on each page view that is "needs review" and "blank"  on the same line as "save" and and "fullscreen", but aligned closer to the "save" button than the fullscreen button.

tbl73:
Need to create a tab that lists the pages that need review.

PR #597, opened 2017-03-21
Mark pages a needs review - #502, #506, #518

tbl73:
This includes #502, #506, and #518.  In addition, we're now using the status and (new) translation_status fields to track the state of a page.  The possible states are nil, blank, review, transcribed (or translated), and indexed, for either the transcription or translation of a page.  Work statistics are based off those states.  We also switch out the statistics/progress bar depending on whether the work is a transcription or translation work.


saracarl:
Ok, found a problem.  

To reproduce on a recent production DB:  
http://localhost:3000/display/read_work?page=2&work_id=308
Mark the first 2 pages as blank
then we get to 100% transcribed, but only 75% indexed, even though all the transcribed pages have subject indexes on them.  Blank needs to count as indexed.

saracarl:
What's the right behavior if a page is marked both blank and needs review?  (Did we discuss that?)  We're treating it as "blank" rather than "needs review", but I suspect it should be the opposite.  (Treat it as "needs review" until they uncheck that box.)

tbl73:
I switched the blank pages to being included as indexed pages.  (I think they were removed from indexed sometime early in the process because they ended up being double-counted in those calculations, before the latest iteration of the statistics.)
On the blank and needs review, I think we discussed it and decided to make blank the priority, but it makes to set it to needs review instead, so I changed that.  The "blank" wording won't show up, because it's marked for review rather than as blank.

saracarl:
One more change here -- When a page is marked blank and saved, the checkbox in the UI switches to "page is not blank".  I prefer that it stays at "page is blank" with a check next to it.  It's clearer.

saracarl:
Is there any way we can toggle the "needs review" and "blank" checkboxes in the UI, i.e. so when you check "needs review" it unchecks "mark page as blank" and when you check "mark page as blank" it unchecks "needs review"?

That would be a much clearer view of what is happening under the covers.

tbl73:
I think they can both be done, although it may take me a little bit.  Certainly keeping the checkbox can. That's actually more complicated than it sounds, but it's similar enough to the needs review one.  In terms of toggling, I probably can do something with jquery, but will need to investigate a bit.

saracarl:
Can you open new issues for those 2 (assign them to yourself, and put them in the ready pile so we get to them sooner rather than later)?   I'm going shipping with these small UI issues as long as we improve them shortly.

Issue #580, opened 2017-03-09
Default subject categories not created for standalone IIIF imported collection


tbl73:
If you import an IIIF manifest and have it create a new collection, the default subject categories aren't being created.  I think the creation code needs to be moved from the collection controller to a before_save hook in the model.

benwbrum:
I agree completely, though an `after_save` hook might work better.

tbl73:
Yeah, now that you say that, after_save does make more sense.

PR #600, opened 2017-03-21
#580 default categories

tbl73:


saracarl:
So when I import a IIIF manifest I don't have the option to create a collection.  I get a drop down with "Select an existing collection" and a list of collections.  ?!???

tbl73:
That's been true since (at least) before I did the last bit of work on IIIF imports - I didn't actually know "create a collection" was supposed to be an option.  If you use the links to find a collection, you can choose to use the parent collection, but otherwise it's just that drop down.

Issue #559, opened 2017-02-16
new users email (rake task) should have links to user profile/activity


saracarl:
I clicked on an email address and it opened an email.  I'd like it (or the user ID, although guests don't have them) to link to their profile so I can see what they did.

saracarl:
Should check the collection activity email to see if it needs the same thing.

tbl73:
@saracarl Can you forward me the email that you're referring to, so I can look at it?

PR #606, opened 2017-03-23
#559 - User email with link to user profile

tbl73:
I changed the user login to show the user profile, with code similar to the code from the collection stats.  The "recent deeds" part of the email doesn't work in my development environment, but I'm assuming that part is okay for you in production, since you're receiving the emails in the first place.


tbl73:
This work should also include #611 , so I'm going to close this pull request and open a new one once that work is done.

PR #640, opened 2017-04-25
Owner emails

tbl73:


app/mailers/admin_mailer.rb
```diff
@@ -26,6 +26,15 @@ def email_stats(hours)
     mail from: SENDING_EMAIL_ADDRESS, to: ADMIN_EMAILS, subject: "FromThePage activity in the last #{hours} hours."
   end
 
+  def collection_stats_by_owner(owner, start_date, end_date)
+    @collections = owner.all_owner_collections.joins(:deeds).where(deeds: {created_at: start_date..end_date})
+    @start_date = start_date
+    @end_date = end_date
+    
+    mail from: SENDING_EMAIL_ADDRESS, to: owner.email, subject: "FromThePage collection activity in the last 24 hours"
```
saracarl: Because users have multiple collections (and therefore multiple emails), is there a way to include the collection name in the subject?

saracarl: Ah, maybe I misunderstand.  One email with all collections listed, similar to the admin email?

tbl73: Yes, they get one email with all the collections listed in it, and no email at all if there hasn't been any activity.

tbl73:
Yep, we can certainly go over it.  The mailer previews are actually for development - they allow you to see what the email will look like while you're creating the layout, without having to send it out every time.

app/views/admin_mailer/collection_stats_by_owner.html.slim
```diff
@@ -0,0 +1,8 @@
+-@collections.each do |c|
+  -new_contributors(c, @start_date, @end_date)
+  
+  h2.nomargin ="#{@collection.title}"
+
+  h2.nomargin ="Contributions in the last 24 hours"
```
saracarl: Can we calculate the hours intead of hard coding it?

lib/mailer_previews/admin_mailer_preview.rb
```diff
@@ -0,0 +1,16 @@
+class AdminMailerPreview < ActionMailer::Preview
+  
+  def contributor_stats
+    AdminMailer.contributor_stats(63, 2.weeks.ago, Time.now, 'trishablewis@gmail.com')
```
saracarl: Change the email to the admin mail address (defined in the initializer, I think.)

tbl73:
Okay, I've made those changes.

lib/mailer_previews/admin_mailer_preview.rb
```diff
@@ -0,0 +1,16 @@
...
+  def contributor_stats
+    AdminMailer.contributor_stats(1, 2.weeks.ago, Time.now, ADMIN_EMAILS)
+  end
+  
+  def email_stats
+    AdminMailer.email_stats(48)
+  end
+  
+  def collection_stats_by_owner
+    owner = User.find_by(login: 'admin')
```
saracarl: I don't actually thing 'admin' owns any collections in the production database.

Issue #636, opened 2017-04-24
trial owner account


saracarl:
Embed code for calendly:

<div class="calendly-inline-widget" data-url="https://calendly.com/fromthepage" style="min-width:320px;height:580px;"></div>
<script type="text/javascript" src="https://calendly.com/assets/external/widget.js"></script>

Also add a "owner" designation to the admin report / new user section

Make the user an owner when the account is created/right after it is created.

saracarl:
please add "2 weeks from now" as the expiration field

PR #646, opened 2017-04-28
#636 Trial owners into fromthepage.com

tbl73:


saracarl:
See issue #691 -- mustfix before this can be tested.

saracarl:
This did not create an owner account -- just a collaborator account.  and it redirected to the pricing page instead of anyplace "useful" (an owner should be sent to the "start a project" page in the owner dashboard).

Issue #649, opened 2017-05-01
tinyURL for collections


saracarl:
-- short string that a user can set on a collection.
-- check to make sure it's unique
-- route to the collection homepage
-- dynamically generate it (based on the collection title, downcased and hyphenated), but allow overriding.

saracarl:
Migration that generates slugs for every existing collection
manually migrate routes file for the manually set up ones (check fromthepage.com routes file)

Issue #416, opened 2016-10-05
SEO friendly work and collection URLs


saracarl:
So:
http://fromthepage.com/collection/show?collection_id=1
Should become one of:
http://fromthepage.com/collection/JuliaBrumfieldDiaries
http://fromthepage.com/JuliaBrumfieldDiaries
http://fromthepage.com/BenBrumfield/JuliaBrumfieldDiaries
we'll have to decide, but I'm leaning toward the last with the work owner by name (not userid), then the collection name.

and 
http://fromthepage.com/display/read_work?work_id=2
Should become one of:
http://fromthepage.com/JuliaBrumfieldDiaries/1918
http://fromthepage.com/BenBrumfield/JuliaBrumfieldDiaries/1918

Should also consider what happens if/when (or if this is part of) a customer org subdomain, i.e.:
http://benbrumfield.fromthepage.com/JuliaBrumfieldDiaries/1918
(or can it be  both?)

atomrab:
Have you thought about clean URIs (for collections, works, and pages) for linking? I know that's at odds with the SEO goal, but perhaps more useful for scholarship/citation.

saracarl:
@atomrab I actually think clean URIs and SEO friendly URLs are the same thing.  (see https://en.wikipedia.org/wiki/Semantic_URL )  

If you have counter examples or improvements to the ones above I'd be happy to discuss them.  I'm not 100% sure what the "right" patterns are, I just know we need meaningful ones.

benwbrum:
We know from the document set breadcrumb issue #289 that we need to look at
routes and nested resources.

At the moment if you navigate from document set S to work W, you can only
get back to collection C, and can never get back to S.

If our URLs looked a bit more like `
http://fromthepage.com/collection/C/work/W` for a set-less work, and `
http://fromthepage.com/colleciton/C/set/S/work/W` or `
http://fromthepage.com/set/S/work/W` then we'd have everything we need to
generate breadcrumbs and other navigation specific to sets.

benwbrum:
Further thoughts on Collection-level slugs:

We should add a new `slug` field to the `collections` table and add a slug input field on the Collection Settings tab.  This needs to have the following validations:
* Slugs must not contain whitespace
* Slugs must be unique in a case-insensitive manner

We should consider populating slugs from CamelCase versions of collection titles.

[...]

PR #659, opened 2017-05-19
#416, #649 SEO friendly URLs

tbl73:


config/routes.rb
```diff
@@ -66,4 +59,28 @@
...
+#    end
+    resources :collection, path: '', only: [:show] do
+#    get 'new_work', on: :member
+#    get 'contributors', on: :member
+      get '/statistics/collection', path: '/statistics', as: :statistics, to: 'statistics#collection'
+
+      match 'display/read_work', path: '/:id', as: :read_work, to: 'display#read_work', via: [:get, :post]
+
+#      get 'transcribe/display_page', path: '/:work_id/:id', as: :transcribe, to: 'transcribe#display_page'
+    end
```
saracarl: clean up the commented out stuff.

app/controllers/application_controller.rb
```diff
@@ -211,6 +222,7 @@ def load_html_blocks
   def authorize_collection
     # skip irrelevant cases
     return unless @collection
+    return if @collection.is_a?(DocumentSet)
```
tbl73: Question about rewriting line 228 - would that affect access if someone is using an old URL to access the document set?

tbl73:
It turns out that notes also include a work_id, so I was able to limit the document_sets.notes method to notes that are in the works from the document set.

app/controllers/application_controller.rb
saracarl: We're willing to break existing document set URLs (so few -- or no -- users), so don't worry about this.

tbl73: Okay, just pushed that change up.

app/views/work/edit.html.slim
```diff
@@ -17,13 +17,13 @@
...
+          td =f.text_field :slug, value: @work.slug
+        tr
           th
           td 
-            p <b>Current URL:</b> The current URL for this work is <b>www.fromthepage.com/#{@collection.owner.slug}/#{@collection.slug}/#{@work.slug}</b>.  If you want to edit the work section of the URL, please use lowercase letters and dashes between any words.
-        tr
-          th =f.label :slug, 'Edit URL'
-          td =f.text_field :slug, value: @work.slug
-
+            |The current URL for this work is <b>www.fromthepage.com/#{@collection.owner.slug}/#{@collection.slug}/#{@work.slug}</b>.  If you want to edit the work section of the URL, please use lowercase letters and dashes between any words.
```
saracarl: here too.

Issue #661, opened 2017-05-22
Document Sets settings tab


tbl73:
In the view of a document set by an owner, we need to create a settings tab to manage the document set settings.  This should include:
-- Text: This is a document set based on the XXX collection. You can manage all works and document sets within that collection here (link to the collection "sets" tab.)
-- partial that is currently displayed when you "Edit" the document set

Issue #660, opened 2017-05-22
Tabs for collection/work


tbl73:
As part of the document sets breadcrumbs and pretty URLs, the tabs show in a collection/work/page view will need to be rewritten to use new routes.   We may also want to consider having a separate tab set for collections vs document sets.  Document sets should include tabs for overview, settings, statistics, subjects, collaborators.  They should not have tabs for exports, sets, add work.

tbl73:
Currently we're setting a variable in the view that's used to pass the collection id.  Using a method on document sets doesn't work in this case, because the ID is being passed directly, and it could be a document set or a collection.  When the tabs are rewritten, we should be able to pass the object rather than the id (using the Rails helpers/paths), and at that point we can get rid of that variable.

Issue #658, opened 2017-05-16
Pretty URLs - pages


tbl73:
Once the SEO-friendly/pretty URLs for collection/work/user are set up, we need to make the same kind of URLs for pages so that they can be easily shared.

Issue #289, opened 2016-04-11
[FTP UT Library install]: Interference between set/collection dashboards


atomrab:
Document sets appear on the same level as "Collections" on the "Collections" page, as below:

![image](https://cloud.githubusercontent.com/assets/4740995/14444067/eae900ea-0008-11e6-99e0-33d3506dde8f.png)

When one clicks on a document set from this page, the initial overview page contains the document set and its documents, which can be viewed individually (from the "read" tab) and viewed page by page (from the "overview" tab) within that set (in a format identical to the collection view, below):

![image](https://cloud.githubusercontent.com/assets/4740995/14444121/47f84156-0009-11e6-9fd0-b178ab63336b.png)

But any click on another tab (versions, transcribe, settings, etc) from any of those pages directs the user (at least an admin user, because I can't see these when I'm not signed in) to the document page in the parent collection for the set, as below:

![image](https://cloud.githubusercontent.com/assets/4740995/14444163/944b3748-0009-11e6-9fd3-fc4d113bf4ca.png)

benwbrum:
Painful as it may be, the right solution to what is fundamentally a breadcrumb problem is probably to implement nested routes, with a top-level for document set and child for work/page.  See the [rails guide on the subject](http://guides.rubyonrails.org/routing.html#nested-resources).

benwbrum:
See also #25 and the work Diego did on this.

atomrab:
Is the nested route solution something the library needs to implement, or something that you will/have implemented? The issue still appears on the current library install -- is there an update we should apply?

PR #670, opened 2017-06-07
Document set breadcrumbs

tbl73:
This includes issues #289, #658, #661, and #660.


saracarl:
Works except for the scenario I describe in #699

Issue #682, opened 2017-06-22
don't fail upload if smtp not configured


atomrab:
When an owner tries to upload a zip file to an existing collection, the "Browse" pathway renders as e.g. C:\fakepath\Archive2.zip, and when the owner clicks the "Upload File" button, the interface returns a 500 Internal Server error. It is not possible to edit the filepath to see if this is the problem (any action in that field after a file has been browsed to triggers a new browse-to-file window).

Is this something going on with our server, or is it a bug in the interface?

saracarl:
I suspect it's a configuration issue.  It looks like the SMTP (mailer) configuration was either wiped out by a recent upgrade or the upgrade got the code to the point where the mailer is required.  

I, [2017-06-22T15:50:29.504327 #2286]  INFO -- : Completed 500 Internal Server Error in 443ms
F, [2017-06-22T15:50:29.506410 #2286] FATAL -- : 
Net::SMTPFatalError (554 Message rejected: Email address is not verified. The following identities failed the check in region US-WEST-2: support@fromthepage.com
):
  app/helpers/add_work_helper.rb:35:in `new_upload'

It should be a pretty fast fix for someone who knows the SMTP settings UT uses and can log into the server.  See step 3 of the following document : 
https://github.com/benwbrum/fromthepage/wiki/Configuring-FromThePage-For-Your-Environment

Or I can look over the config/environments/production.rb file and see what looks to be the problem. (send it via email rather than posting it here).

saracarl:
And I think once we get this specific issue resolved we should fix the code path that causes uploads to fail without a successful email -- this is the second time in as many weeks I've had this problem happen in a customer environment.

tbl73:
Question - do you still want those system mails to get sent if a document is uploaded?  (I'll change the code so an unsuccessful email doesn't break the upload process, but should they get sent in general?)

saracarl:
Yes, emails should be sent on a successful upload. 

Sara

PR #698, opened 2017-07-11
#682 - Uploads failing on SMTP errors

tbl73:
I set up a module for SMTP error handling, to prevent the uploads from failing if there's something wrong with the SMTP configuration.  Right now it's only applied to the document uploads (in both the document upload process and the rake task), but it could be applied to other SMTP code if needed.

### DIFF
--- a/lib/tasks/ingestor.rake
+++ b/lib/tasks/ingestor.rake
@@ -1,6 +1,5 @@
 require 'image_helper'
 require 'open-uri' # TODO: Move elsewhere
-
 namespace :fromthepage do
 
   desc "Resize image file or directories of image files"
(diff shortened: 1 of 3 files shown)


lib/tasks/ingestor.rake
```diff
@@ -41,9 +43,13 @@ namespace :fromthepage do
       document_upload.save
     end
 
-    if SMTP_ENABLED    
+    if SMTP_ENABLED
+      begin
         SystemMailer.upload_succeeded(document_upload).deliver!
         UserMailer.upload_finished(document_upload).deliver!
+      rescue *SMTP_ERRORS => e
+        print "SMTP Failed: Exception: #{e.message}"
```
benwbrum: Shouldn't this be a call to `ErrorHelper` as well?

tbl73: The *SMTP_ERRORS list comes from ErrorHelper, although it isn't explicitly called.  I didn't call log_smtp_error from within ErrorHandler because the rake task doesn't recognize the logger call from that method.  I also thought it made sense to have any errors go into the process log, since that's where the rest of the logging goes.

saracarl: That makes sense.

Issue #708, opened 2017-07-12
newline transcription convention


saracarl:
This line in transcription conventions always shows up -- can't get rid of it:

A single newline indicates a line-break in the original document, and will not appear as a break in the text in some views or exports. Two newlines indicate a paragraph, and will appear as a paragraph break in all views. 

move to default conventions, out of the template

tbl73:
We'll run a migration to update the default conventions of existing collections.  Collections with conventions that are different than our default will need those lines to be manually added.

PR #725, opened 2017-07-24
#708 - update default transcription conventions

tbl73:
This includes a migration to update the transcription conventions on collections which use the default.  Any collections with changes to the defaults will need to be manually updated.


benwbrum:
Merged

Issue #156, opened 2015-04-19
Contents page actions wrong for OCR


benwbrum:
The contents page needs its logic changed to be aware of page statuses and OCR

PR #734, opened 2017-07-26
#156 - work contents page

tbl73:


saracarl:
So the UI here is confusing:

1)  a "complete" button to me means "push this button to complete the page".  I guessed it was going to take me to transcribe.  It doesn't.  It actually means "completed".  Can we try it with that word?

2)  "Annotate" is not a word we use anywhere else in the site.  I think we need to use the word "index" to reflect our subject indexing language elsewhere.  

3)  I think we need to add a "needs review" to this page as well.

tbl73:
1 and 2 are both clear - I don't really understand what you mean with 3,
though.  The buttons should be showing "review" if the page is marked as
needs review.  If you mean filtering the pages, I think we'll do that when
we rewrite the page.

Issue #211, opened 2015-08-10
Owner stat: number of index entries in last 7 days


benwbrum:

tbl73:
PR #735

PR #735, opened 2017-07-27
#211 last 7 days index stat

tbl73:


saracarl:
When I tested this for a collection, the "references" count was updated, but "pages indexed" was not.  I was working on a work that had some things indexed, but went to a page that did not have anything indexed and auto-indexed it.  I would have expected the "pages indexed" for "last 7 days" to increment by one.

tbl73:
I used an existing statistic to display that the "pages indexed" number -
I'll have to look more closely at what it's actually doing.

tbl73:
Can you test it again?  I tried in a couple of collections, both with things indexed and not, and the count seems to be incrementing correctly for me.  (I also tried pulling down the code into a branch, and it still seemed fine.)

saracarl:
This is still flaky, for reasons I don't get.  If you look at the attached screenshot you'll see that the "overall" transcribe count is 1, and there's no way when you look at the status of each document that that is correct.

bad_transcribe_count

Not to mention the transcribe count is for last seven days.

I think that maybe we aren't counting some set of pages, and if I index one of those pages, it doesn't get counted in the index statistic either.  (Theory, I can't prove that.)

Issue #736, opened 2017-07-27
Show/hide completed works in collection homepage


tbl73:
Give users the option of showing or hiding completed works via a checkbox.  Our default should hide completed works.  Add a setting in the collection settings to allow owners to set the default to showing completed works.  Use javascript with the checkbox so that everything is available for google searches but completed works are not visible.

tbl73:
Note - something similar to this link: https://transcription.si.edu/browse?filter=owner%3A16&sort=popular

tbl73:
The logic we're using to determine if a work is completed: unless subject indexing is turned off, a work is complete when it is 100% indexed, and a translation work is complete when the translation is 100% indexed.  If subject indexing is turned off, it's complete when it's 100% transcribed (or 100% translated).

PR #746, opened 2017-08-01
#736 - Show/hide completed works

tbl73:


saracarl:
This is not surviving testing in the real world.  Even the Julia Brumfield diaries, which are heavily indexed, have pages that don't have subject linking.  (At least one.)  I'm not sure what we should do about that.  

I spent time trying to get 1918 and 1919 to a "complete" state, and gave up because (for instance) it doesn't make sense to mark pages that exist but are missing images as blank.  But there's not transcript and no indexing, so it won't ever be complete.

@benwbrum this needs your eyes.

saracarl:
Ok.  Ben and I talked about this and (unfortunately!) the right logic for "done" is page is transcribed or marked blank.  Indexing can't be part of the deal.

Can you also document in the issue where we are hiding the works and where we are showing them?

tbl73:
Changing the setting to consider transcribed but not indexed works as complete does mean that by default, works will be hidden if they are transcribed but not indexed.  The owner of the collection can change that default, but it's definitely something we'll need to let people know, if they're big into the indexing.  (Yaquina Lights definitely has works that are at 100% for both indexed and transcribed, though I realize that may be unusual.)

Issue #243, opened 2015-09-20
finding a new user is hard (admin)


saracarl:
For the admin to find a user and make them an owner, you have to scroll to the last page (26 as of today) to find the last created user (assuming that's the one you are changing.)  We need a search screen to find a user.

PR #768, opened 2017-09-07
#243 - Find users in admin console

tbl73:


saracarl:
Tested manually -- works really nicely!

Issue #697, opened 2017-07-11
Assign Categories does not DISTINCT subjects


benwbrum:
Transcribe a new page and tag the same subject twice:  `I saw [[John Smith]] today.  [[John Smith|Mr Smith]] said he was doing well.`

This should create a single subject and ask the user to categorize `John Smith`.  However, the user is asked to categorize the new subject twice.

Issue #675, opened 2017-06-16
Subjects - pages for a subject


tbl73:
The code for displaying the number of pages referring to a subject appears to show each page_article_link for a subject rather than displaying distinct pages, which is confusing (when it's multiple links for the same page).

Issue #512, opened 2017-01-11
subject links to translations


saracarl:
If a subject is linked in a translation, clicking on the reference to the subject should take you to the translation view of the page rather than the default page.

So you can have 2 links to the same page, one for the transcribed instance of the subject and another for the translated instance of the subject.  Both take you to the same view, and that didn't seem quite right.

PR #773, opened 2017-09-11
Issues #512, #675, #697 - distinct subjects

tbl73:


saracarl:
Verified #697 assign categories does not DISTINCT subjects.

saracarl:
Ok.  the solution to #512 looks good -- the page link takes you to *either* the transcription or translation, which is great.

The UI on the subject page, however, is confusing.  I see the page listed twice (once for transcription and once for translation) with no indication of the difference.  Can we add a parenthetical (translation) after the page name for the translation page link?

saracarl:
tests passed.

app/views/article/_article_links.html.slim
```diff
@@ -1,11 +1,12 @@
 -unless @article.show_links(@collection).empty?
-  h5 ="#{pluralize(@article.show_links(@collection).size, 'page')} refer to #{@article.title}"
+  h5 ="#{pluralize((@article.show_links(@collection).count).count, 'page')} refer to #{@article.title}"
```
benwbrum: I don't understand the second `.count` here.  Won't that blow up?

tbl73: I know, it looks totally bizarre.  Article.show_links now returns an active record relation on the page_article_links that's grouped by text_type and page number (so that it doesn't matter how many times it's on the page).  The first count gives you a hash of the the grouped relation, with the group as keys and the count as values (i.e., {["transcription", 26895]=>2, ["translation", 26895]=>1, ["transcription", 26899]=>1}).  And the second count gives you a count of the key/value pairs in the hash.  So it's weird because of the grouping, but it does use both counts.  Although I just double-checked, and we could use size instead of the second count, which might be a better choice for consistency with the other counts we have in the UI.

Issue #303, opened 2016-04-24
Email error&transcript when page save fails


benwbrum:
I'd lke us to recieve email whenever a user tries to save a page and gets an error instead.

We should be able to modify the code in https://github.com/benwbrum/fromthepage/blob/master/app/controllers/transcribe_controller.rb#L143-L205 to append the `log_attempt` message to an instance variable, then have `log_failure` or `log_error` (or whatever it's called) email the sysadmins.

PR #774, opened 2017-09-11
#303, email page save errors

tbl73:


saracarl:
So this would send ~60 emails in the last 6 months in production.  Not a horrible amount, although I have concerns it will all be "user" error -- i.e. unclosed brackets, etc.  We'll probably want a way to turn them off at some point in the future.

Issue #758, opened 2017-08-24
deleted user's notes breaks stuff


saracarl:
F, [2017-08-24T09:15:22.558214 #17287] FATAL -- : 
ActionView::Template::Error (undefined method `display_name' for nil:NilClass):
    77: 
    78:     h3 Recent Notes
    79:     -@collection.notes.limit(10).each do |note|
    80:       -user = link_to note.user.display_name, user_profile_path(note.use
r)
    81:       -text = link_to truncate(note.title, :length => 100), collection_d
isplay_page_path(@collection.owner, @collection, note.page.work, note.page)
    82:       .deed-short.small
    83:         =time_tag(note.created_at, class: 'legend legend-left')
  app/views/collection/show.html.slim:80:in `block in _app_views_collection_show
_html_slim__1578740875081384648_70303786501100'
  app/views/collection/show.html.slim:79:in `_app_views_collection_show_html_sli
m__1578740875081384648_70303786501100'

saracarl:
, [2017-08-24T09:09:45.434586 #17287] FATAL -- : 
ActionView::Template::Error (undefined method `id' for nil:NilClass):
    1: .user-bubble.editable(id="note-#{note.id}")
    2:   =link_to({ :controller => 'user', :action => 'profile', :user_id => not
e.user.id }, class: 'userpic')
    3:     =gravatar_image_tag note.user.email, :alt => note.user.display_name
    4:   .user-bubble_content
    5:     -if note.user == current_user
  app/views/notes/_note.html.slim:2:in `_app_views_notes__note_html_slim___39816
55073463191156_70303821692360'
  app/views/notes/_notes.html.slim:8:in `block in _app_views_notes__notes_html_s
lim__3440401845444989744_70303786098420'
  app/views/notes/_notes.html.slim:7:in `_app_views_notes__notes_html_slim__3440
401845444989744_70303786098420'
  app/views/display/display_page.html.slim:72:in `_app_views_display_display_pag
e_html_slim___479179666841464208_70303821822420'

benwbrum:
This also breaks on http://97.107.141.53/library/john-torrey-papers/beck-lewis-c-and-john-torrey-correspondence-1819 with the following error:
```

I, [2017-08-25T11:58:07.521673 #26978]  INFO -- : Started GET "/library/john-torrey-papers/beck-lewis-c-and-john-torrey-correspondence-1819" for 70.112.51.219 at 2017-08-25 11:58:07 -0400
[...]

PR #778, opened 2017-09-14
Delete dependent records on user deletion #758

saracarl:
You can assign this to me once the merge conflicts are resolved.


saracarl:
failing tests:

Failures:

  1) convention related tasks reverts to collection level transcription conventions
     Failure/Error: expect(convention_work.transcription_conventions).to eq nil
     
       expected: nil
            got: "Work level transcription conventions"
     
       (compared using ==)
     # ./spec/features/zz_convention_spec.rb:89:in `block (2 levels) in <top (required)>'

tbl73:
That may or may not be a real failure.  I haven't looked at what the code
is supposed to be doing (so it could be), but that particular jquery test
is brittle and sometimes just fails.  I haven't figured out how to resolve
the issue (partly because I don't understand why sometimes it fails and
sometimes it passes without any code changes).

saracarl:
tests passed the 2nd time.  I think this ends up being a timing issue where that spec test fails.

PR #779, opened 2017-09-14
Customer styling

tbl73:
Edits that give the customer the ability to change the color scheme and dashboard wording.  This also includes the splash page from the master branch (rather than from fromthepage.com).


app/assets/stylesheets/application.scss
```diff
@@ -1,3 +1,5 @@
+@import "base/ftp_color_scheme";
+
```
saracarl: Does this need to include an import of customer_color_scheme as well?

app/assets/stylesheets/base/ftp_color_scheme.scss
```diff
@@ -0,0 +1,58 @@
...
+
+$headerHeight:              80px;
+$headerFg:                  $themeDark;
+$headerBg:                  $themeLight;
+
+$footerHeight:              80px;
+$footerFg:                  mix(#FFF, $themeDark, 50);
+$footerBg:                  $themeDark;
+
+@import "base/customer_color_scheme";
```
saracarl: Ah, ok.  I see it being imported here.

PR #780, opened 2017-09-14
775 comment spam

benwbrum:


saracarl:
Fails this test:

  1) URL tests edits a user slug
     Failure/Error: expect(page).to have_field('user[slug]', with: @user.slug)
       expected to find field "user[slug]" with value "eleanor" but there were no matches
     # ./spec/features/url_spec.rb:151:in `block (2 levels) in <top (required)>'

Issue #772, opened 2017-09-09
IIIF Collection Import


benwbrum:
We have agreed to implement import of entire IIIF collections for Stanford University Archives in collaboration with @blalbrit, who will be overseeing creation of the collection manifests  

## Goals
* Reduce work load for project owners importing collections -- Importing collection `A-F` is easier than importing item manifests `A`, `B`, `C`, `D`, `E`, and `F` one at a time.
* Reduce potential for accidentally missing an item or importing the same item twice -- Importing collection `A-F` is less likely to result in `(ABCEF)` or `(ABCDDEF)`
* Allow incremental imports of collections, as when `(ABCD)` was imported, then had items `E` and `F` added to the collection manifest.
* Allow imports of items from more than one IIIF collection manifest into the same collection.
* Allow imports from one IIIF collection into more than one FromThePage collection.

## Out of scope
* Importing collections of collections is out of scope.  Only the item manifests listed in the imported collection manifest will be imported; sub-collections will not.  (This should be made explicit to the user.)

# User Interface
We will create a new collection import screen (or modify the existing collection view).  This will be accessed when
* A project owner navigates to a collection that contains manifests starting from _Import from a IIIF Repository_, or
* A project owner manually enters a collection manifest instead of an item manifest in the _Import an IIIF Manifest_ field on the _Start a Project_ screen.

The collection import screen will
* List each item manifest for the collection
* If that item manifest has already been imported into the system, it will warn the user that the item has been imported, listing the collection it was imported into, the owner of that collection, and linking to the imported work.  It is probable that a double-import is a mistake, but it's also possible that another user has imported a manifest for an unrelated project (e.g. a classroom project), and that a double-import is desirable.
* Display a checkbox next to each item--checked by default unless the item has been imported elsewhere--allowing the project owner to decide whether to include or exclude an item from the import.
* Display a "check/uncheck all" box to minimize clicking.
* Present a target collection selector, including the ability to create a new collection from the IIIF Collection manifest data.
* Display some big button to actually do the import.

# Import
The import should be performed in the background by a rake script, which will need to take the target collection and the list of IIIF item manifest IDs to import into it.  Upon completion, the script should email the project owner with a success message and a link to the imported collection.

# Open Questions

* Should the new collection import screen and the existing collection display screen be merged?
* Should item manifest imports also be run in the background?  (At present, the delay is not a big problem, and the user experience of being directed straight to the imported work is rewarding.)

saracarl:
[...]

PR #789, opened 2017-09-22
#772 IIIF collection import

tbl73:
There are a couple of outstanding issues:

- We decided to wait on showing whether the manifest has already been imported in the checkbox list.

- The IIIF content search is currently having issues, so I can't test it.  There might be more dead code to take out, but I don't want to do that until we've been able to test everything.

- Spec tests are a slight problem.  I fixed our current tests, so we can test individual manifest imports.  I'm having a little trouble with the collection import as a test, though, for two reasons.  The first is jquery-related - I was trying to get it to select only certain manifests, and I can't get the checkboxes (which trigger jquery) to actually do what they should within the test. (The test passes, but it's not actually working.)  I can get it to import an entire collection, but I don't know if that's the right choice unless we can find a really small one. Because that's the other part I'm not entirely sure about with this test - since importing is a rake task, we'd need to put in some kind of sleep before checking to determine if the manifests have actually been imported, which would potentially add rather a lot of time to the test suite.  We need to determine the best method of testing.


benwbrum:
Something strange is going on with URL encoding when you try to import a collection by pasting its url in to the field, if the URL has a space in it.  

From _Start A Project_, add this url: 
`http://wellcomelibrary.org/service/collections/topics/Alcoholic%20beverages/`

The application errors with:
```
Started POST "/sc_collections/import" for 127.0.0.1 at 2017-09-25 08:20:10 -0500
Processing by ScCollectionsController#import as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"g5Ax+cAEq1qLvG5LmQwszRzdUvGn+znavouQYjPlgtM=", "at_id"=>"http://wellcomelibrary.org/service/collections/topics/Alcoholic%20beverages/", "button"=>""}
  PageBlock Load (0.3ms)  SELECT `page_blocks`.* FROM `page_blocks`  WHERE `page_blocks`.`controller` = 'sc_collections' AND `page_blocks`.`view` = 'import'
  User Load (0.3ms)  SELECT  `users`.* FROM `users`  WHERE `users`.`id` = 2  ORDER BY `users`.`id` ASC LIMIT 1
Completed 500 Internal Server Error in 8ms

URI::InvalidURIError (bad URI(is not URI?): http://wellcomelibrary.org/service/collections/topics/Alcoholic beverages/):
  app/models/sc_collection.rb:38:in `collection_for_at_id'
  app/controllers/sc_collections_controller.rb:36:in `import'
```

The same problem does not happen if you paste in `http://wellcomelibrary.org/service/collections/topics/Alcohol`, which does not have a space.

benwbrum:
Passing in an invalid URL like `I am a bad URL` results in a TemplateMissingException:

```
Started POST "/sc_collections/import" for 127.0.0.1 at 2017-09-25 08:25:56 -0500
  ActiveRecord::SchemaMigration Load (0.3ms)  SELECT `schema_migrations`.* FROM `schema_migrations`
Processing by ScCollectionsController#import as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"g5Ax+cAEq1qLvG5LmQwszRzdUvGn+znavouQYjPlgtM=", "at_id"=>"I am a bad URL", "button"=>""}
  PageBlock Load (0.2ms)  SELECT `page_blocks`.* FROM `page_blocks`  WHERE `page_blocks`.`controller` = 'sc_collections' AND `page_blocks`.`view` = 'import'
  User Load (0.2ms)  SELECT  `users`.* FROM `users`  WHERE `users`.`id` = 2  ORDER BY `users`.`id` ASC LIMIT 1
Completed 500 Internal Server Error in 67ms

ActionView::MissingTemplate (Missing template sc_collections/import, application/import with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :slim]}. Searched in:
  * "/home/benwbrum/dev/products/fromthepage/fromthepage/app/views"
  * "/home/benwbrum/.rvm/gems/ruby-2.2.0@fromthepage_rails_3_from_github/gems/devise-3.4.1/app/views"
```

benwbrum:
The Yale Center for British Art photograph collection manifest `http://manifests.britishart.yale.edu/collection/photographs` gives the following error when I try to import it from the _Start a Project_ screen:
```
Started POST "/sc_collections/import" for 127.0.0.1 at 2017-09-25 08:39:17 -0500
  ActiveRecord::SchemaMigration Load (0.3ms)  SELECT `schema_migrations`.* FROM `schema_migrations`
Processing by ScCollectionsController#import as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"g5Ax+cAEq1qLvG5LmQwszRzdUvGn+znavouQYjPlgtM=", "at_id"=>"http://manifests.britishart.yale.edu/collection/photographs", "button"=>""}
  PageBlock Load (0.7ms)  SELECT `page_blocks`.* FROM `page_blocks`  WHERE `page_blocks`.`controller` = 'sc_collections' AND `page_blocks`.`view` = 'import'
  User Load (0.2ms)  SELECT  `users`.* FROM `users`  WHERE `users`.`id` = 2  ORDER BY `users`.`id` ASC LIMIT 1
Completed 500 Internal Server Error in 478ms

NoMethodError (undefined method `[]' for nil:NilClass):
  app/controllers/sc_collections_controller.rb:27:in `import'
```

Strangely, it does not give the same error if I navigate to it by _Explore_, then _Yale Center for British Art_, then (wait forever because their manifest is huge) click _Photography_.  Possibly this has something to do with the way we try to construct breadcrumbs when there is no `within` element in the manifest?

saracarl:
passed the automated test suite.

benwbrum:
On the collection chooser drop-down, how about "Select a target collection" or "Select a collection to import into", rather than "Select an import collection"?

tbl73:
On the issue with the space in the URL - the code had a bunch of CGI::unescape calls that I didn't change - if I take that out, it works fine.  So I guess the question is, do the urls need to be unescaped anywhere, or should I just be taking those calls out of the controller?

benwbrum:
On the URL escaping, I think he reason that the code had all those `unescape` calls was me struggling to look up existing manifests in the database for the `iiif/for` API in `iiif_controller`.  So there was lots of flailing going on.

Thinking harder, I suspect that we should settle on a consistent internal representation of `@id` URIs within the application.  But should those be escaped or unescped?  Perhaps the most sensiblle thing is to store escaped URLs, look up escaped URLs, and assume we're working with escaped URLs in all cases (except perhaps user-friendly inputs [i.e. if the URL passed in is invalid, try escaping it])?

What do you think, @tbl73 and @saracarl?

benwbrum:
Internally, it looks like IIIF manifests escape all URLs:
See the Wellcome's list of topics, which includes URIs with spaces:
```
    {
      "@id": "https://wellcomelibrary.org/service/collections/topics/Abdominal%20Cavity%2C%20abnormalities/",
      "@type": "sc:Collection",
      "label": "Abdominal Cavity, abnormalities"
    },
```

So let's assume everything we're handed is escaped correctly, and get rid of my old CGI.unescape cruft.

saracarl:
For automated testing -- here's a collection with just 3 works:  

https://wellcomelibrary.org/service/collections/topics/Abdominal%20aorta/ 

And here's one with just 1 work:
https://wellcomelibrary.org/service/collections/topics/Abdominal%20aneurysm/

tbl73:
Sounds good on the assuming things are escaped correctly.  I did notice things seemed to be working without it.

tbl73:
The Wellcome library manifests don't actually work, unfortunately - there's an error from the IIIF::Presentation gem that it can't parse the manifests.  it seems to be an issue with the gem, rather than on our end. (E-codices also has the same error.)  TextGrid has a collection with just 1 work that I think might be good: https://textgridlab.org/1.0/iiif/manifests/collection/published.json

benwbrum:
If you go to *Start a Project* and enter `http://norman.hrc.utexas.edu/notDM/objectManifest/p15878coll64/41` for the Manifest URL, the user gets an error saying "Please enter a valid IIIF manifest URL".

This manifest passes validation, and I don't see an exception in the logs, so I don't know what's going on.   (Note that this is a manifest the Harry Ransom Center is using to test the tool, but their implementation is new.)

benwbrum:
The same problem occurs with the Gallica manifest at `http://gallica.bnf.fr/iiif/ark:/12148/btv1b105243418/manifest.json` -- if I paste it in, it gives me an Invalid URL error.

benwbrum:
It's not clear to me when an import is done.  Should we email the work owner?  (Perhaps we're already doing so and I just have SMTP disabled on my desktop?)

tbl73:
We're sending an email on success (or failure, for that matter), but it only works if SMTP is enabled.  I'll check into those two manifests and see what I can see.

tbl73:
It looks like in both cases, they don't have a "within" field, which I was using to figure out what the parent at_id/collection is.  Apparently I need to use something else, though, since clearly "within" doesn't always work - any suggestions?

benwbrum:
Unfortunately, `within` is an optional element within IIIF, and efforts to make it required have been rejected.  (Admittedly, for some decent reasons.)

As a result, we're going to have to handle the case of a user importing a parentless manifest.  In this case I suspect we won't suggest a collection for them, but just have them choose or create a FromThePage collection.

tbl73:
We actually do already handle parentless manifests - it was more the check
for the parent that I (apparently) need to fix.  But if within is optional,
then I'll test for that first, then separately get the @id.  (I was testing
for the @id inside of within, which was why the code was failing.)

tbl73:
It looks like the failure is actually happening on something I'm checking in the jquery, so I'm trying to track down what it should say instead.

saracarl:
tests passed.

saracarl:
one more change -- can we swap these two UI bits?  It makes more sense to me for the "select all" to be immediately preceding the list of manifests. 

image

tbl73:
Just pushed it up.  (And I merged in the new ui-design code, just FYI.)

Issue #786, opened 2017-09-20
Escaped HTML in IIIF manifest labels


tbl73:
Some IIIF manifest labels include escaped HTML and therefore are being imported into FtP works with titles that include the escape codes.  Best practice is not to have HTML in label, but we need to handle the possibility, since we don't have control over content providers.

benwbrum:
Let's assume that `label` strings are not HTML, but attempt to handle this gracefully.

When we convert an `sc_manifest` to a `work`, let's wrap the line setting `work.title = sc_collection.label` in a new `cleanup` function which tests for escaped HTML and unescapes it if detected.  (Searching for `&amp;` and `&quot;` will probably be good enough.

I've worked in databases that stored all strings as escaped HTML, and *NEVER AGAIN*.

tbl73:
I put in a function that does a gsub for &amp; &apos; and &quot; which converts the labels for both manifests and collections (depending on whether you're creating a work or a collection).

PR #796, opened 2017-10-04
#786 - escaped HTML in IIIF manifest labels

tbl73:


saracarl:
passed automated tests and manual.  dependent on IIIF-collections so not merging until after we merge that in.

PR #807, opened 2017-10-13
Update mysql2, add tzinfo-data gem dependencies (fixes #349)

### LINKED ISSUE #349: Mysql2 gem version 0.3.16 doesn't support MySql Server 5.7
The gemfile specifies the  0.3.16 version of mysql2 gem, which doesn't work with MySql Server 5.7, the current version, which comes with Ubuntu 16.04.  I was able to get the gem installed, but rails won't start.  It comes up with a gem load error: uninitialized constant Mysql2::Client::SECURE_CONNECTION (NameError).  There are two issues listed in the mysql2 github that state that you need to upgrade mysql2 to a higher version to run with MySql Server 5.7.

bencomp:
To allow FromThePage to work with MySQL 5.7, the mysql2 gem is updated. The `tzinfo-data` gem was recommended on <https://github.com/tzinfo/tzinfo/wiki/Resolving-TZInfo::DataSourceNotFound-Errors> and helped resolve the error described on that page that I got when building FromThePage in Docker.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'capistrano-rails', '= 1.1.3'
 gem 'capistrano-bundler', '~> 1.1.2'
 gem 'jquery-rails'
 
-gem 'mysql2','0.3.16'
+gem 'mysql2','0.3.21'
 
 gem 'recaptcha', '0.3.6'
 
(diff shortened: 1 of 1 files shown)


bencomp:
Should I also update Gemfile.lock when updating Gemfile, @benwbrum?

saracarl:
Need to test with Server version: 10.0.32-MariaDB-1~xenial mariadb.org binary distribution, which is what we're running in production.

saracarl:
Closing this -- we made the change in the gemfile.  Opened Issue #1024 for the tzinfo gem addition.

PR #817, opened 2017-10-18
815 external identifier

benwbrum:


tbl73:
@benwbrum - Nevermind, I just saw that a new issue was added to add the identifier to the exports.

Issue #835, opened 2017-10-25
TCMI owner dashboard works count wrong


benwbrum:
The TCMI owner dashboard which is shown to the user `childrensmuseum`displays as `0` even though the owner has thousands of works.

benwbrum:
![owner_dashboard_works__count](https://user-images.githubusercontent.com/199961/32016554-df7e886a-b989-11e7-9f0a-fe19e8318d25.png)

tbl73:
This seems to have gotten broken at some point, and no one realized it.  It's actually wrong in production, too.  Just pushed up a PR with a fix.

PR #836, opened 2017-10-26
#835 incorrectly listing owned works

tbl73:


tbl73:
Note that this is compared to UI-design - should there be one for indianapolis?

saracarl:
Yes, please.  Can you do a PR against indianapolis as well (and make sure it doesn't mess up the styling...).

PR #841, opened 2017-10-26
Render last deed only when it exists (fixes #840)

### LINKED ISSUE #840: Admin page returns 500 Internal Server error when no deeds exist
This is a minor issue, but it took me a while to figure out why on first setup I got this Internal Server error for the admin overview.

The problem is that before a deed exists, the last line in `views/admin/index.html.slim`

```
h4.acenter="Last deed logged #{time_ago_in_words Deed.last.created_at} ago by #{Deed.last.user.display_name}"
```

results in an error. I found a solution in adding a condition before this. PR is almost ready :)

bencomp:
As described in #840, the Admin overview tries to render the most recent Deed's creation, which fails right after setup, when no deeds exist. This results in an Internal Server error.


saracarl:
Reviewed, tested (with our new continuous integration tests!) & merged.

Thanks for contributing.

Issue #865, opened 2017-11-16
"Pages that need transcription" button


saracarl:
Can we add a "Pages That Need Transcription" button right before the "Pages that need review" on a work view?

tbl73:
Also added pages that need translation.

PR #871, opened 2017-11-17
#865 Pages that need transcription

tbl73:


saracarl:
I have a scenario where page 1 is not transcribed, but pages 2-3 are, then 4 isn't, then 5 is. then 6+ isn't.  The "show pages needing transcription" works for 4, 6+, but misses page 1.  Can we test this some more, please.  Page 1 is not marked as needing review or blank.

tbl73:
Right now the "needs transcription" pages are the ones that have no status - can you check and see what the page status for page 1?  It would help to figure out why it isn't showing up.  (It's possible that I need to check for some other status, but I don't know which one.)  Also, if you're in the work - overview (with the list of pages), does that page show up as "this page has not been transcribed"?

PR #896, opened 2017-12-30
FromThePage IIIF API support, v2

benwbrum:
Implement the second revision of support for IIIF, including IIIF-based API access to all contributions made for a project within a time window.


app/models/deed.rb
```diff
@@ -13,6 +13,21 @@ class Deed < ActiveRecord::Base
...
+    NOTE_ADDED,
+    PAGE_TRANSLATED,
+    PAGE_TRANSLATION_EDIT,
+    OCR_CORRECTED,
+    NEEDS_REVIEW,
+    TRANSLATION_REVIEW,
+    TRANSLATION_INDEXED
+  ]
+
+
```
saracarl: Didn't we already have CONTRIBUTOR_DEED_TYPES ?  Why did we have to add it here?

app/controllers/export_controller.rb
```diff
@@ -94,6 +94,48 @@ def export_all_works
...
+  end
+
+  def work_plaintext_translation_emended
+    render  :layout => false, :content_type => "text/plain", :text => @work.emended_translation_plaintext
+  end
+
+  def work_plaintext_searchable
+    render  :layout => false, :content_type => "text/plain", :text => @work.searchable_plaintext
+  end
+
```
saracarl: Please open issues to consider adding these extra export formats to the Export tab.

app/models/work_statistic.rb
```diff
@@ -55,6 +55,23 @@ def pct_translation_blank
...
+  def pct_transcribed_or_blank
+    pct_blank + pct_transcribed
+  end
+  
+  def pct_translated_or_blank
+    pct_translation_blank + pct_translated    
+  end
+  
+  def pct_transcribed_or_blank_not_review
+
```
saracarl: Um.  This is a blank method?

app/models/work_statistic.rb
```diff
@@ -55,6 +55,23 @@ def pct_translation_blank
...
+  def pct_translated_or_blank
+    pct_translation_blank + pct_translated    
+  end
+  
+  def pct_transcribed_or_blank_not_review
+    
+  end
+  
+  def pct_translated_or_blank_not_review
+
```
saracarl: This is a blank method?

app/controllers/export_controller.rb
benwbrum: Done -- see #898

app/models/deed.rb
benwbrum: Not that I was able to find.  Where have you seen it?

tbl73: I don't think we have contributor_deed_types, but deeds are already associated with user_ids.  What are these new contributor_deed_types used for?

Issue #781, opened 2017-09-15
Voice Recognition in text fields


saracarl:
We need to add voice recognition to the transcript, translate, notes, and subject articles field.  

For research:  https://www.labnol.org/software/add-speech-recognition-to-website/19989/

https://developers.google.com/web/updates/2013/01/Voice-Driven-Web-Apps-Introduction-to-the-Web-Speech-API

tbl73:
This appears to be Chrome-specific, just FYI.  Also, it only works with an https server now (apparently - it won't even ask me if it can access the microphone), so I'm not going to be able to do a whole lot with it unless/until I get my dev environment set up with SSL.

PR #912, opened 2018-01-17
#781 Voice recognition/dictation in text fields

tbl73:


saracarl:
We need to turn this off for field based transcription.  If you have field based enabled, don't show (or grey out) the checkbox.  If you have voice dictation enabled, turn if off when you set up field based transcription.  (a warning would be good, too.)

Unless I'm wrong and this actually works?

saracarl:
Let's get rid of the "Click microphone to dictate" words and just have the microphone.  I think it's enough.

tbl73:
I think you're right about the field-based transcription - it wouldn't work for that as currently written.  Not entirely sure I agree about the wording next to the microphone, but I can get rid of it.

PR #926, opened 2018-01-26
support for single sign on -- google oauth2 and saml

saracarl:


tbl73:
The require_relative 01fromthepage.rb in the devise initializer should be removed, since it was added while debugging something.

tbl73:
Right now, SAML sign in is set to true in the initializer and Google sign in is set to false - did you mean for that to be shipped that way or the reverse?
Also, when I click the sign in with Google link, I get the error shown below:
image

saracarl:
fixed the require, turned the saml auth to false.

Issue #923, opened 2018-01-24
Add email address identifier in drop downs to distinguish similar names


tbl73:
Display names aren't currently distinct.  If you add an owner or collaborator to a collection, you can't distinguish between two users with the same display name.  We need to add email address as an identifier.

PR #942, opened 2018-02-05
#923 - Add email identifier to user select lists

tbl73:


saracarl:
So the behavior is different on "allowed collaborators" (on a work) and on "add an owner" (on a collection).  the allowed collaborators one works much better than the add an owner one.  Any idea why?

tbl73:
How is the behavior different?  (What does "works much better" mean in this context?)

saracarl:
The filtering works a lot better on collaborators.  If I type "ben" then I get a list of users that match ben in the string.  That doesn't happen on owners.

tbl73:
Huh. I thought that worked for me on owners.  I'll take another look at it.

tbl73:
That's strange - if I try typing Ben into the owners, I get the same users that I get in allowed collaborators, although they are in a different order.  Are you not getting a filtered list at all on owners?

tbl73:
Okay, we figured out that the problem was with field-based transcriptions, which don't offer voice transcriptions.  The jQuery on the field-based settings was looking for a selector that doesn't exist (for voice-based transcription), which was then making the select2 jquery code (that those dropdown boxes use) fail.  I made a couple of changes - see if it's working for you now.

saracarl:
works!  yay!

Issue #566, opened 2017-02-26
option to align text to the right


saracarl:
For Arabic, for example.

benwbrum:
Needs more definition.  I did a write up a couple of years ago in an email.

benwbrum:
This is suddenly of high importance.

We need to allow administrators to configure projects to use a specific language.  Eventually languages may vary between works within the same collection (e.g. letters in Polish and Yiddish from different correspondents may appear in a 19th century Eastern European's papers) or within a particular text, but at the moment we will focus on single-language collections and on alignment/justification for right-to-left languages.  If the language of a text is a right-to-left (RTL) language, we should add the appropriate tags to HTML forms, paragraphs, divs and spans to support proper text justification.

To support this feature:

- A project owner should be able to select a langage on the collection settings screen, using a drop-down populated with ISO-639 language names and storing the code in a new column on `collections`.  See the [ISO-639 gem] (https://github.com/alphabetum/iso-639)
- When displaying text (page overview, read work, subject article, possibly subject list screens) we should add `lang` and `dir` attributes to the [relevant HTML elements](http://sites.psu.edu/symbolcodes/web/rtl/).  The [rtl gem](https://github.com/abarrak/rtl) may be useful here.
- When transcribing text, the transcription form should be justified appropriately, also using `lang` and `dir` attributes.
- When exporting text as TEI, the `profileDesc` should have a [new `langUsage` block](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-langUsage.html) containing a `language` element with the appropriate code.

benwbrum:
We should test the experience of adding subject links, as entering LTR characters `[[` and `]]` within RTL text can cause the cursor to jump around unnervingly.

saracarl:
Let's spin the TEI export and subject linking out to a 2nd issue.  It's not a requirement for the customer who needs this RSN.  (We should still do it, but I'd like to deploy the first half first.)

PR #951, opened 2018-02-16
#566 - support for RTL languages

tbl73:


tbl73:
This work doesn't include exports or the process of subject linking - see #952

benwbrum:
The RTL code works!  Better, it fixes the problems associated with wikilink data entry, and the display and data entry for text and subjects all works.

It does need a few revisions, based largely on clarification from the client.

To change:
ISO-639-2 instead of ISO-639-1 in the `collection_controller` so that we can pick up support for obscure and dead languages.  (I've tested this and the `RTL.rtl?` method works fine with ISO-639-2.)

HTML attributes to be created by a view helper, so that repetition of sections like:
```
direction = Rtl.rtl?(@collection.text_language) ? 'rtl' : 'ltr'
table.form lang="#{@collection.text_language}" dir=direction class=direction
```
becomes
`table.form language_attrs`

Due to clarification from the client that (presently) instructions, description, discussion, and metadata about the text is to be in English, the HTML attributes should be stripped from:
The collection edit screen
The collection show screen (Including the work listing)
And also (apparently) the note entry for each page.

It would be lovely if this last bit were done in a single commit, as I can see us backing it out or conditionalizing it in the future as new multi-lingual projects come online.

app/views/static/transcribe_help.html.slim
```diff
@@ -1,7 +1,8 @@
 =render({ :partial => '/shared/work_tabs', :locals => { :selected => 9, :work_id => @work.id }})
 
 .columns
-  article.maincol
+  -direction = Rtl.rtl?(@collection.text_language) ? 'rtl' : 'ltr'
```
benwbrum: Transcription help needs to be un-conditionalized, as the project plans to list its instructions in English.

app/views/transcribe/translate.html.slim
```diff
@@ -68,7 +68,8 @@
       -if @preview_xml
         .page-preview ==xml_to_html(@preview_xml)
 
-      .page-editarea(style=("display:none" if @preview_xml))
+      -direction = Rtl.rtl?(@collection.text_language) ? 'rtl' : 'ltr'
+      .page-editarea(style=("display:none" if @preview_xml) lang="#{@collection.text_language}" dir=direction class=direction)
```
benwbrum: On the assumption that most translation target languages continue to be LTR (mostly English, but likely Spanish soon), we should deconditionalize this.  When we need to support RTL target languages for translation, we'll need to create a new `target_language` field on `collections` and drive the logic off of that, rather than the source text.

saracarl:
The setts for a work are still showing up right to left.  (i.e. http://localhost:3000/saracarl/topic-abdominal-aorta/fromthepagestructureddataproposal11-28-17-odt/edit )  

image

(and all the text fields below the screenshot)

app/helpers/application_helper.rb
```diff
@@ -130,4 +128,12 @@ def pontiiif_server
     Rails.application.config.respond_to?(:pontiiif_server) && Rails.application.config.pontiiif_server
   end
 
+  def language_attrs(collection)
+    collection = collection
```
benwbrum: Redundant?

tbl73: Whoops - totally redundant.  I was using an instance variable initially, and then forgot when I switched it.

benwbrum:
The remainder looks good.  Let's merge once the work settings change is made.

Issue #947, opened 2018-02-08
Email users overnight about collection activity


tbl73:
We want to send users an email overnight for certain activity in the collection:

1. If a work or works are added to a collection they've worked in.
2. If a page that they've transcribed/edited is edited by someone else.

This should all be included in one email, and only sent to users with activity (who haven't turned these emails off).

PR #954, opened 2018-02-19
#947 Overnight user activity emails

tbl73:
These are the emails that a user receives if activity has occurred on pages/collections they've worked in.  Note that I can't really write any new features tests for this work, because it's a rake task and isn't triggered that way.


app/views/user/update_profile.html.slim
```diff
@@ -29,5 +29,28 @@
       tr
         th =f.label :location
         td =f.text_field :location, :placeholder => 'Where are you from?'
+
+    p.email_notifications Select which of the following notifications you want to receive:
+    table.form.email_notifications
```
saracarl: Where is the "Nightly emails on collections you work on" option?  (and let's list that first.)

app/views/user_mailer/nightly_user_activity.html.slim
```diff
@@ -0,0 +1,24 @@
...
+    p
+p
+-unless @active_note_pages.blank?
+  -@active_note_pages.each do |page|
+    =link_to "Click here", collection_transcribe_page_path(page.collection.owner, page.collection, page.work, page, only_path: false)
+    |&nbsp; to view the note on page #{page.title} in #{page.collection.title}: #{page.work.title}, which has been added since you worked on it.
+    p
+
+p Thanks, 
+|FromThePage
```
saracarl: We need a link to the settings where you adjust these emails at the end of all of the emails.

app/views/user/update_profile.html.slim
```diff
@@ -29,5 +29,28 @@
...
+        td =f.check_box("notifications[add_as_owner]", {checked: @user.notification.add_as_owner})
+        th =f.label :add_as_owner, "Added to private work as an owner"
+      tr
+        td =f.check_box("notifications[note_added]", {checked: @user.notification.note_added})
+        th =f.label :page_edited, "Note added in response to a comment you made"
+      tr
+        td =f.check_box("notifications[work_added]", {checked: @user.notification.work_added})
+        th =f.label :work_added, "Work added to collection you've worked on"
+      tr
+        td =f.check_box("notifications[page_edited]", {checked: @user.notification.page_edited})
```
saracarl: This one needs to be unchecked by default.

tbl73: Which one needs to be unchecked by default?  There are a couple of different lines highlighted here.

saracarl: page edits.  I don't think you want all page edits emailed to you right away by default.

tbl73: Well, they aren't being immediately emailed under any circumstances - they're in the overnight emails.  Though you may still not want them sent to you by default; what do you think?

app/views/user/update_profile.html.slim
tbl73: We don't have one of those, and we should - I think originally it was only page edits being sent overnight, and when we added additional information, I didn't think about it.

tbl73:
We had some further discussion about the goals of this work, and this needs some changes - closing the PR and will re-open once it's ready.

Issue #960, opened 2018-02-26
iiif manifest 500 error


saracarl:
This page, and many pages, are causing 500 errors when requesting the IIIF manifest because we're expecting a translation and there is none (or something similar):

https://fromthepage.com/iiif/335/manifest

I, [2018-02-26T15:42:14.485865 #12159]  INFO -- : Completed 500 Internal Server Error in 40ms
F, [2018-02-26T15:42:14.487495 #12159] FATAL -- : 
NoMethodError (undefined method `translation' for #<Page:0x007f18b054e730>):
  app/controllers/iiif_controller.rb:680:in `status_service_for_page'
  app/controllers/iiif_controller.rb:581:in `add_services_to_canvas'
  app/controllers/iiif_controller.rb:522:in `canvas_from_iiif_page'
  app/controllers/iiif_controller.rb:404:in `block in iiif_sequence_from_work_id'
  app/controllers/iiif_controller.rb:402:in `iiif_sequence_from_work_id'
  app/controllers/iiif_controller.rb:181:in `manifest'

jbhoward-dublin:
Note that this problem cascades to the iiif/for API as well, for example:

https://fromthepage.com/iiif/for/https://data.ucd.ie/api/img/manifests/ivrla:3849

saracarl:
Fixed and deployed.

PR #965, opened 2018-02-27
#960 - IIIF manifest bug

tbl73:


saracarl:
failed travis CI issue is just the pgp download from MIT failing again.  Not worried about that.

PR #969, opened 2018-02-27
874 contentdm import

saracarl:


saracarl:
https://github.com/benwbrum/fromthepage/issues/978 needs to be fixed.

Issue #983, opened 2018-03-06
Wrong `@collection` for LTR/RTL and breadcrumbs


benwbrum:
Occasionally--usually when clicking on a page on an activity stream--we load the correct page and work, but see the breadcrumb title and RTL setting for a different collection.

Two leads on this:

- The activity stream  links don't use friendly URLs, so they are passing `page_id` without any collection or work slugs.
- The code for loading the `@collection` for the `page_id` is loading the collection based on a session variable `col_id`, then blanking it after it was loaded.  If the `col_id` variable is set elsewhere, then we try to load a `page_id`, we will load the wrong collection.

Relevant code from `application_controller.rb`:
```
    if params[:page_id]
      @page = Page.find(params[:page_id])
      @work = @page.work
      if session[:col_id] != nil
        @collection = set_friendly_collection(session[:col_id])
        session[:col_id] = nil
      else
        @collection = @page.collection
      end
    end
```
I still don't know how to reproduce this.

See also #879

tbl73:
You can reproduced by clicking on a work in a collection, then going to the Collections page and clicking on a page in the activity stream.  Possibly the col_id isn't being correctly blanked out at this point?
The activity stream links are in the deed/deed partial and should probably be updated to use the Rails url helpers.

PR #989, opened 2018-03-09
#983 breadcrumb issues

tbl73:


app/controllers/dashboard_controller.rb
```diff
@@ -35,6 +36,13 @@ def get_data
     logger.debug("DEBUG: #{current_user.inspect}")
   end
 
+  def remove_col_id
```
benwbrum: Can you move the definition `remove_col_id` to `ApplicationController` and remove some of the repeated code?

app/controllers/user_controller.rb
```diff
@@ -8,6 +8,13 @@ def demo
     redirect_to dashboard_path
   end
 
+  def remove_col_id
```
benwbrum: Move to `ApplicationController`

app/controllers/deed_controller.rb
```diff
@@ -3,6 +3,10 @@ class DeedController < ApplicationController
   PAGES_PER_SCREEN = 50
 
   def list
+    #get rid of col_id if no breadcrumbs
```
benwbrum: Call `remove_col_id` from `ApplicationController`

tbl73:
I just pushed up the refactored code.

Issue #996, opened 2018-03-11
"pages needing transcription" and "pages needing review" for a collection


saracarl:
We need 2 new views for a collection -- "pages needing transcription" and "pages needing review".  If there are pages in a work that need either, include that work in the view, with a list of the page names under the work name.  

we'll use it in the the monthly owner emails, the 80% complete email, and we should include buttons on the collection page, like we have for the work page, that link directly to these views.

saracarl:
Combine with #949

Issue #994, opened 2018-03-11
"Transcribe Now" button on collection homepage


saracarl:
"Transcribe Now" button on a collection homepage.  Takes you straight to the transcribe tab of the first untranscribed page of a work in the collection that has not been edited for 12 hours.  If no works fit that criteria, don't display the button.

tbl73:
This seems to relate to #949.

saracarl:
I think we have this done already.  closing.

Issue #949, opened 2018-02-11
Start Transcribing/Start Reviewing Widgets


benwbrum:
We need a reusable widget that drops users straight into transcribing a page in the collection.

### Reusability:
The interface should be a friendly URL of the form `project_name/start-transcribing`.  Hitting this URL should take the user to the first qualifying page (see below), or to the collection homepage with a flash message apologizing if there are no qualifying pages.

This URL should be displayed on the collection settings page, so that project owners can cut-and-paste it into their newsletter and/ or email messages.  _(What about crawling?  We should try not to spam ourselves here.)_  Possibly, we should display an HTML snippet owners can cut-and-paste if they can't do their own links

### Qualifying pages
For the collection, the first qualifying page is:
* A page that is not transcribed that
* Is in a work which has not been edited in the last 48 hours

tbl73:
In terms of the "first" page - do you want it to actually be the first page that has those qualifications, or do you want it to be a random page that has those qualifications?

PR #1014, opened 2018-03-16
#996, #949, #994 - Pages needs transcription/review and Start transcribing views

tbl73:
Note: per #949, I added the URL to the collection settings page, but I don't know whether crawling/self-spamming will be an issue - I didn't do anything special there.
Also, the period of time for the "start transcribing" button/link is 48 hours, per Ben.


tbl73:
Closing until more work is done.

Issue #1016, opened 2018-03-18
rewrite owner page


saracarl:
See meeting Tuesday morning with accompanying Google doc

Issue #957, opened 2018-02-20
Transcriber Landing Page UI


tbl73:
Create a new landing page for transcribers.
Include a carousel -
Highlight collections that meet the following criteria:
Has untranscribed/uncorrected material
Has a collection image
Has collection text
Isn’t private
Work linked to isn’t restricted
Display image, snippet of description, link to an unrestricted work in the collection, owner.
Use one collection per owner (if they have appropriate collections)

Project List
Organize the project list by owner, highlighting the owner name/description/logo/image.
List qualifiying projects for the owner, based on the collection criteria used for the carousel

Only list a handful of collections per owner -- 3?   Should be based on incompleteness (stuff that is available to transcribe or view), first;  then should be either random, rotate per day, or based on owner preferences.  (let’s not do owner preferences at this point in time.)

If we have filtered out collections for brevity (or completion), a “See More” link (better “See 4 more projects”?)  which takes them to the owner homepage.

Include a search bar of collections and owners
Page might be called Find a Project

Have a list to the full collection list at the bottom of the transcriber landing page?

saracarl:
this is done, should have automatically closed and didn't.

Issue #948, opened 2018-02-09
Swap "Collections" and "Dashboard" links on navigation


benwbrum:
We want users who are not logged in to see the list of collections (or whatever our landing page for transcribers might be).  They are most likely to click on the first link on the navbar, which is currently Dashboard.  Swapping this with Collections will probably help

PR #1022, opened 2018-03-25
#957, #948, #1016 - Landing page UI/Owner homepage

tbl73:
Note that this includes the "transcribe a page" link in the landing page carousel - if it's causing too much of a slowdown, we may need to move the logic into a controller action (and have the link whether or not there are pages available to transcribe).


tbl73:
In app/views/layouts/application.html.slim, line 48 (-if !(Collection.all.count > 100)) and in app/controllers/dashboard_controller.rb, line 41 (unless (Collection.all.count > 100)), I'm changing 100 to 1000 as a temporary hack until the code is actually rolled out.

PR #1025, opened 2018-03-26
removed deed filter on collection list for performance

saracarl:


saracarl:
The only failed travis CI test was the one where we couldn't download mysql 57.

tbl73:
You can also remove the call to include works, because we don't display any information about works anymore.  (It was needed for the deeds, but not if you aren't using the deed information.)
So:  collections = Collection.includes(:owner).distinct, and the same for document sets.

saracarl:
closing in favor of new collections page

Issue #958, opened 2018-02-21
User profile copy -- needs updating


saracarl:
"Your name is the only required field here, but we would appreciate if you tell the FromThePage community a little bit more about yourself. "

There's no field other than Name.

saracarl:
http://localhost:3000/freetrial#update-profile

tbl73:
FWIW, all the information is available if you're an owner but not if you're a regular user.

saracarl:
Verify that this does not show up on the account creation screen as well.  (Probably doesn't, but please check.)

jackweinbender:
There also seems to be a "Location" box (perhaps this was added since this issue was opened). So, "A little bit more about yourself" still makes sense.

benwbrum:
Let's get rid of Location, then.  We're probably going to need to do a GDPR audit in the future, so minimizing the information we ask of users is wise.

PR #1051, opened 2018-05-03
#958 Removed location field from user profile

jackweinbender:
For Issue #958
* Removed `Location` from user profile view and update form
* Removed required fields note from user profile update (there's only one field)
* Retained `Location` for owner profile
* Retained required fields note for owners


saracarl:
Reviewed; looks good.  Merging.

Issue #1053, opened 2018-05-03
Collaborators tab CSV download


benwbrum:
Currently we mine the Ahoy logs to determine the total time users have worked on a collection, and display that time next to the list of active collaborators.

Those of our users who actually need this data end up cutting and pasting it into other documents, so the format isn't very useful.

We should tighten up the list of contributors, and add a new CSV download option that produces a row per (active) contributor, with columns for the contributor display name, email address, minutes worked (raw), and proportional minutes worked (see #937).  The date range of the report and collection title/slug make up part of the filename.

Quoth one project owner: _What I need most from that page is the total time recorded at the end.  I have all INSTITUTION staff and students using the same log-in, so my plan for volunteer stats was to subtract the times for those two users (plus the main INSTITUTION log-in, and possibly ARCHIVIST) from the total amount at the bottom._

We can point to a lot of examples of CSV download creation within FromThePage and on other projects we've done.

benwbrum:
A [decent example](https://github.com/VictorianResearchWeb/TheCurranIndex/blob/master/app/controllers/articles_controller.rb#L56) of CSV download of ActiveRecord model contents retrieved from a query.

Issue #937, opened 2018-01-31
Minutes worked report should be proportional


benwbrum:
If a user works on collection A and collection B during the same session (logging deeds or ahoy activity for both), their total time will appear under each collection's activity report.  Worse, their activity will get multiplied if the user is working on the same account.

This is causing problems for Stanford, as one user (who worked on four collections) is appearing as having worked for 222 minutes on each collection. 

Suggested resolution: when displaying total time, adjust it proportionally for the collection by 
`figure_to_show = total_minutes * (deeds_for_this_collection_during_this_session/all_deeds_for_this_session)`

jackweinbender:
Assuming people aren't flipping back and forth between tabs every other deed, we could probably group events (sorted chronologically) by collection by taking the time difference between deeds of differing `collection_id`.

Depending on how granular/consistent the deeds are, however, the benefit may not outweigh the computation cost.

jackweinbender:
I ended up writing it as you suggested, FYI.

PR #1060, opened 2018-05-08
Collaborators CSV export & Proportionate user-time

jackweinbender:
#1053 
  * Added "Download CSV" to `collaborators_body` partial. links to:
  * `collaborator_download` action which returns date-scoped user-time, including proportional time per collection

#937 
  * Added `@user_time_proportional` to the `collaborators_helper`
  * Added proportional time to collaborator display as: `USER: (Collection: 5 minutes | Total: 20 minutes)`


saracarl:
So the CSV download causes the UI to "spin" (our loading icon) and the spinner never stops, even after the file is downloaded.  I'd suggest looking at how we launch other CSV downloads from the export screen.

Issue #1061, opened 2018-05-08
Add Lat/Lon and URI fields to Subjects (documentation needed)


benwbrum:
Subjects are the core of indexing and semantic mark-up in FromThePage, representing indexable people, places, themes, events, species, animals, and other project-defined entities within FromThePage.  (For historical reasons, they are stored in the `Article` model.)

Adding latitude/longitude attributes for GIS support has been funded through the Pelagios RDG:

> *FTP2 Enhancing subject data in FromThePage:* ​Currently subjects within
> FromThePage are represented with title and article text attributes as well as index entries
> into the places they are mentioned within the text. We will add new fields for latitude and
> longitude for any subjects that are places. In addition, we will add URIs attributes for all
> subjects regardless of category, allowing them to be cross referenced to authorities like
> Pleiades or VIAF. All exports will include these new fields.
> 

To implement this, we should allow subjects in the `Places` category (or child categories of `Places`) to accept an additional two fields from editors to represent latitude and longitude.  In addition, all subjects should accept a new URI field for pointers to external authority records like Pleiades.

Open questions:
* What datatype is standard for latitude/longitude in MySQL?  What's compatible with Postgres, in the event that we upgrade databases in the future to take better adantage of plugins like PostGIS?
* Should we accept coodinates in decimal only, or in degrees, minutes, and seconds?
* How do we add the lat/lon fields to Place subjects but not to other subjects?  Is this a case for Single Table Inheritance?  Should a poor-man's version be implemented by adding the fields and just conditionally hiding them?  One can imagine a future enhancement that adds something like _alphebetized name_ to Person subjects.  (Note also the many-to-many relationship between subjects and categories.)

jackweinbender:
### On Storing Lat/Lon

There's no question that Postgres/postgis is the preferred DB for GIS applications; the consensus seems to be that the MySQL Spacial extensions (MySQL's attempt, introduced in MySQL 5.7) is well-behind Postgres/postgis in just about every way.

That said, if we're just storing lat/lon pairs and not doing spacial analysis on sets of coordinates, it seems like MySQL would be just fine---even without the spacial extensions---for storing decimal numbers. The major benefits of the geospacial data as its own datatype is to offload the computation  to the database for finding "nearby" things or "things within X miles of", etc. Even there, assuming we aren't dealing with millions of records, surely this could be done effectively with traditional DB queries.

[...]

PR #1072, opened 2018-05-17
Adds Lat/Lon option to categories

jackweinbender:
Issue #1061
  * Adds Lat/Lon to subject/article table
    * `DECIMAL` type to allow for specific decimal precision
    * Lat: validates between -90/90; Lon validates to -180/180; null allowed
    * Includes `gis_enabled?` method to determine if any related categories are GIS enabled
    * Conditionally displays Lat/Lon fields on subject/edit view based on `gis_enabled?`
    * Fixed form validation error on subject/edit form
  * Adds `gis_enabled` to category
    * Defaults to `false`
    * Enabling/Disabling GIS through "actions" menu affects all descendant nodes in tree
    * Changed "rename category" to "edit category" and added "Enable GIS" setting
    * Added "Enable GIS" to create category form
  * Adds URI field to all subjects

### DIFF
--- a/app/controllers/category_controller.rb
+++ b/app/controllers/category_controller.rb
@@ -47,14 +47,24 @@ class CategoryController < ApplicationController
     @category.update_attribute(:gis_enabled, true)
     @category.descendants.each {|d| d.update_attribute(:gis_enabled, true)}
     
-    flash[:notice] = "GIS Enabled for #{@category.title}"
+    notice = "GIS Enabled for #{@category.title}"
+    if @category.descendants.count > 0
+      notice << " and #{@category.descendants.count} child categories"
+    end
+    
+    flash[:notice] = notice
     ajax_redirect_to collection_subjects_path(@collection.owner, @collection, {:anchor => "category-#{@category.id }"})
   end
   def disable_gis
     @category.update_attribute(:gis_enabled, false)
     @category.descendants.each {|d| d.update_attribute(:gis_enabled, false)}
     
-    flash[:notice] = "GIS Disabled for #{@category.title}"
+    notice = "GIS Disabled for #{@category.title}"
+    if @category.descendants.count > 0
+      notice << " and #{@category.descendants.count} child categories"
+    end
+    
+    flash[:notice] = notice
     ajax_redirect_to collection_subjects_path(@collection.owner, @collection, {:anchor => "category-#{@category.id }"})
   end
 end
--- a/app/controllers/category_controller.rb
+++ b/app/controllers/category_controller.rb
@@ -39,5 +39,14 @@ class CategoryController < ApplicationController
     flash[:notice] = "Category has been deleted"
     redirect_to "#{request.env['HTTP_REFERER']}#{anchor}"
   end
-
+  def enable_gis
+    @category.update_attribute(:gis_enabled, true)
+    flash[:notice] = "GIS Enabled for #{@category.title}"
+    ajax_redirect_to "#{request.env['HTTP_REFERER']}#category-#{@category.id}"
+  end
+  def disable_gis
+    @category.update_attribute(:gis_enabled, false)
+    flash[:notice] = "GIS Disabled for #{@category.title}"
+    ajax_redirect_to "#{request.env['HTTP_REFERER']}#category-#{@category.id}"
+  end
 end
--- a/app/controllers/category_controller.rb
+++ b/app/controllers/category_controller.rb
@@ -19,7 +19,10 @@ class CategoryController < ApplicationController
 
   def add_new
     @new_category = Category.new({ :collection_id => @collection.id })
-    @new_category.parent = @category if @category.present?
+    if @category.present?
+      @new_category.parent = @category 
+      @new_category.gis_enabled = @category.gis_enabled
+    end
   end
 
   def create
--- a/app/controllers/category_controller.rb
+++ b/app/controllers/category_controller.rb
@@ -11,7 +11,7 @@ class CategoryController < ApplicationController
   def update
     if @category.update_attributes(params[:category])
       flash[:notice] = "Category has been updated"
-      ajax_redirect_to "#{request.env['HTTP_REFERER']}#category-#{@category.id}"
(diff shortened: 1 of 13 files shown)


app/controllers/category_controller.rb
```diff
@@ -36,8 +39,22 @@ def create
...
     flash[:notice] = "Category has been deleted"
     redirect_to "#{request.env['HTTP_REFERER']}#{anchor}"
   end
-
+  def enable_gis
+    @category.update_attribute(:gis_enabled, true)
+    @category.descendants.each {|d| d.update_attribute(:gis_enabled, true)}
+    
+    flash[:notice] = "GIS Enabled for #{@category.title}"
+    ajax_redirect_to "#{request.env['HTTP_REFERER']}#category-#{@category.id}"
```
benwbrum: Path generation by mucking about with `HTTP_REFERER` is pretty ugly.  Do we already have a category_path URL helper we can use?  If not, it would be worth making one.  (We might need to pass in the controller as well.)

app/controllers/category_controller.rb
```diff
@@ -36,8 +39,22 @@ def create
...
+  end
+  def disable_gis
+    @category.update_attribute(:gis_enabled, false)
+    @category.descendants.each {|d| d.update_attribute(:gis_enabled, false)}
+    
+    flash[:notice] = "GIS Disabled for #{@category.title}"
+    ajax_redirect_to "#{request.env['HTTP_REFERER']}#category-#{@category.id}"
```
benwbrum: See above comment on path helpers.

app/models/article.rb
```diff
@@ -14,6 +14,10 @@ class Article < ActiveRecord::Base
 
   validates_presence_of :title
 
+  validates :latitude, allow_blank: true, numericality: { less_than_or_equal_to: 90, greater_than_or_equal_to: -90}
+  validates :longitude, allow_blank: true, numericality: { less_than_or_equal_to: 180, greater_than_or_equal_to: -180}
```
benwbrum: Let's add the following comment to make these fields explicit:
`# latitude and longitude contain modern coordinates for places and are normalized as such.  They do not represent coordinates found within transcribed texts.`

app/controllers/category_controller.rb
```diff
@@ -36,8 +39,22 @@ def create
...
+    
     flash[:notice] = "Category has been deleted"
     redirect_to "#{request.env['HTTP_REFERER']}#{anchor}"
   end
-
+  def enable_gis
+    @category.update_attribute(:gis_enabled, true)
+    @category.descendants.each {|d| d.update_attribute(:gis_enabled, true)}
+    
+    flash[:notice] = "GIS Enabled for #{@category.title}"
```
benwbrum: Can we have the success notice mention the children, to make this explicit?  Something like 
`GIS Enabled for #{@category.title} and #{@category.descendants.count} child categories` or `GIS Enabled for #{@category.title} and all children` would make the cascading behavior clearer to users.

app/models/article.rb
```diff
@@ -65,6 +69,10 @@ def self.delete_orphan_articles
   def related_article_ranks
 
   end
+  
+  def gis_enabled?
+    self.categories.any? { |c| c.gis_enabled }
```
benwbrum: Can we rewrite this to push the logic into the database query, rather than needing to loop through results?  `self.categories.where(:gis_enabled => true).present?` might work.

app/models/article.rb
benwbrum: I've just tried this out in the UI, and wonder about the proper behavior when people enter digits longer than the precision allows.

On one hand, we truncate the values silently, which is usually the wrong thing to do.  On the other hand, truncating lat/lon can be absolutely correct for GIS data, in which transformations (division by something close to one, like `3456/3457` may be applied to adjust data from one coordinate system to another) may extend digits of data far beyond what is sane or sensible: We have seen city boundary points expressed with enough precision to identify positions at the sub-atomic level due to these transformations.

I think that if we truncated values to a number of significant digits representing 20m precision or less, and flash that as part of the success message when a user saves, that will be the best compromise.

benwbrum:
We also need a set of spec tests before we can merge this.  You may be able to find an existing test suite that modifies subjects/articles and categories and add tests for the permutations of gis attribute cascading and lat/lon field display there.

app/controllers/category_controller.rb
jackweinbender: This whole controller seemed to use that method, so I went with it, assuming that there was something about it that I didn't know. My guess was that it had something to do with redirecting to specific anchors, though, that is perfectly doable with path helpers. I'll fix them all.

app/models/article.rb
jackweinbender: 1. Is the lat/long comment meant to be for the user (and thus, in the view), or just noted in the model?

2. I completely agree re: truncation. Should we change the precision of the DB table, then, to align with this ~20m precision? I found a handy SO answer which indicates that would mean truncating to 4 decimal places (that's 11m precision), which is about as good as standard consumer GPS does. Or maybe we should make it 5 decimal places just to be a bit conservative? It's currently at 8, which is obviously too precise; I just picked it because Google seemed to use that for Maps (though, I'm counting 7, now).

jackweinbender:
- [x] Use path helpers
- [ ] Update Lat/Long Fields
- [ ] Add Lat/Long truncation flash message
- [x] Added affected children count to En/Disable GIS action
- [x] Moved `.any?` logic into Db query
- [x] Added tests for GIS enable/disable and conditional dispaly of Lat/Lon fields

app/models/article.rb
jackweinbender: I went with 5 digits of precision for the DB table, and added a `GIS_DECIMAL PRECISION` constant to the article controller which governs how it determines if the DB truncated anything as well as how many zeros to pad with. I thought the padding was a nice way to indicate to users what the precision should be (?).

saracarl:
Is this ready?

jackweinbender:
Yes, this is ready. 
Sorry, I wasn't sure how to mark the code review part as "complete."

Issue #1070, opened 2018-05-15
Start and end dates for `iiif/contributions` get route not found


benwbrum:
The contributions API for FromThePage should accept _originating domain_, _terminus a quo_, and _terminus ad quem_ as parameters, but currently only the first works, and passing other parameters gets a route not found:

* [working](https://fromthepage.com/iiif/contributions/ucd.ie)
* [should work but doesn't](https://fromthepage.com/iiif/contributions/ucd.ie/2018-01-01)

No route matches [GET] "/iiif/contributions/ucd.ie/2017-01-01"

For more information on what's supposed to happen, see the [API docs](https://github.com/benwbrum/fromthepage/wiki/FromThePage-Support-for-the-IIIF-Presentation-API-and-Web-Annotations#iiifcontributions-user-contribution-api)

PR #1075, opened 2018-05-23
iiif contribution route failure

jackweinbender:
#1070 
The "." in the `:domain` parameter was causing the other matches to fail. Adding a `.*` constraint to the `:domain` parameter for the date-specific routes solved it.


saracarl:
@jackweinbender -- can you add a spec test for this to your to-do list?  Thanks!

jackweinbender:
Done.  Travis is giving me that GPG error again on one of the builds (that's why it's failing), so I'll try it again in a couple hours.

Issue #1073, opened 2018-05-18
assign_categories view needs more context


benwbrum:
Whenever a user marks up a new subject, we present them with a screen to assign categories to the subjects on the next page.  For subject-heavy texts like Alcedo, it can be bewildering to see the subjects outside of their context within the transcripts.  Presenting the following words for their occurrence in the transcribed page would help immensely, and possibly the preceding words as well.

Example:
![uncategorized](https://user-images.githubusercontent.com/199961/40213253-20808dc8-5a1a-11e8-9288-4e418ed0dd86.png)

Compare with (hypothetical):

*Abary* ABARY, a small river of Guayana, between

*Abbeville* ABBEVILLE County, in Ninety-six district, S. Carolina

*South Carolina* S. Carolina, bounded on the n.e. by the Saluda

*Abbots* ABBOTS, a small river of N. Carolina

jackweinbender:
#1076

benwbrum:
We've had a request that the number of characters providing the context be at least doubled.

PR #1076, opened 2018-05-23
Add context to `assign_categories` view

jackweinbender:
#1073 
I wasn't expecting this one to be a one-liner! Here's what it looks like:

image

A couple things:

- I feel like using the `page.source_text` content makes for a rather busy display (showing other sets of brackets, etc.), but because we are matching on the `article.title` we can't use `page.search_text`, since the titles don't always match the display text. Also, I suppose it actually is useful to include both the canonical subject title as well as the displayed text so the user can see what s/he is linking.
- Since the context goes forward and backward, I went ahead and put it on its own line. I tried to strike a balance between being able to quickly identify the word in the context (by highlighting it), and trying to keep it from dominating the page (mono-font, not too big, etc.)

If we aren't happy with this, an alternative to giving the context for each subject might be to show the whole page (like the "preview") off to the right, then on the front-end doing a hover/highlight to show the context. So, for example, show the whole page on the right (like in the preview) but, when you hover over each of the unique terms that we are categorizing on the left, the terms as they appear in the page would highlight. Does that make sense?

### DIFF
--- a/app/views/transcribe/assign_categories.html.slim
+++ b/app/views/transcribe/assign_categories.html.slim
@@ -13,6 +13,7 @@ p.big You have uncategorized subjects mentioned in the page transcription. Pleas
     -update_article_url = url_for({ :controller => 'article', :action => 'article_category', :article_id => article.id })
     .page-uncategorized
       h4.page-uncategorized_subject(data-prefix="Subject: " *language_attrs(@collection)) =="&ldquo;#{article.title}&rdquo;"
+      code= highlight(excerpt(@page.source_text, article.title, separator: ' ', radius: 6), article.title)
       select(data-assign-categories="#{update_article_url}" size="1" aria-label="Assign categories" multiple)
         -@collection.categories.walk_tree do |c, level|
           -selected = true if article.categories.include?(c)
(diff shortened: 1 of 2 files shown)


saracarl:
I like this solution a lot.  source_text is definitely better to use than search_text (we *produce* search text, we don't use it, for the most part).

app/views/transcribe/assign_categories.html.slim
```diff
@@ -13,6 +13,7 @@ p.big You have uncategorized subjects mentioned in the page transcription. Pleas
     -update_article_url = url_for({ :controller => 'article', :action => 'article_category', :article_id => article.id })
     .page-uncategorized
       h4.page-uncategorized_subject(data-prefix="Subject: " *language_attrs(@collection)) =="&ldquo;#{article.title}&rdquo;"
+      code= highlight(excerpt(@page.source_text, article.title, separator: ' ', radius: 6), article.title)
```
benwbrum: I've never seen either `highlight` or `excerpt` before -- very cool!

jackweinbender:
Since this almost demands that the line will need to wrap, I thought it would be more useful to preserve the newlines in the document that is being transcribed. It, likely, will also be helpful for the transcriber to keep the spacial relationship between the tern in the transcription and in the excerpt. I thought 3 lines of context on either side made sense.

image

Issue #1078, opened 2018-06-02
IIIF API does not expose works imported from Internet Archive


benwbrum:
The IIIF controller explicitly checks to see whether a work has been imported from the Internet Archive, then does not display it, because we did not understand pass-through annotations/derivative manifests when this code was written.  This prevents NYBG from using the API.

PR #1079, opened 2018-06-02
Internet Archive imported works now support IIIF API

benwbrum:
Since the IIIF server API was built before the Internet Archive supported IIIF
it did not expose any Internet Archive content.  Now it re-presents IIIF canvases
and works, so an IA work can be imported into FromThePage, transcribed, and
then viewed in IIIF clients like Mirador.


benwbrum:
This pull request is for issue #1078

PR #1106, opened 2018-06-28
Adds Page Number column for Transcription Fields

AustinCarr:
Adds page_number column to transcription_fields table.

Filters transcription fields based on associated work's page number.

### DIFF
--- a/app/models/transcription_field.rb
+++ b/app/models/transcription_field.rb
@@ -5,9 +5,10 @@ acts_as_list :scope => :collection
 has_many :table_cells
 validates :options, presence: true, if: Proc.new {|field| field.input_type == 'select'}, on: [:create, :update]
 
-attr_accessible :label, :collection_id, :input_type, :options, :line_number, :position, :percentage
+attr_accessible :label, :collection_id, :input_type, :options, :line_number, :position, :percentage, :page_number
 
 validates :percentage, numericality: { allow_nil: true, greater_than: 0, less_than_or_equal_to: 100 }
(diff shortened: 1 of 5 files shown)


saracarl:
This change is failing 1 test:

 1) collection settings js tasks uses page arrows with unsaved transcription
     Failure/Error: page.fill_in('fields_1_First_field', with: "Field one")
     
     Capybara::ElementNotFound:
       Unable to find visible field "fields_1_First_field" that is not disabled
     # ./spec/features/field_based_spec.rb:118:in `block (2 levels) in <top (required)>'

We'll need to investigate why and fix it before we merge it in.

AustinCarr:
There isn't a transcription field on the second page because I changed the "_edits fields for transcription_" test to fill in the newly created Page Number field with the value of "1".

Because all three of the test fields then have a Page Number value of 1, they're only displayed on the first page of each Work. When the "_uses page arrows with unsaved transcription_" test visits the page 2, it couldn't find the 'fields_1_First_field' input box, because it was set to only display on page 1.

I'm going to switch the "_edits fields for transcription_" test back so the newly-created Page Number field is no longer assigned, which will display the transcription field on all pages. I'm also going to change the "_uses page arrows with unsaved transcription_" test back to using the second page.

AustinCarr:
This feature is for multi-page works that have different fields on each page. When a Page Number value is added for a Transcription Field, the user will only see fields associated with the page instead of every field in the work on every page.

For example if there are two pages to a work, the fields associated with page one will show with the first image of the work; only the fields associated with page two will show on page two of the work. And it's set up so that if no Page Number value is entered, then that field will be displayed on all pages of the work.

PR #1117, opened 2018-07-17
Expanded subject categorization context to three lines

jackweinbender:
Makes context three lines. See the discussion on #1073.

image

I had to replace `\n` with `<br />` because `.excerpt()` strips newlines.


jackweinbender:
This will needs #1119 before it will pass tests.

PR #1118, opened 2018-07-17
Update to Ruby 2.3.7

jackweinbender:
Support for the Ruby 2.2 series has ended.


jackweinbender:
This needs #1119 before it will pass tests.

PR #1121, opened 2018-07-20
Implement redesign

techpeace:
The following changes are implemented:

* Redesign the homepage
* Redesign sign in page
* Add a new "Software" page
* Add a new "Contact Us" page and mailer

I added the [`dotenv-rails` gem](https://github.com/bkeepers/dotenv) to enable setting the `TO` address on emails via environment variable. If you don't mind setting this in public code, just let me know which address to use and I'll drop it.

I also added the [`quiet-assets` gem](https://github.com/evrone/quiet_assets) to the development environment, which is a handy addition to Rails 4 projects. It just silences all of the asset retrieval lines in your logs, which makes it easier to follow what's happening in the app as you tail the logs.


config/routes.rb
```diff
@@ -64,6 +64,15 @@
   get '/rails/mailers' => "rails/mailers#index"
   get '/rails/mailers/*path' => "rails/mailers#preview"
 
+  get '/software', to: 'static#software', as: :about
+  get '/faq', to: 'static#faq', as: :faq
+  get '/static/faq', to: redirect('/faq', status: 301)
```
techpeace: I redirected the former `/static/foo` URLs to `/foo`, as shorter, more readable URLs can help SEO. If there was another reason you were doing this I don't know about, just let me know and I'll revert. :smile:

techpeace:
I'll run down those test failures.

saracarl:
Looks good.

I was able to fix the test problem by changing the mysql57 specific gemfile travisCI uses (in gemfiles) for those legacy environments.  

also added "live" page transcribed counts and a real email address.  

I'm going to close this PR since I couldn't add to your fork; you can see my PR with both sets of changes here if you're interested:
https://github.com/benwbrum/fromthepage/pull/1127

techpeace:
Oh my goodness, I didn't see those additional Gemfiles. That explains it, haha. Thanks, Sara!

PR #1127, opened 2018-07-25
Landing redesign

saracarl:


techpeace:
If you ever have any secrets that you'd like to keep out of the public repo, you can `.gitignore` the `.env` file and simply require that it be present when setting up the local dev environment. You can instruct folks to use `.env.example` as a template.

saracarl:
I need to figure out how to keep capistrano from wiping it out in production every time we deploy if we use that path.  (which I agree, would be better.)

techpeace:
Ah, good call - looks like this might do the trick: https://github.com/rjocoleman/capistrano-env-config

techpeace:
Oh nm, that might just be for using `.env` in Capfiles themselves. The other technique I've seen is to hold the production `.env` somewhere outside the deploy directory and copy it in afterward.

PR #1133, opened 2018-07-26
Use values from 01fromthepage initializer for Devise

### LINKED ISSUE #1132: Devise initializer ignores Google/SAML constants set in 01fromthepage initializer
To enable Google OAuth and/or SAML login, you need to set some constants in `01fromthepage.rb`. However, not all of these constants are actually used in `devise.rb`:

https://github.com/benwbrum/fromthepage/blob/8f41c24e45c8b9c1c058af147dd62b29d816d359/config/initializers/devise.rb#L234-L245

I will try to fix this.

bencomp:
This fixes #1132.

(I cherry-picked the commit that fixes the Travis CI build)

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -232,7 +232,7 @@ Devise.setup do |config|
   # up on your models and hooks.
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
   if ENABLE_GOOGLEOAUTH then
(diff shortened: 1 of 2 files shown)


bencomp:
I'm not sure why the build errored; the log doesn't say anything after "network is available"

saracarl:
@bencomp -- did you test this in an actual SAML environment?  travisCI doesn't; so either I need to (I don't have a saml environment, so all I can do is the samling one) or need to know if you did.  Thanks!

PR #1137, opened 2018-07-26
Adds option to preserve filenames as page titles when uploading zip files

### LINKED ISSUE #1083: Uploaded files (from zip) should maintain filenames as titles
Each image filename should become the title for the page it represents.

jackweinbender:
Closes #1083 

Adds the option to use the names of the files contained in an uploaded zip file as page titles when imported as a work.


benwbrum:
One of the IIIF manifests we use to test with has gone walkabout, so it’s probably not you.

benwbrum:
I'm merging `development` into this feature branch to see if the fixes there get the build passing.

PR #1140, opened 2018-07-27
User-defined dictation preference

jackweinbender:
Users can now specify their default dictation language in their user profiles. If unspecified, the dictation language should default to the collection-specified language (which defaults to `en-US`).


jackweinbender:
I pulled out a couple assertions that no longer made sense. The two remaining test failures seem to be this same IIIF manifest issue.

Issue #1156, opened 2018-08-02
Assign categories gives 500 error on Alcedo


benwbrum:
Saving any of the first four pages in Alcedo results in a 500 error.  The logfiles indicate that we have an unexpected null value somewhere:

```

I, [2018-08-02T05:09:00.533119 #8279]  INFO -- : TRANSCRIPTION	2018-08-02 05:09:00 -0500
TRANSCRIPTION	User	ID: 3	Email: benwbrum@gmail.com	Display Name: benwbrum
TRANSCRIPTION	Collection	ID: 23	Title:LatAm Digital Edition and Gazetteer	Owner Email: benwbrum@gmail.com
TRANSCRIPTION	Work	ID: 627	Title: The geographical and historical dictionary of America and the West Indies [volume 1]
TRANSCRIPTION	Page	ID: 8764	Position: 1	Title:Untitled Page 1
TRANSCRIPTION	Source Text:
BEGIN_SOURCE_TEXT
THE

GEOGRAPHICAL AND HISTORICAL

DICTIONARY

OF

AMERICA AND THE WEST INDIES.

ABA

[Aaronsburgh lies at the head of Penn’s
Creek, Northumberland county, Pennsylvania ;
about SO miles w. from Lewisburgh, and 40 w. by
n. from Sunbury. Lat. 40° 52' 30" n. Long. 77°
SI' SO" O).]

==[[ABACACTIS, or Abacaris,]]==, a settlement of
Indians, of this name, in the [[province of the Ama-
zonas]], and in the part or territory possessed by the
Portuguese. It is a reduccion of the religious
order of the Carmelites of this nation, situate on
the shores of a lake of the same name. It lies
between this lake and a river, which is also so
called, and which is a large arm of the Madeira,
which, passing through this territory, afterwards
returns to that from whence it flowed, forming the
[[island of Topinambes]].

[==[[Abaco|ABACO]]==, one of the largest and most northern
of the [[Bahama islands]], situate upon the s. e. end
of the Little Bahama bank. The Hole in the
Rock, or (as it is most commonly called) the
Hole in the Wall, is the most southern point of
the island, and bears about 18 leagues north from
the island of New Providence, about 9 or 10
leagues in a n. w. direction from Egg Island,
and about 10 or 12 in a n. e. direction from
the Berry islands. About 10 leagues to the n. of
the Hole in the Wall, on the e. side of the island,
is Little Harbour, the entrance to which is be-
[...]

PR #1157, opened 2018-08-05
Don't blow up if we can't find the subject title in the text

benwbrum:
Stopgap for UT's installation.


benwbrum:
Do not close the underlying issue if we merge the pull request; this is just a stop-gap fix.

benwbrum:
This addresses issue #1156

Issue #1154, opened 2018-08-01
transcription fields: automatically add unique column names for transcription fields in export


AustinCarr:
Each transcription field already has unique identifiers associated with it in its line_number and position.

Can't those be automatically added onto the column names in export, so there's no collisions?

saracarl:
We believe this is fixed.

PR #1165, opened 2018-08-22
Unique exports

AustinCarr:
Fixes issue #1154 

Skips exporting "instruction" input types.

Creates a new function that creates an array of unique identifiers by using the line number and position of each field.

Adds the unique identifier to the headers.


AustinCarr:
The "#get headings from non-field-based" comment seems like it's incorrect.

What's the difference between a table cell and a transcription field?
Do non-field-based transcriptions have table cells?

benwbrum:
Table cells and transcription fields are the same thing.  We originally created the table export (and the infrastructure underlying field-based transcription) for tables embedded within textual documents, like lists of specimens collected within a naturalists' field notes, or lists of shipments of bacon to a plantation in agricultural journals.

The field-based interface simply creates a table on a page that has a single row, and suppresses the ability to add narative text around it.

app/controllers/export_controller.rb
```diff
@@ -185,27 +185,43 @@ def update_contentdm_credentials
...
+  def get_unique_identifiers(fields)
+    uniqueIdentifiers = Array.new(fields.length)
+
+    for i in 0..(fields.length - 1) do
+      uniqueIdentifiers[i] = "#{fields[i].line_number}.#{fields[i].position}"
+    end
+
+    uniqueIdentifiers
+  end
+
```
benwbrum: No need to change this before we deploy, but a the Ruby `map` function can do some lovely magic here:
```ruby
unique_identifiers = fields.map { |field| "#{field.line_number}.#{field.position}"}
```

benwbrum:
The code review is done, but we need to test this code against the Graves Diaries, Stagville Accounts, and either ADAH or IARA before we deploy to make sure that data prior to the line_number work still exports correctly.

benwbrum:
I've tested this, and it works just fine for text+table exports.  It does change the column headers for field-based collections in ways that might be surprising to people not using the page feature.

What would you say to conditionally adding the suffix in unique_identifiers, so that a field with a unique name did not acquire the suffix, but one that had a duplicate field name would get the duplicate?  (I can implement this, if you like, but want to run the idea past you.)  That would be less likely to surprise other users.

AustinCarr:
I think that's a great idea!

The main goal was just to allow users to name the fields whatever they want and not have to worry about column name collisions at all.

This was the most straight-forward solution I could think of.

benwbrum:
Replaced with #1230

PR #1158, opened 2018-08-06
Better excerpting for subject categorization review

jackweinbender:
Closes #1156

More robust treatment of uncategorized subjects review page.

Todo:
- [x]  Handle excerpting and highlighting for subjects with linebreaks
- [x] Handle the Translation case
- [x] Make sure updating works with links containing newlines
- [x] Figure out how to diambiguate `[[FooBar]]` from `[[Foo Bar]]` when `[[Foo\nBar]]` 
- [x] Test all the things.

PR #1167, opened 2018-08-25
1062 section headers

### LINKED ISSUE #1062: Support subjects within section headers
From the Pelagios RDG narrative:

> FTP1 Enhance FromThePage’s markup for place hierarchies​: Currently
> FromThePage supports markup of subjects which can be categorized as places, as well as
> headers delimiting a section of text. We will extend this capability so that section headers
> may contain subject tags and appear inline as head words of an entry. This will allow
> hierarchal export of the place names and constituent places extracted from the entry.
> Administrative units will be categorized so that administrative hierarchy can be exported.
> The following example is from the English translation of Alcedo11:
> 
> > AUYAMAS, a river of the province and government of Santa Maria in the kingdom of
> > Tierra Firme, it is very abundant, rises in the snowy sierra, traversing in a rapid course the
> > valley of Upar, and after running 72 leagues, it enters the e. side of the river Magdalena.
> > 
> 
> We will mark up each section header (location name) with a section header specifier (===)
> as well as a subject link and then subject link each additional geographical subject within the
> section. (A section runs until the next section header is encountered.) Example:
> 
> > `===[[AUYAMAS]]===`, a `[[river]]` of the province and government of `[[Santa Maria]]` in the
> > kingdom of `[[Tierra Firme]]`, it is very abundant, rises in the snowy sierra, traversing in a
> > rapid course the `[[valley of Upar]]`, and after running 72 leagues, it enters the e. side of the
> > `[[river Magdalena]]`.
> > 
> 

Currently, FromThePage's mark-up supports section headers in an undocumented (and probably buggy) fashion.  Section headers certainly do not support subject mark-up.  We need to allow this kind of subject embedding and come up with some way to flag the fact that this is a lemma.

jackweinbender:
Closes #1062 
Depends on #1158 

I still need to come up with a good testing strategy for the new features, but I did write some tests beforehand to prevent regressions.

Because neither the DB nor the XML contained all of the necessary data for each link (e.g., we don't store categories in the XML and section data must be read from the context of the whole *work* and not just the page), this class generates a Hash from the XML of each page containing the link ID mapped to the current section "path".

Using the old code for generating the CSV, the class performs lookups for each link ID to retrieve the section in which it was found.

At the moment, jumps from heading depths *up* in the hierarchy (h4 followed by h2) works fine, but jumps larger than 1 going deeper into the hierarchy (h2 followed by h4) is still not 100%.


app/models/subject_exporter.rb
```diff
@@ -1,39 +1,74 @@
+require 'nokogiri'
+
```
benwbrum: Would it make sense to move this class to `lib` instead of `app/models`?

jackweinbender: I was unsure about where to put this kind of code. `lib`, now that I see it, seems to make sense.

jackweinbender:
I seem to have only applied the section stuff to the `xml_text` field. As usual, `translation` is neglected.

Todo:
- [ ] Refactor factories using multiline strings
- [ ] Apply export functions to translations, too
- [ ] Move export code to `lib` directory

benwbrum:
We've had further requests on this for:
* A column in the export for each subject, showing the URI for the subject itself

jackweinbender:
Just to be clear, do we mean the URI of the page for a given subject on FTP, or the data stored in the URI field on the `subject` model?

jackweinbender:
The current state of this PR uses the URI field from the Article/Subject  model. If that's what we want, this is ready to go.

benwbrum:
It turns out that we need both in the export; the URI field value as well as the link to the FtP subject page.

jackweinbender:
So, as I was playing with some of the UT data today, I noticed a couple pages weren't outputting any `section` data. I couldn't account for it. But, after clicking `Save` for both the transcription and translations, everything worked perfectly. I'm thinking. maybe the XML formatting, perhaps, has changed since those pages were transcribed?

jackweinbender:
That'd be a pretty straight-forward way to take care of it, if that's what's wrong. 
I'll want to look a bit more closely to make sure that's what's actually wrong, though.

PR #1168, opened 2018-08-26
1166 edit user

### LINKED ISSUE #1166: can't update email address or description in account
The UI supports updating both of these, but the change isn't kept

jackweinbender:
Closes #1166 

Devise needed to be told explicitly to allow email, login, etc. to be set as assignable for the `user#update` route.

Updating the user info also redirected to the application root, which (I think) is unexpected behavior, so I changed that to stay on the `user#edit` view after the fact.

~~Still needs tests; I've worked on it some, but there seems to be an issue persisting the session after updating the user's info that (again, I think) is related to how Devise manages these updates. Anyhow, that's why there's no tests.~~
There's still some weirdness here, but I've got a passing test now.


jackweinbender:
Turns out I broke this when I added the `dictation_language` stuff.
I added a test for it and refactored a misleading `if-else` structure.

saracarl:
So this is still not working 100%.  If I edit my name, ORCID, URL or location in the user profile it does not save successfully.

benwbrum:
This is actually working and ready to merge.

PR #1186, opened 2018-09-03
Adds contributors and notes columns to table export

### LINKED ISSUE #1164: include contributors and notes in spreadsheet export
table data export should include the contributors per record

jackweinbender:
Closes #1164 

Adds Contributors and Notes to table export.

Contributors are in the format `Name<email@address>` and delimited with pipes.
Notes are in the format `[Name<email@address>]: Note` delimited with pipes. Notes are sanitized by replacing pipes with `//`; White space is stripped and replaced with `.gsub(/\s+/, ' ')`

### DIFF
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -263,7 +263,7 @@ private
         page_notes = page.notes
           .map{ |n| "[#{n.user.display_name}<#{n.user.email}>]: #{n.body}" }.join('|').gsub('|', '//').gsub(/\s+/, ' ')
         page_contributors = all_deeds
-          .select{ |d| d.page_id == page.id && d.deed_type == 'page_edit'}
+          .select{ |d| d.page_id == page.id}
           .map{ |d| "#{d.user.display_name}<#{d.user.email}>".gsub('|', '//') }
           .uniq.join('|')
 
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -261,10 +261,10 @@ private
       unless page.table_cells.empty?
         page_url=url_for({:controller=>'display',:action => 'display_page', :page_id => page.id, :only_path => false})
         page_notes = page.notes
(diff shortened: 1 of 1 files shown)


app/controllers/export_controller.rb
```diff
@@ -221,26 +221,62 @@ def export_tables_as_csv(table_obj)
...
   def generate_csv(work, csv, col_sections)
+    all_deeds = work.deeds
     work.pages.includes(:table_cells).each do |page|
       unless page.table_cells.empty?
         page_url=url_for({:controller=>'display',:action => 'display_page', :page_id => page.id, :only_path => false})
-        page_cells = [work.title, work.identifier, page.title, page.position, page_url]
+        page_notes = page.notes
+          .map{ |n| "[#{n.user.display_name}<#{n.user.email}>]: #{n.body}" }.join('|').gsub(/\s+/, ' ')
+        page_contributors = all_deeds
+          .select{ |d| d.deed_type = 'page_edit' }
```
jackweinbender: Here. I pulled all the deeds for the work to avoid making multiple database calls, then forgot to filter to the page.

jackweinbender: Also, double-equals. Rookie mistake.

saracarl:
Page contributors are now blank!

benwbrum:
Rough reproduction steps:
* Field-based transcription
* Two different users transcribed two different pages
* Export the table CSV for the individual work.

jackweinbender:
So, in an effort to scope the relevant users (see first problem with this PR), I over-zealously scoped the kinds of deeds to only register `page_edit` deeds, which are not what field-based transcriptions utilize (they use `page_trans`) deeds. I got a false positive on this part earlier due to a typo (`=` for `==`; since the assignment worked, `=` returned true! Agh!), which is how I missed it.

Anyhow, I removed the `deed_type` from my `.select` function, since it's not really necessary, and now anyone who committed a deed on a particular page will show up in the list.

PR #1198, opened 2018-09-13
Openseadragon `maxZoomPIxelRatio`

### LINKED ISSUE #1193: Improve maxZoomPixelRatio setting on OpenSeaDragon
If users are on very high-resolution monitors, or are looking at low-resolution scans, they may want to zoom in beyond the one-to-one pixel ratio.  This is an option configurable on OpenSeaDragon.

https://openseadragon.github.io/docs/OpenSeadragon.html#.Options

jackweinbender:
Closes #1193

Set it to `10` for no specific reason. I didn't notice a terribly adverse UX by using such a high number.

I suppose we could make this a setting in the `work` if it becomes distracting for higher-res images, but I doubt that'll be necessary.

### DIFF
--- a/app/views/shared/_iiif_div.html.erb
+++ b/app/views/shared/_iiif_div.html.erb
@@ -57,6 +57,7 @@
 		        preserveViewport:   true,
 		        visibilityRatio:    1,
 		        minZoomLevel:       1,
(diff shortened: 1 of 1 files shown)


saracarl:
This is pretty amazing!

PR #1200, opened 2018-09-17
1190 link to api documentation

### LINKED ISSUE #1190: Link to API Documentation in export tab
Include a link to this page:  https://github.com/benwbrum/fromthepage/wiki/FromThePage-Support-for-the-IIIF-Presentation-API-and-Web-Annotations

Towards the end of the export page with the text "We also have an API for programmatically exporting transcripts for your collections.  Learn more here."  (and link the "here" to the documentation)

lortza:
Resolves #1190 

I've added the link to the API doc to the bottom of the `/export` page for issue #1190. In the process, I removed trailing whitespace from two lines in the `export/index.html.slim`, since my text editor does this automatically. I see that html.slim is whitespace-sensitive. I have not seen an issue with the removal of these particular trailing whitespaces on this page. Do you foresee this as a problem?

screenshot 2018-09-17 15 47 26

### DIFF
--- a/app/views/export/index.html.slim
+++ b/app/views/export/index.html.slim
@@ -87,6 +87,11 @@ table.datagrid
 
 =render(:partial => 'shared/pagination', :locals => { :model => @works, :entries_info => true })
 
+h2 API
+p
+  'We also have an API for programmatically exporting transcripts for your collections. Learn more
+  =link_to('here', 'https://github.com/benwbrum/fromthepage/wiki/FromThePage-Support-for-the-IIIF-Presentation-API-and-Web-Annotations')
+  '.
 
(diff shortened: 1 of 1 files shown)


lortza:
I'm working through some issues with my local environment and have had to comment out the `capybara-webkit` gem for the time being. You won't see this in the PR, but this means I have to run front-end tests via a push to GH and Travis. For this reason, you may see several pushes from me until I have this resolved.

saracarl:
Push as many times as you need -- we may need a way to indicate that's it's ready for review.  I also think tests will run if you just push to your local branch (no PR needed).  

I don't think getting rid of whitespace at the *end* of a slim template line is a problem.  At the beginning, it would be.

PR #1206, opened 2018-09-20
1174 hide pages need indexing

### LINKED ISSUE #1174: hide "pages needing indexing" if indexing disabled.
In the work view (at least) the "Pages That need indexing" shows up even though indexing is not enabled.  We should not show that button if the collection does not have indexing enabled.

lortza:
Resolves #1174 

This PR resolves the problem that in the work view, the "Pages That need indexing" button was showing up even though indexing is not enabled for the work's collection. This button no longer appears. This solution makes the assumptions that in order for a `work` to need indexing:
1. its associated `collection` does support document sets (`collection.supports_document_sets == true`)
2. at least _one_ of its associated `pages` has not been indexed yet. If all of the `pages` have been indexed, the work does not require indexing.

### Before:
With the button still in place
screenshot 2018-09-26 18 23 22

### After:
Without the button
screenshot 2018-09-26 18 23 44


saracarl:
Ok, so we need to clarify part of the product here.  "Subject Indexing" lets you link certain words in the transcribed text to canonical subject entries for those words.  See the red-underlined words on these pages as an example:  https://fromthepage.com/benwbrum/julia-brumfield-diaries/1918/display/378  

Project owners can turn on or off this feature as a setting on a collection.  A collection has an attribute call :subjects_disabled that indicates indexing is disabled for a collection.

Here's a very old presentation on subjects:  https://content.fromthepage.com/category/subject-links/

Here's some documentation on Document sets -- they are more like a database view.  https://content.fromthepage.com/project-owner-documentation/document-sets/

PR #1218, opened 2018-10-17
Add optional ReCAPTCHA enforcement

### LINKED ISSUE #1050: spam transcription
We recently experienced some spam transcriptions.

This one has a user that wasn't deleted:
https://fromthepage.com/digitalindy/may-wright-sewall-papers/katherine-gould-to-may-wright-sewall/versions/172548
(I've since deleted the user, but this is their ID and visit:  user_id=54960&visit_id=469142  )

This one transcribed and then deleted the user:
https://fromthepage.com/page_version/show?page_version_id=363355

The URL in the spam transcription is not a live link (on purpose), so this may discourage spammers.

It's been 10 days and we haven't seen any additional spam.  I think we should wait and see if they come back before we take action.

We think the right first step is to add recaptcha to the user registration -- both guest and signed in users.

jackweinbender:
Closes #1050 

The basic feature of this PR is done, however, it currently relies on a global flag set in the `01fromthepage.rb` file, which (I imagine) will break anyone who doesn't update that file. So, I still need a better way of determining globally whether the ReCAPTCHA stuff is configured. That said the ReCAPTCHA stuff seems to be working well in both the `Sign Up` and `Guest` contexts.

TODO:

- [x] Figure out how to check for config variables elegantly
- [x] Replace all conditionals reliant on global constants with the above


jackweinbender:
I've added some logic to the `recaptcha.rb` initializer to help keep the app from blowing up if a user doesn't have the ReCAPTCHA stuff configured.

For new installs, I've added two new constants to the `01fromthepage.rb` file that draw from environment variables. These are set to `nil` if no environment variable is set.

For current installs (who may choose not to update `01fromthepage.rb`) I've added some conditionals to assign `` and `` to `nil` if they aren't defined at all.

In both cases, the global constant (eek!) `RECAPTCHA_ENABLED` is set to `true` only if the other two constants are not false-y. This is the variable used within the app to determine whether the ReCAPTCHA should be displayed and whether to use it in authentication.

Finally, the application config variables for ReCAPTCHA are only set if both of the key constants are not falsey; attempting to pass `nil` to this config blows up the app.

Issue #1214, opened 2018-10-12
"Mark page blank" should log a deed


benwbrum:
We do not currently log deeds whenever a page is marked blank.  This may seem trivial, but turns out to be important for the contributions API.

lortza:
Views where activity stream happen
- [ ] Edits section: collections and recent ac
- [ ] Projects Tab: activities
- [ ] Show more tab: long form activity stream
- [ ] app/views/deeds/list
- [ ] deeds helper
- [ ] collection show line 96

PR #1221, opened 2018-10-20
1220 clean up deed type strings

### LINKED ISSUE #1220: clean up deed_type (turn strings to constants)
Per @lortza : 
As I was making the Deed::... to DeedType::... replacements, I noticed that there are a lot of instances of the raw deed_type string (ex: page_trans) instead of the constant (Deed::PAGE_TRANSCRIPTION) in the codebase. It may be worthwhile to convert those to constants and to create methods for some of those longer arrays. Ex:

fromthepage/app/helpers/user_helper.rb

Line 10 in 8b8a7f0

lortza:
Closes #1220 
This PR is dependent on PR #1214  being merged into `development` first.

## Summary
This branch replaces instances of where a deed type is referred to as its string (ex: `page_trans`) instead of its constant (ex: `DeedType::PAGE_TRANSCRIPTION`). It also addresses logic for groups of deed types that are reused with methods like `DeedType.edited_pages` replacing a whole array like `[DeedType::PAGE_EDIT, DeedType::NEEDS_REVIEW, DeedType::OCR_CORRECTED ]`. The goal of these replacements is to reduce the chance of errors when listing a deed type or set of deed types. A refactor plus new testing for the `nightly_user_activity` rake task is also included.

There are 3 instances in the `specs/fixtures/deeds.yml` file that I have not been able to figure out how to convert to their corresponding constant.

## Need for Input
There are a few groups of deed types in the `app/helpers/user_helper.rb` that look like they might be redundant. I'd love some input on what deed types should be used in those places. 

## Testing for Mailers
Since part of these string replacements involved the `nightly_user_activity.rake`, we discovered there was no testing in place for the mailer. I've refactored the mailer and added testing for it. In the process, I've added config to our development environments so that we can use [MailCatcher](https://mailcatcher.me/). MailCatcher captures the emails we generate in the dev env from things like rake tasks and displays them for us in an inbox like this:
screenshot 2018-10-26 06 28 28

With MailCatcher, we can confidently work through mailer testing & debugging without worrying that our customers will be getting emails from us. Hooray! The creators of MailCatcher recommend that we do *not* put this gem in our Gemfile, but rather to install it locally. 

1.  `gem install mailcatcher`
2. run `mailcatcher` to start the mail server
3. Go to http://localhost:1080/
4. Send a mailer as you normally would and see it appear in the browser
5. Stop the server with the "Close" button in the browser

## Bug Fixed
See issue #1223 
There is some follow-up work in the production database to be done.

## Recommendation for Similar Work
While doing this work, I noticed that the `Page` model statuses are in a similar condition with their constants being used in some places and their strings being used in other places. This may be worth looking into as well.


benwbrum:
@benwbrum to do a count of distributions of `deed_type` values in production.

app/mailers/user_mailer.rb
```diff
@@ -48,15 +46,79 @@ def work_collaborator(user, work)
...
+
+    def initialize(added_works:, active_pages:, active_translations:, active_note_pages:, has_activity:)
+      @added_works = added_works
+      @active_pages = active_pages
+      @active_translations = active_translations
+      @active_note_pages = active_note_pages
+      @has_activity = has_activity
+    end
+
+    class << self
```
lortza: @benwbrum I chose to go with a `build` method instead of doing the work in the `initialize` so that we can instantiate it independently of all of that `build` logic if we have to. I have a feeling this will come in handy for mock/stub testing purposes.

app/assets/stylesheets/sections/deed.scss.erb
```diff
@@ -19,7 +19,7 @@
   $color: #B30;
   $opacity: 0.4;
   $rotate: 0deg;
-  @each $type in page_edit, page_trans, page_index, art_edit, note_add, pg_xlat, pg_xlat_ed {
+  @each $type in <%= DeedType.all_types %> {
     &-#{$type} { background: rgba(adjust-hue($color, $rotate), $opacity); }
```
lortza: Here is an example of how we've replaced arrays of hard-coded values with class method.

app/mailers/user_mailer.rb
```diff
@@ -46,17 +44,82 @@ def work_collaborator(user, work)
...
-    @user = user
-    user_activity(@user)
-    if @active_user
-      mail to: @user.email, subject: "New FromThePage Activity"
+  def nightly_user_activity(user_activity)
+    @user_activity = user_activity
+
+    if @user_activity.has_contributions?
+      mail to: @user_activity.user.email, subject: "New FromThePage Activity"
     end
```
lortza: All of the instance variables that were generated here but were hidden by a helper method have been encapsulated in a single `@user_activity` instance variable.

app/mailers/user_mailer.rb
```diff
@@ -46,17 +44,82 @@ def work_collaborator(user, work)
...
+    class << self
+      def build(user)
+        #find which pages the user has worked on
+        user_page_ids ||= user.deeds.pluck(:page_id).uniq.compact
+
+        Activity.new(
+          {
+            user: user,
+            added_works: works_added_to_users_collection_in_past_day(user),
+            active_pages: user_pages_edited_in_past_day(user, user_page_ids),
```
lortza: I've moved all of the query logic into clearly-named methods, so we have a better idea of what they're providing. The goal is make editing the components of the `Activity` object easier too.

app/mailers/user_mailer.rb
```diff
@@ -46,17 +44,82 @@ def work_collaborator(user, work)
...
+        # collections the user has worked in
+        user_collection_ids = user.deeds.pluck(:collection_id).uniq
+        # works that have been added to those collections by someone other than the user in the past day
+        Work.where(collection_id: user_collection_ids).joins(:deeds).where(deeds: {deed_type: DeedType::WORK_ADDED}).merge(Deed.past_day).where.not(deeds: {user_id: user.id}).distinct
+      end
+    end #end class << self
+
+    def has_contributions?
+      (
+        @added_works +
```
lortza: @benwbrum I noticed that there were queries in place to get `works_added_to_users_collection_in_past_day`, which we were capturing in `added_works`, but we weren't actually using that to determine if a work had contributions. This may have been a bug. I added `added_works` into that logic, which will result in emails being sent when a work is added.

app/models/deed.rb
```diff
@@ -1,75 +1,23 @@
...
-  ARTICLE_EDIT = 'art_edit'
-  NOTE_ADDED = 'note_add'
-  PAGE_TRANSLATED = 'pg_xlat'
-  PAGE_TRANSLATION_EDIT = 'pg_xlat_ed'
-  OCR_CORRECTED = 'ocr_corr'
-  NEEDS_REVIEW = 'review'
-  TRANSLATION_REVIEW = 'xlat_rev'
-  TRANSLATION_INDEXED = 'xlat_index'
-  WORK_ADDED = 'work_add'
-
```
lortza: All of these Deed "types" are moved into their own class, `DeedType`.

app/models/document_set_statistic.rb
```diff
@@ -69,23 +69,10 @@ def last_days_clause(last_days, column = "created_at")
...
-      Deed::NOTE_ADDED => 0,
-      Deed::PAGE_TRANSLATED => 0,
-      Deed::PAGE_TRANSLATION_EDIT => 0,
-      Deed::OCR_CORRECTED => 0,
-      Deed::NEEDS_REVIEW => 0,
-      Deed::TRANSLATION_REVIEW => 0,
-      Deed::TRANSLATION_INDEXED => 0,
-      Deed::WORK_ADDED => 0,
-    }
+    deeds = DeedType.generate_zero_counts_hash
```
lortza: Having methods available on the `DeedType` class lets us access all DeedTypes where and when we need them. It was pretty handy here for replacing this hard-coded hash with a shiny new method.

db/seeds.rb
```diff
@@ -5,3 +5,38 @@
...
+# contributor_deed1.user_id = contributor.id
+# contributor_deed1.save!
+#
+# contributor_deed2 = Deed.new
+# contributor_deed2.deed_type = DeedType::PAGE_TRANSLATED
+# contributor_deed2.page_id = author_page.id
+# contributor_deed2.work_id = author_work.id
+# contributor_deed2.collection_id = author_collection.id
+# contributor_deed2.user_id = contributor.id
+# contributor_deed2.save!
```
lortza: These seeds are very handy for working through Deed logic, but do not need to be in a factory and do not need to be in production. This is why they exist, but are commented out.

lib/mailer_previews/user_mailer_preview.rb
```diff
@@ -36,7 +36,7 @@ def work_collaborator
 
   def nightly_user_activity
     user = User.find_by(login: 'admin')
-    UserMailer.nightly_user_activity(user)
+    UserMailer.nightly_user_activity(UserMailer::Activity.build(user))
```
lortza: Using dependency injection here lets us be selective about the types of activities we want to use with the nightly mailer. It also makes testing _much_ easier.

spec/features/display_marked_as_blank_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    visit 'deed/list'
+    expect(page).to have_content('Page Marked Blank')
+
+    # Tear down Factories
+    Deed.destroy(deed.id)
+    Page.destroy(page1.id)
+    Work.destroy(work.id)
+    Collection.destroy(collection.id)
+    User.destroy(user.id)
+  end
```
lortza: Since we're using fixtures and not doing a database cleanup after each `it` statement, I've found that I need to generate and destroy my factory data manually each time in order to keep my tests clean. If there is a better way, I'd love to implement that.

spec/mailers/user_mailer_activity_spec.rb
```diff
@@ -0,0 +1,182 @@
+require 'spec_helper'
+
+
+RSpec.describe UserMailer::Activity do
```
lortza: Breaking the mailer's helper method out into its own class let us separate concerns when testing. Here we can test that the `Activity` class build method is doing what it should. This allows the actual mailer spec to focus on its mailing duties.

spec/mailers/user_mailer_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'spec_helper'
```
lortza: From what I read, a mailer spec just tests that the values brought to the mailer are rendered and that the mailer is delivered. This is one of the resources I used: https://www.lucascaton.com.br/2010/10/25/how-to-test-mailers-in-rails-with-rspec/

spec/spec_helper.rb
```diff
@@ -77,6 +77,13 @@
   config.asset_host = "http://localhost:3000"
 end
 
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
+end
+
```
lortza: ShouldaMatchers allow for efficient testing of model relationships, associations, delegations, etc.

benwbrum:
Closing in favor of #1265

PR #1238, opened 2018-11-06
BAD--1231 better usability for saving pages

### LINKED ISSUE #1231: better usability for saving pages
Currently we have a "Save Changes" button at the top of the transcription page.
We also have a "Submit" button on the notes field at the bottom of the transcription page.
This causes problems because some users click on the "Submit" button to save the transcription page and either get an error ("field can't be blank") because they don't have a note or they save the note but *not* the transcription like they thought.

We'd like to do 2 small UI changes to make this better:
- [x] Duplicate the "Save Changes" button on the bottom of the transcription field, just above the transcription conventions.
- [x] Change the text on the notes submission button from "Submit" to "Save Note"  

We need to visually verify this with 
- [ ] over and under and 
- [ ] side by side transcriptions; both 
- [ ] full text and 
- [ ] field based.

lortza:
Resolves #1231  [Better Usability for Saving Pages](https://waffle.io/benwbrum/fromthepage/cards/5bdb16eedda52b001b7d9d70)

## This is a bad PR that had the compromised `development` branch merged into it
Do not use.

-------------------

## Problem
Currently we have a "Save Changes" button at the top of the transcription page. We also have a "Submit" button on the notes field at the bottom of the transcription page. This causes problems because some users click on the "Submit" button to save the transcription page and either get an error ("field can't be blank") because they don't have a note or they save the note but not the transcription like they thought.

## Solution
Place additional "Save Changes" button to the bottom of the transcribe form.
screenshot 2018-11-05 21 08 08

Clarify the note form's "Submit" button's purpose by renaming it to "Save Note"
screenshot 2018-11-05 21 11 53

## Some Pain Points Resolved
The addition of the second "Save Changes" button introduced a number of collisions in the test suite. There are several views that have buttons with that text. The note form's button text change also required some careful dealings to find the correct instance of the commonly-named "Submit" button. To resolve these problems, I switched to using id selectors. If we take this approach in the future, it will save us lots of time in debugging tests.


benwbrum:
Superseded by #1239

PR #1239, opened 2018-11-07
1231 better usability for saving pages no bad merge

### LINKED ISSUE #1231: better usability for saving pages

benwbrum:
This replaces #1238 with the same code, but cherry-picked into a new branch from `development` which does not contain the bad October 23 merge.

---------------

Resolves #1231  [Better Usability for Saving Pages](https://waffle.io/benwbrum/fromthepage/cards/5bdb16eedda52b001b7d9d70)

## Problem
Currently we have a "Save Changes" button at the top of the transcription page. We also have a "Submit" button on the notes field at the bottom of the transcription page. This causes problems because some users click on the "Submit" button to save the transcription page and either get an error ("field can't be blank") because they don't have a note or they save the note but not the transcription like they thought.

## Solution
Place additional "Save Changes" button to the bottom of the transcribe form.
screenshot 2018-11-09 07 19 00

Clarify the note form's "Submit" button's purpose by renaming it to "Save Note"
screenshot 2018-11-05 21 11 53

## Some Pain Points Resolved
The addition of the second "Save Changes" button introduced a number of collisions in the test suite. There are several views that have buttons with that text. The note form's button text change also required some careful dealings to find the correct instance of the commonly-named "Submit" button. To resolve these problems, I switched to using id selectors. If we take this approach in the future, it will save us lots of time in debugging tests.


benwbrum:
This feature needs two fixes:
* Please move the new, lower "Save" button horizontally to be in line with the old, upper Save button.
* Please test the UI with over/under transcription view as well as the side-by-side view, since we noticed that the lower button seems to disappear.

lortza:
I'm getting a clean run of tests locally, but 3 of 5 Travis checks are failing with: 
```
1) uploads data for collections imports IIIF manifests
     Failure/Error: expect(page).to have_content("Metadata")
       expected to find text "Metadata" in "Please enter a valid IIIF manifest URL.×ProjectsDashboardSigned In 

...

rspec ./spec/features/add_data_spec.rb:35 # uploads data for collections imports IIIF manifests
rspec ./spec/features/needs_review_spec.rb:55 # needs review marks translated pages as blank
rspec ./spec/features/needs_review_spec.rb:67 # needs review marks pages as needing review
rspec ./spec/features/needs_review_spec.rb:93 # needs review marks translated pages as needing review
rspec ./spec/features/needs_review_spec.rb:108 # needs review filters list of review pages
rspec ./spec/features/needs_review_spec.rb:143 # needs review views collection pages that need review
rspec ./spec/features/needs_review_spec.rb:220 # needs review marks pages as no longer needing review
rspec ./spec/features/needs_review_spec.rb:238 # needs review marks translated pages as no longer needing review
rspec ./spec/features/owner_actions_spec.rb:172 # owner actions doesn't move a work with articles
```
@benwbrum Do you have any suggestions on how I can troubleshoot this locally?

PR #1254, opened 2018-11-23
Unique URL for trial signup completion

### LINKED ISSUE #1228: unique redirect for free trial
On fromthepage.com only:

Within your Google Analytics instance, there is a goal set up for the "Contact Us" form, but not the "Sign Up for a Free Trial" form. To establish this goal and track conversions, once someone signs up for a free trial they need to be directed to a link that is unique for new users (this can be accomplished by using an anchor or a parameter). Once you have this and send it my way, I can set up that goal in your analytics. I would also recommend that the user be redirected to the Dashboard as opposed to the Software page.

jackweinbender:
Closes #1228 

After a user signs up for a free trial, the user is taken to the owner dashboard with `#freetrial` appended to the URL.


jackweinbender:
Test failures are all linked to the alternate Gemfile

PR #1259, opened 2018-12-08
1245 prerender deeds

benwbrum:


saracarl:
cherry picking appropriate commits to development instead of this PR, to avoid getting fromthepage.com files (which were merged into 1245-prerender-deeds)

PR #1323, opened 2019-04-10
Disable Guest Transcription for Install

### LINKED ISSUE #1308: Disable Guest Registration
Or more appropriately "transcribe as guest"

Helping a prof out. Is there a way to disable guest registration entirely? (I understand there is a captcha setting and a way to set the number of saves required, so 0 would require immediate registration or infinite?) But trying to limit the number of pharmacies and other sales organizations around the world we support.

(enhancement perhaps, still learning the vocabulary of the application, but the code is helpful, so rather than restrict work to a set of scribes, what might also be helpful, in addition, is to restrict work to all registered scribes for when guests are allowed?)

Thanks

jackweinbender:
This PR closes #1308 

FTP traditionally has allowed guest transcribers for all projects that are not private or restrict scribes. This PR changes the default behavior to disallow guest transcription by default. To enable guest transcription, the flag `GUEST_TRANSCRIPTION_ENABLED` must be set to `true` (it is `false` by default).

Formerly, future-scribes would see this page on when they first attempted to transcribe:
image

The above is still the page future-scribes are greeted with when `GUEST_TRANSCRIPTION_ENABLED` is `true`. Otherwise (and by default) future scribes now see this:

image

Because transcribing as a "guest" actually makes a guest account, the route authorization filters did not need to be changed; in this case, since the user is not able to "login as a guest" they cannot get to the *actual* transcribe page.


spec/features/guest_spec.rb
```diff
@@ -11,7 +11,26 @@
...
 
-  it "tests guest account creation and migration" do
+  before :each do |test|
+    if test.metadata[:guest_enabled]
+      # Guest Transcription is no longer the default and is enabled via
+      # a global flag in an initializer. To test guest processes,
+      # we must enable the flag for those (old) tests. New tests are not
+      # be affected.
+      silence_warnings do
+        GUEST_TRANSCRIPTION_ENABLED = true
```
benwbrum: Today I learned about `silence_warnings`.

PR #1332, opened 2019-04-15
Export formatting for works with indexeing disabled

### LINKED ISSUE #1277: double brackets not exporting (no subject linking)
We decided to use double brackets to mark non-textual elements on a page, for example [[official seal]] or [[postcard]].  We have disabled  "subject indexing" but have found that the double brackets aren't exporting with the text. We simply get postcard and not [[postcard]].

We would like to be able to use double brackets in this way if possible. Is this something that can be addressed?

jackweinbender:
Closes #1277 

I was able to accomplish this without completely refactoring the `wiki_to_xml` method into two. In fact, the final solution turned out to be quite simple. I was really hesitant to make any significant changes to this section because of how core this functionality is; we didn't have any tests, so I couldn't feel confident I wasn't regressing.

But, we have some tests now. One minor change I made was to change the method signature to accept the whole page object rather than only the wiki text. This allowed a more clean way to reference the page's collection object in order to determine if subject linking was disabled.

The Unit tests are simple; they basically just test that the XML output is what we expect. I had to assume that the former functionality was correct, but this will help prevent regressions, anyway.

The integration tests test both the display (reading) view of the page as well as the HTML export view.

### DIFF
--- a/app/models/xml_source_processor.rb
+++ b/app/models/xml_source_processor.rb
@@ -91,15 +91,15 @@ module XmlSourceProcessor
     end
   end
 
-  def wiki_to_xml(wiki, text_type)
+  def wiki_to_xml(wiki, text_type, subjects_disabled=false)
     xml_string = String.new(wiki || "")
     xml_string = process_latex_snippets(xml_string)
     xml_string = clean_bad_braces(xml_string)
-    xml_string = process_square_braces(xml_string)
+    xml_string = process_square_braces(xml_string) unless subjects_disabled
     xml_string = process_linewise_markup(xml_string)
     xml_string = process_line_breaks(xml_string)
     xml_string = valid_xml_from_source(xml_string)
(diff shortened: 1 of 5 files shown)


benwbrum:
see issue for refactoring/changes that need to be done with this PR. #1277

app/models/xml_source_processor.rb
```diff
@@ -83,33 +83,39 @@ def validate_links(text)
...
     postprocess_sections
     xml_string    
   end
 
-  def generate_preview(text_type)
-    xml_string = wiki_to_xml(self.source_text, text_type)
-    xml_string = update_links_and_xml(xml_string, true, text_type)
-    return xml_string
-  end
-
```
saracarl: Was this dead code?  Or did it end up somewhere else?  The preview works, so I don't think it's a problem.

PR #1336, opened 2019-05-01
Untranscribed Works for Getting Started Path

### LINKED ISSUE #1324: "Find Untranscribed Works" button on collections
We've been getting feedback from people who run large projects that volunteers are having trouble finding works that aren't "claimed" but other transcribers.  Maryland, for instance, says that volunteers are hesitant to transcribe in a work that someone else is already transcribing in.  

I think instead of "Start transcribing" on the top right of the collection home screen, we should have a "Find Untranscribed Works" -- it's clearer what it will show you.  It should give you a page similar to the collection screen, but with works that have *no* transcription listed first (only or first??)   That way, transcribers can find something "unclaimed" to work on, but they still get to choose from a list.

So what do we do for collections that don't have any untranscribed works?  Were works are partially transcribed?  Do we list least to most transcribed?  Do we show least-recently-active first?  Do we change the button name to something else?  This needs more thought and discussion.

### LINKED ISSUE #1325: start transcribing on owner landing page.
Review #994 #949 and #1324 as part of this.  We need a "get started" path on the owner home page as well as the collection home page.

jackweinbender:
Closes #1324 
Closes #1325 

Todo:
- [x] Provide `untranscribed works` route
- [x] Update Collection `Start Transcribing` Button
- [x] Add `Find Untranscribed Works` button to owner home page
- [x] Tests

We've updated this PR to the following:

## Collection Page
On the collection page, the `Find Untranscribed Works` button now takes you straight to the first page of an entirely untranscribed work within the collection.

If there are no completely untranscribed works, we present the first untranscribed page of the most-stale work in the collection (the work that has least-recently been edited).

image

## Owner Page
We also have included this button within each collection-details section on the owner page

image

### Additional note:
The tests for this are passing and OK. But, it made two things very clear to me: 1) we need to isolate our feature tests better. 2) we should work to reduce fixtures---particularly those that represent multi-relational things (Collections->works->pages). We should probably be moving many of them (perhaps not all) to factories.


app/views/collection/show.html.slim
```diff
@@ -12,11 +12,10 @@
 
     div.collection-work-list
       h2 Works
-      -pages = find_transcribe_pages
 
       .start_link
-        -unless pages.blank?
-          =link_to "Start Transcribing", collection_start_transcribing_path(@collection.owner, @collection), class: 'button'
+        -if @collection.has_untranscribed_pages?
+          =link_to "Find Untranscribed Works", collection_start_transcribing_path(@collection.owner, @collection), class: 'button'
```
saracarl: Let's change the text back to "Start Transcribing" but leave the new logic.

app/views/dashboard/_alphabetical_collections_and_document_sets.html.slim
```diff
@@ -10,6 +10,9 @@
                 =image_tag(cds.picture_url(:thumb), alt: cds.title)
             .collection_details
               h4.collection_title =link_to cds.title, collection_path(cds.owner, cds)
+              -if cds.has_untranscribed_pages?
+                p
+                  =link_to "Find Untranscribed Works", collection_start_transcribing_path(cds.owner, cds), class: 'button'
```
saracarl: Same here.  Change the button text to "Start Transcribing"

spec/features/editor_actions_spec.rb
```diff
@@ -309,14 +309,8 @@
...
     expect(page).to have_content("Works")
-    expect(page).not_to have_selector('a', text: "Start Transcribing")
-    col = Collection.third
-    visit collection_path(col.owner, col)
-    expect(page).to have_selector('h1', text: col.title)
-    expect(page).to have_content("About")
-    expect(page).to have_content("Works")
-    expect(page).to have_selector('a', text: "Start Transcribing")
-    click_link("Start Transcribing")
+    expect(page).to have_selector('a', text: "Find Untranscribed Works")
```
saracarl: Change here.

spec/features/editor_actions_spec.rb
```diff
@@ -309,14 +309,8 @@
...
+    click_link("Find Untranscribed Works")
```
saracarl: And here.

PR #1353, opened 2019-06-05
Add translate from Image feature

jackweinbender:
This feature provides a toggle to show page images rather than the transcription for translations. The toggle preference persists via a cookie.

The default state remains the same:
image

Clicking on the "Show Image" button displays the page (I don't have images to load in my test env, hence the empty box and spinner):
image


benwbrum:
This works pretty nicely!

PR #1354, opened 2019-06-06
Add http mocking libraries for select iiif api calls

### LINKED ISSUE #1148: Use mocks for testing resources we don't control
Especially IA imports and IIIF Imports.

jackweinbender:
Closes #1148

I've added two gems (that work together) to mock-out API calls:

    gem 'webmock', require: false
    gem 'vcr'

[Webmock](https://github.com/bblimke/webmock) intercepts all http calls in our test suite, and [VCR](https://github.com/vcr/vcr) which 'records' (saves as yaml) and 'plays back' (uses mocks for) http reqs during tests. So, after the first run of the test suite---having saved the real API responses---it uses the mocks. This is quite a bit faster, too.

I've gone for an "opt-in" approach to using these libraries, since we make a good number of API calls. For this PR, I've only mocked-out the IIIF endpoints that were failing. In fact, I actually changed the IIIF endpoint to one that is currently working, so the mocks are not really necessary. That said, since this isn't the first time one of our external sources went sideways, I suppose it's good to have the infrastructure ready for when it (inevitably) happens again.

We can also be a bit more intentional and work to remove ALL external API calls from our tests. Apart from removing the network dependency, it will also make our tests run quite a bit faster (I think).


spec/features/zz_convention_spec.rb
```diff
@@ -85,9 +85,6 @@
     expect(page.find('#work_transcription_conventions')).to have_content @work_convention
     expect(page).to have_button('Revert')
     page.find_button('Revert').trigger(:click)
-    sleep(3)
-    convention_work = Work.find_by(id: @work.id)
```
saracarl: Why are we getting rid of this? (the next line)

spec/features/zz_convention_spec.rb
```diff
@@ -85,9 +85,6 @@
...
-    expect(convention_work.transcription_conventions).to eq nil
```
saracarl: Why are we getting rid of this?

saracarl: Covered in (new) line 94 with a functional test rather than a db check.

PR #1356, opened 2019-06-07
Accessibility Compliance

### LINKED ISSUE #1346: UI contrast for accessibility
We need to increase the color contrast for accessibility.  See the email from UT.

jackweinbender:
Closes #1346 

The email about accessibility compliance lists three things for us to look at. Contrast, Form Labels, and Language attributes.

I worked hard on this, but without the ability to run it through the (paid) linter that they used, I was stuck clicking around the site looking for errors. So, perhaps after merging this, we may ask UT to run it again? 

# Changes

## Contrast
I tried to be as light-handed as possible making these edits. Where I had to be more aggressive, I tried to maintain the look and feel of the site as best as I could. The two "major" changes I made (viz. more than just tweaking the text-color to be a bit darker) were to the Deed label coloring and to the transcription preview/autolink buttons.

## Form Labels
I *think* I've got a good label for all form elements. I confess, I'm not sure that I looked at every form, but I've looked at most of them and linted them with an a11y tool.

## Aria Labels
I was only able to find two places that needed aria labels, the close "X" link used in the Litebox plugin (most other symbols have text, which negate the need for aria labels, as I understand it), and the prev/next navigation buttons.

## Tab Indexing
This one wasn't on the a11y note, but I noticed it while I was working. Basically, we should be able to navigate around the site using the keyboard and that wasn't really possible before. One of the main problems was that all the `focus` state visual cues had been overridden in our stylesheet to be invisible. I added some subtle outlining for the `focus` state.

This also required that I make some additions to the JavaScript that toggled the menus. Since the event listener was only set on `click` tabbing into the menus wasn't possible (both the main `login/logout` menu as well as the `Actions` menus). So I was able to add event listeners for `focusin` that open and close the menus at the appropriate times. Because the menus were made with Definition Lists, they additionally needed to be explicitly marked as tabbable with `tabindex=0`.

## Lang Attributes
After we got that email back, I was able to go look for what the gentleman was talking about. The problem was that we use the `ISO-639-2` style (three-letter) codes and these a11y linters only like the two-letter versions (`ISO-639-1`).

Since our database is full of three-letter codes, I thought it best to only change how we display the codes. We already had the `ISO-639` library installed, so, I was able to use that to make the changes. The problem with this approach is that `ISO-639-1` has many fewer languages in its standard (the reason, I suspect, we opted for the three-letter version!). Since many of these excluded languages seem to be the dead variety, I figured we'd still need them, so, in cases where there is no two-letter code, the three-letter version is used instead. I thought this was better than leaving off the `lang` attribute altogether and no worse than `lang=""`.

** UPDATE**
This is all wrong. There is a new standard, it seems, laid out in BCP 47 that provides even more information than either of the ISO639 standards. The list of codes can be found here. This standard seems to have the comprehensive coverage of the standard we are using, but uses many two-letter codes that look like the older standard.

To make things more complicated, it seems that the Library of Congress uses the newer ISO standard, so there is, perhaps, good reason for us to retain that standard in our database (especially).

But, for displaying HTML, it seems the official answer is that we should be converting these ISO639-2 codes into IANA subtags.

## Other
I noticed a couple minor CSS/HTML errors, too, which I fixed. One being the extra white bar on the bottom of the footer where the "contact us" link is (on all pages!). Another was that empty `<th>` tags make the a11y linter complain, so I changed it to a `<td>` and added a little CSS to fix the appearance.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -77,6 +77,7 @@ $.fn.dropdown = function(s) {
     $trigger.on('mousedown.DropDown', function() {
       $element.toggleClass(cfg.openclass);
     });
+    //Open on focus
     $trigger.on('focusin.DropDown', function() {
       $element.addClass(cfg.openclass);
     });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -74,8 +74,11 @@ $.fn.dropdown = function(s) {
     var $trigger = $(cfg.trigger, this);
 
     // Open/close when clicked on the trigger
-    $trigger.on('click.DropDown', function() {
+    $trigger.on('mousedown.DropDown', function() {
       $element.toggleClass(cfg.openclass);
+    });
+    $trigger.on('focusin.DropDown', function() {
+      $element.addClass(cfg.openclass);
     });
 
     // Close when clicked on an item
@@ -88,6 +91,12 @@ $.fn.dropdown = function(s) {
 
     // Close if clicked outside
     $(document).on('click.DropDown', function(e) {
+      if($(e.target).closest($element).length === 0) {
+        $element.removeClass(cfg.openclass);
+      }
+    });
+    // Close if focus leaves
+    $(document).on('focusin.DropDown', function(e) {
       if($(e.target).closest($element).length === 0) {
         $element.removeClass(cfg.openclass);
       }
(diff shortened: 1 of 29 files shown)


jackweinbender:
We decided that for now we will use the ISO-639-1 codes, then use an empty string `""` for languages that aren't in that standard, making them "unknown."

Issue #1366, opened 2019-06-14
Ahoy not logging visits


benwbrum:
The last visit logged to production was May 13, immediately after deploying PR #1333 to fix #1314.  We need to revert this fix quickly and reopen #1314 once it's deployed.

PR #1367, opened 2019-06-14
Log visits to ahoy again

benwbrum:


benwbrum:
This fixed issue #1366

PR #1383, opened 2019-07-04
Allow multiline notes

### LINKED ISSUE #1375: Notes shouldn't strip out line breaks
If you hit a carriage return in a note, we should respect that as a line break in the note.

Test to make sure exports and activity streams still work right after fixing this.

Elliot N. for SRST reported this.

jackweinbender:
Closes #1375 

This PR allows users to submit multiline notes. It uses the `simple_format` rails helper in conjunction with the `sanitize` helper. It does not change how the notes are saved in the DB; this only affects how the note is displayed.

image

The one problem I had was that the `update` function was only handled on the client, which made it so edited notes  wouldn't be processed by these same functions. The old update function simply returned a `204` code and left it to the client to render the change. In the new version. the server returns a success code and a re-rendered partial of the note, which then replaces the old version of the note.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -1,5 +1,5 @@
 class NotesController < ApplicationController
-
+  include ActionView::Helpers::TextHelper
   def create
     @note = Note.new(params[:note])
     # truncate the body for the title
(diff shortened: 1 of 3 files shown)


saracarl:
@jackweinbender can we just revert the test to what it was before?   I know we won't be testing multi-line notes, but I think the feature is small enough (vs. the time it will take to get the javascript testing working) that I'm OK with us skipping that test.

jackweinbender:
Ok, no trouble. I can do that. I'll try to revert the tests this evening so you can merge tomorrow.

PR #1392, opened 2019-07-11
feature 898 add new export formats to export tab

diegoviola:


spec/features/export_spec.rb
```diff
@@ -34,13 +34,22 @@
...
     expect(page).to have_content(@page.title)
   end
 
+  it "exports a work as plain text" do
+    visit "/export?collection_id=#{@collection.id}"
+    expect(page).to have_content("Export Individual Works")
+    page.find('tr', text: @work.title).click_link("Plain text")
+    expect(page.current_path).to eq ("/export/work_plaintext_verbatim")
+    expect(page.all('pre', text: @work.title))
+    expect(page.all('pre', text: @page.title))
```
saracarl: Why is this line repeated?

saracarl: ah, work.title vs page.title

PR #1399, opened 2019-07-13
iiif_controller.rb: skip over generating a canvas

### LINKED ISSUE #1326: IIIF manifests give 500 errors
Several of the Julia Brumfield Diaries manifests are giving errors, including returning the wrong item and producing 500 errors.

The following manifests yield 500 errors:
* https://fromthepage.com/iiif/17/manifest
* https://fromthepage.com/iiif/6/manifest

diegoviola:
Fixes #1326


app/controllers/iiif_controller.rb
```diff
@@ -562,23 +562,25 @@ def canvas_from_ia_page(page)
...
-    canvas.width = page.base_width
-    canvas.height = page.base_height
-    canvas['@id'] = canvas_id_from_page(page)
-
-    annotation = IIIF::Presentation::Annotation.new
-    annotation.resource = iiif_create_image_resource(page)
-    annotation['on'] = canvas['@id']
-    annotation['@id'] = "#{url_for(:root)}image-service/#{page.id}"
-    canvas.images << annotation
+    unless page.base_width.nil? && page.base_height.nil?
```
saracarl: This should be an "or" rather than an "and" -- if either is nil, then don't generate a canvas.

diegoviola:
> So what happens to the calling code if "canvas_from_page" returns nil? (Because we definitely could return nil in this code, and I doubt that the calling code handles that nil.)

The `sequence.canvases` array (line 402) gets populated with nil values, I didn't notice any errors when I tested.

saracarl:
Diego, do you know why this change would have resulted in a 25% increase in page load times?

So the working example:  http://localhost:3000/iiif/9/manifest
takes 28.18 seconds to load. 

```Oink statistics:

Completed 200 OK in 26360ms (Views: 21.0ms | ActiveRecord: 4748.4ms)
Oink Action: iiif#manifest
Memory usage: 591608 | PID: 3171
Instantiation Breakdown: Total: 2050 | User: 746 | Work: 374 | Collection: 374 | Page: 372 | ActiveRecord::SchemaMigration: 175 | Note: 5 | Visit: 2 | WorkStatistic: 1 | Ahoy::Event: 1
Oink Log Entry Complete
```

In the development branch, the same request takes
20.44 seconds to load
```Completed 200 OK in 18322ms (Views: 10.9ms | ActiveRecord: 4430.2ms)
Oink Action: iiif#manifest
Memory usage: 591708 | PID: 3319
Instantiation Breakdown: Total: 2050 | User: 746 | Work: 374 | Collection: 374 | Page: 372 | ActiveRecord::SchemaMigration: 175 | Note: 5 | Visit: 2 | WorkStatistic: 1 | Ahoy::Event: 1
Oink Log Entry Complete```

The same number of objects are being instantiated, so the Oink info isn't useful -- memory is the same.  That implies the difference is in requests or iterations over objects.

saracarl:
I tried to multiple times with curl on the command line.  (which let me time it, too) and there it just takes 6-8 seconds.  Before this PR, it took 4-5, but still.  6-8 seconds is much lower than the ~25 seconds I'm seeing in Chrome.  I think this is caused by that weird "let's load the collections page when we do API queries" problem we're seeing -- #1408 .

PR #1410, opened 2019-07-24
1081 contributors in iiif export

diegoviola:


app/models/page.rb
```diff
@@ -358,6 +358,20 @@ def update_work_stats
     end
   end
 
+  def contributors(page_id)
+    users = []
+
+    page = Page.find(page_id)
```
benwbrum: Since this is an instance method, it shouldn't need page_id as a parameter.  Replacing this line with `page = self` and removing `page_id` from the parameter list (and calling function) would be much nicer.

diegoviola: @benwbrum Good catch, I made the change. Thanks.

PR #1414, opened 2019-07-30
dashboard: fix collections_list performance

diegoviola:
By not calling Collection.show_to? and #has_untranscribed_pages? inside
a each loop (mariadb is flooded with lots of queries, thus creating a
bottleneck).

Fixes: #1175.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -46,9 +46,7 @@ class DashboardController < ApplicationController
   end
 
   def collections_list
-    collections = Collection.includes(:owner).distinct
-    document_sets = DocumentSet.includes(:owner).distinct
-    @collections_and_document_sets = (collections + document_sets).sort { |a,b| a.title <=> b.title }
+    @collections = Collection.all
   end
 
   #Owner Dashboard - start project
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -49,7 +49,7 @@ class DashboardController < ApplicationController
     if user_signed_in?
       @collections = Collection.by_user(current_user) + DocumentSet.by_user(current_user)
     else
-      @collections = Collection.where(restricted: false) + DocumentSet.where(restricted: false)
+      @collections = Collection.where(restricted: false) + DocumentSet.where(is_public: true)
     end
   end
 
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -46,7 +46,11 @@ class DashboardController < ApplicationController
   end
 
   def collections_list
-    @collections = Collection.all
+    if user_signed_in?
+      @collections = Collection.by_user(current_user) + DocumentSet.by_user(current_user)
+    else
+      @collections = Collection.where(restricted: false)
+    end
   end
 
   #Owner Dashboard - start project
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -49,7 +49,7 @@ class DashboardController < ApplicationController
     if user_signed_in?
       @collections = Collection.by_user(current_user) + DocumentSet.by_user(current_user)
     else
-      @collections = Collection.where(restricted: false)
+      @collections = Collection.where(restricted: false) + DocumentSet.where(restricted: false)
     end
   end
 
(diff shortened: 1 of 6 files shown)


benwbrum:
The view partial `alphabetical_collections_and_document_sets` is called by `collection_list` but also by `owner#profile`.  We need to make sure that the owner profile still displays as it did before, including document sets.  

(I assume that we can still pass in `collections_and_document_sets` from the `owner#profile` action, and that this new code will work -- we just need to make sure it does.)

Users to test this on include:
`lva` (which has hierarchical collections), `adah` (which does not), and `alourie`.

benwbrum:
One additional consideration we never thought about when deciding to remove deeds and document sets from the collection list page was what happens for single-tenant sites like [NYBG](http://97.107.141.53/collections) or [UT-Austin](https://fromthepage.lib.utexas.edu/collections) as well as other institutions that have installed the tool.  For those installations, the collection list page is the main page letting users find collections to work on or see what's happening on the site (rather than `findaproject` on multi-tenant sites like fromthepage.com).  

Here's what I think we should do (or should have done) to support these sites *and* support multi-tenant use cases on `fromthepage.com`:

* Create a new action `basic_collection_list` to support removed functionality.
* Move the code that removes document sets there.  Have that action pass an additional parameter (like `:basic => true` to the view)
* Create another new action `full_collection_list` to  match functionality of the old collection list. 
* Move the old code that fetches document sets and collections to this view.  Use the same logic that avoids calling `show_to?` (as you did already), and add similar logic for document set collaborators. 
* Render the view with some flag like `:basic => false`
* Conditionalize the view to display the deed list in full mode but not basic mode.
* Change the existing `collection_list` action to test if `Collection.count > 30` then return `basic_collection_list` else return `full_collection_list`

diegoviola:
> The view partial `alphabetical_collections_and_document_sets` is called by `collection_list` but also by `owner#profile`. We need to make sure that the owner profile still displays as it did before, including document sets.
> 
> (I assume that we can still pass in `collections_and_document_sets` from the `owner#profile` action, and that this new code will work -- we just need to make sure it does.)
> 
> Users to test this on include:
> `lva` (which has hierarchical collections), `adah` (which does not), and `alourie`.

Good point and thanks for bringing this up. I think I was quick with removing the document sets (not realizing it was being used in different places).

I added it back so the user profile is working again.

diegoviola:
I'll try taking a different approach to this problem, because removing these two methods break existing tests. Closing.

app/controllers/dashboard_controller.rb
```diff
@@ -46,9 +46,11 @@ def index
   end
 
   def collections_list
-    collections = Collection.includes(:owner).distinct
-    document_sets = DocumentSet.includes(:owner).distinct
-    @collections_and_document_sets = (collections + document_sets).sort { |a,b| a.title <=> b.title }
+    if user_signed_in?
+      @collections = Collection.by_user(current_user) + DocumentSet.by_user(current_user)
+    else
+      @collections = Collection.where(restricted: false)
```
benwbrum: I suspect line 52 should be `@collections = Collection.where(restricted: false) + DocumentSet.where(restricted: false)`

diegoviola: @benwbrum I just tried your suggestion (thanks), but it fails because DocumentSet doesn't have a `restricted` field.

app/models/collection.rb
```diff
@@ -66,6 +66,10 @@ def show_to?(user)
     (!self.restricted && self.works.present?) || (user && user.like_owner?(self)) || (user && user.collaborator?(self))
   end
 
+  def self.by_user(user)
+    Collection.includes(:owners).where(owner_user_id: user.id).joins(:collaborators)
```
benwbrum: Okay, there are two different N:N mappings between `users` and `collections`.  This one should only look at collaborators, so `Collection.joins(collaborators).where("collection_collaborations.user_id = ?", user.id)` is probably the query we want.

app/models/document_set.rb
```diff
@@ -21,6 +21,10 @@ class DocumentSet < ActiveRecord::Base
   def show_to?(user)
     self.is_public? || (user && user.collaborator?(self)) || self.collection.show_to?(user)
   end
+
+  def self.by_user(user)
+    DocumentSet.includes(:owner).where(owner_user_id: user.id).joins(:collaborators)
```
benwbrum: See my comment above about collaborators instead of owners on collections -- the same thing should be true on document sets.

app/models/collection.rb
jackweinbender: I may be wrong, but I think there's a nice shorthand for this on the user model
`user.collection_collaborations`

diegoviola: @benwbrum @jackweinbender I tried both of your suggestions, but the tests continue to fail.

diegoviola:
Closing in favor of  #1420.

PR #1416, opened 2019-07-30
Enable/Disable OCR of all works from collection

jackweinbender:


benwbrum:
Can we conditionalize the display of the Enable OCR/Disable OCR buttons based on `@collection.works.where(:ocr_enabled => true).present?` and the inverse?

jackweinbender:
I'm doing some profiling on this, and want to record the results as they come.

The naive approach that I put forth originally (not realizing the other callbacks were getting called, making it an O(n^2) time complexity) after a few runs with Rack Mini Profiler against one of the larger ADAH collections (500+ works) is showing pretty abysmal (if anticipated) results:

## Naive
Queries: ~15,000 (30 seconds), with a total time of 80-90 seconds on my 8-core desktop.

app/views/collection/edit.html.slim
```diff
@@ -169,6 +169,16 @@
     -else
       =link_to({ :action => 'disable_document_sets', :collection_id => @collection.id }, class: 'button')
         span Disable Document Sets
+    
+    h3 OCR Correction
+    p Enable or disable OCR correction for all works in this collection.
+    -if @collection.works.any? { |w| w.ocr_correction.blank? }
```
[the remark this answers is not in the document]
jackweinbender: `.any?` on an active record relation fires off a `COUNT` query, so this isn't the same as the array method that (I think) you are referencing.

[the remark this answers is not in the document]
jackweinbender: I've added a scope for `works` that is `ocr_enabled/disabled` and now use that scope with `any?` (without the block, so it's just a DB query, as you suggest).

app/views/collection/edit.html.slim
```diff
@@ -169,6 +169,16 @@
...
       =link_to({ :action => 'disable_document_sets', :collection_id => @collection.id }, class: 'button')
         span Disable Document Sets
+    
+    h3 OCR Correction
+    p Enable or disable OCR correction for all works in this collection.
+    -if @collection.works.any? { |w| w.ocr_correction.blank? } 
+      =link_to({ :action => 'enable_ocr', :collection_id => @collection.id }, class: 'button')
+        span Enable OCR
+    ' 
+    -if @collection.works.any? { |w| !w.ocr_correction.blank? }
```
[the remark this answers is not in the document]
jackweinbender: Added migration to this effect.

app/models/collection.rb
```diff
@@ -162,6 +162,54 @@ def has_untranscribed_pages?
     self.works.joins(:work_statistic).where('work_statistics.total_pages > work_statistics.transcribed_pages').exists?
   end
 
+  def update_works_stats
+    works = self.works.includes(:work_statistic)
+    works_stats = get_works_stats_hash
+    works.each do |w|
+      w.work_statistic.recalculate_from_hash(works_stats[w.id]) if works_stats.key?(w.id)
```
benwbrum: What happens to work statistics if there is no work_stats hash containing the work key?  Are there any conditions under which that might happen?

benwbrum: One option might be to have `Work#recalculate_from_hash` simply test for `nil` in its argument and fall-back to recalculating from the database.

app/models/collection.rb
```diff
@@ -162,6 +162,54 @@ def has_untranscribed_pages?
...
+    end
+    calculate_complete
```
benwbrum: What is `calculate_complete`?  I don't see it anywhere else in this file.

app/models/collection.rb
```diff
@@ -162,6 +162,54 @@ def has_untranscribed_pages?
...
+  end
+
+  def disable_ocr
+    works.update_all(ocr_correction: false)
+    update_works_stats
+  end
+
+  def get_works_stats_hash
+    stats = {}
+    work_prototype = {
```
benwbrum: If this is immutable, should it be a constant `WORK_PROTOTYPE` or similar?

app/models/collection.rb
```diff
@@ -162,6 +162,54 @@ def has_untranscribed_pages?
...
+      translation: {},
+      total: 0
+    }
+    work_ids = works.ids
+    transcription = Page.where(work_id: work_ids).group(:work_id, :status).count
+    translation = Page.where(work_id: work_ids).group(:work_id, :translation_status).count
+    totals = Page.where(work_id: work_ids).group(:work_id).count
+    
+    transcription.each do |(id, status), value|
+      stats[id] = work_prototype unless stats.has_key?(id)
```
benwbrum: I'm not sure why I dislike testing on `unless stats.has_key?(id)` instead of `unless stats[id]` or `if stats[id].nil?`

A shorter alternative is `stats[id] ||= work_prototype`, although I've sometimes found that hard to read.

app/models/collection.rb
```diff
@@ -162,6 +162,54 @@ def has_untranscribed_pages?
...
+      stats[id][:translation][status] = value
+    end
+
+    totals.each do |id, value|
+      stats[id] = work_prototype unless stats.has_key?(id)
+      stats[id][:total] = value
+    end
+    stats
+  end
```
benwbrum: I'd be really tempted to make this block more concise like this:
```
work_ids.each do |work_id|
  stats[id]=work_prototype  # consider a copy/clone/dup if this is immutable or we are overwriting it?
  stats[id][:total]=transcription[id]
  stats[id][:translation]=translation[id]
  stats[id][:total]=totals[id]
end
```

It's just style, but I think it makes it a bit clearer.

app/models/work_statistic.rb
```diff
@@ -76,59 +75,49 @@ def pct_translation_completed
...
+      translation: work.pages.group(:translation_status).count,
+      total: work.pages.count
+    }
+  end
+  private
 
+  # current logic to recalculate statistics for parent document set and parent collection
+  def recalculate_parent_statistics
+    # save completed information for collections/document sets
+    work.collection&.calculate_complete
```
benwbrum: Oooh -- what's that `&.` doing there?

app/models/work_statistic.rb
```diff
@@ -76,59 +75,49 @@ def pct_translation_completed
...
+    unless work.document_sets.empty?
+      work.document_sets.each(&:calculate_complete)
```
benwbrum: And that `&:`, too?

db/migrate/20190902182932_set_default_ocr_correction_value.rb
```diff
@@ -0,0 +1,8 @@
+class SetDefaultOcrCorrectionValue < ActiveRecord::Migration
+  def change
+    # Set default value to false
+    change_column :works, :ocr_correction, :boolean, default: false
+    # Retroactively set NULL to false
+    Work.where(ocr_correction: nil).update_all(ocr_correction: false)
```
benwbrum: Check me on this, but isn't this unnecessary once the default is set on the column?

benwbrum: I was mistaken: in MySQL/MariaDB, this sets the default for new rows but does not update values in existing rows.

benwbrum:
I've tested this locally, and the performance is excellent now.

app/models/work_statistic.rb
jackweinbender: Here it's beign used as a "[Safe Navigation Operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/)"... though I don't know why I would have put it here.

app/models/work_statistic.rb
jackweinbender: Here it's being used as a stand in for `.each(|d| d.calculate_complete)`. Just a shorthand.

app/models/collection.rb
jackweinbender: This is actually the `Collection`s version of `recalculate`. it is defined in the `CollectionStatistics` module.

app/models/collection.rb
jackweinbender: Agreed. Fixed.

app/models/collection.rb
jackweinbender: opted for `if stats[id].nil?`, because I think it is nice and readable and communicates the intent of the code well.

app/models/collection.rb
jackweinbender: I think that guard statement `is meant only to prevent any major catastrophes in the event that (somehow) the key doesn't exist. It  *should* in every case. That said, I think your approach is better and makes the `recalculate_from_hash` a more usable helper function. For example, as you suggest:
```
def recalculate_from_hash(stats=nil)
    stats = get_stats_hash if stats.nil?

    self[:total_pages] = stats[:total]
...
```
Also allows the old `recalculate` function to go from:
```
def recalculate(_options = {})
    stats = get_stats_hash
    recalculate_from_hash(stats)
    recalculate_parent_statistics
  end

```
to
```
  def recalculate(_options = {})
    recalculate_from_hash
    recalculate_parent_statistics
  end
```
And, of course, makes the guard statement unnecessary.

I've lightly refactored `get_works_stats_hash`, however,  to make the explicit mapping between the `works` and the `stats_hash` by passing in a list of `ids` rather than generating this list within the method's body. Thus, by definition, the keys in the stats hash are those of the works list.

app/models/collection.rb
jackweinbender: I think there's actually a bit more going on here that you may have missed. The resulting values of the `transcription` and `translation` queries are grouped by `work.id` AND by `work.status` to the result is a somewhat de-normalized list of counts like this with the format `[work_id, status] => count` :
```
[15, 'needs_review'] => 5
[15, 'transcribed'] => 10
... 
```
So, it's not a nested hash, like you'd expect (I think this is what your example is assuming):
```
15 => {
  'needs_review' => 5,
  'transcribed' => 10
}
```
I certainly could rewrite the code as the above, but I'd need to write a helper function to convert the query results into the right format first.

jackweinbender:
I also note that the "Works statistics" PR was, infact, a strict subset of this PR and I have closed it. All the work I did in that PR is included here.

PR #1419, opened 2019-08-02
Contentdm collection import

### LINKED ISSUE #1412: Allow CONTENTdm Collection Import
Currently, we can only import individual `items` (`works` in FTP terms) into a FTP `collection` from CONTENTdm. We should be able to import full collections from CONTENTdm into FTP.

jackweinbender:
Closes #1412 

## Purpose
CONTENTdm URLs would only be considered valid for "works". This meant the IIIF manifests were always of type `sc:Manifest`. Adding the ability to browse and import collections was only a matter of accepting and pointing collection- and repository- level URLs to valid IIIF `manifest.json` files. The infrastructure to display and navigate with those `sc:Collection` type manifests is already built-in to FTP, so I didn't have to do anything special. It Just Worked™.

## Functionality
All resulting IIIF URIs have been updated to the current description of the [IIIF Presentation API on CONTENTdm's site](https://help.oclc.org/Metadata_Services/CONTENTdm/Advanced_website_customization/API_Reference/IIIF_API_reference)

### Full item-level URLs (No change)
ex. `https://cdm123.contentdm.oclc.org/digital/collection/COL1D/id/123/` yeilds
`https://cdm123.contentdm.oclc.org/iiif/info/COL1D/123/manifest.json`
image

### Collection-level URLs
ex. `https://cdm123.contentdm.oclc.org/digital/collection/COL1D/` yeilds
`https://cdm123.contentdm.oclc.org/iiif/info/COL1D/manifest.json`
image

### Repository-level URLs
ex. `https://cdm123.contentdm.oclc.org` yeilds
`https://cdm123.contentdm.oclc.org/iiif/info/manifest.json`
image

### DIFF
--- a/lib/contentdm_translator.rb
+++ b/lib/contentdm_translator.rb
@@ -181,13 +181,18 @@ module ContentdmTranslator
   end
 
   def self.cdm_url_to_iiif(url)
-    matches = url.match(/https?:\/\/(cdm\d+).*collection\/(\w+)\/id\/(\d+)/)
-    if matches
-      server = matches[1]
-      collection = matches[2]
-      record = matches[3]
-      
-      "https://#{server}.contentdm.oclc.org/digital/iiif-info/#{collection}/#{record}/manifest.json"
+    matches = url.match(/https?:\/\/(cdm\d+)(?:.*collection\/(\w+)(?:\/id\/(\d+))?)?/)
+    
+    server = matches[1]
+    collection = matches[2]
+    record = matches[3]
+    
+    if server && collection && record
+      "https://#{server}.contentdm.oclc.org/iiif/info/#{collection}/#{record}/manifest.json"
+    elsif server && collection
+      "https://#{server}.contentdm.oclc.org/iiif/info/#{collection}/manifest.json"
+    elsif server
+      "https://#{server}.contentdm.oclc.org/iiif/info/manifest.json"
     else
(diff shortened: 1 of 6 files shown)


benwbrum:
After we import a IIIF manifest from CONTENTdm, we fire off a rake task to query the CONTENTdm REST API and import page-level metadata like OCR text and configuration-specific metadata fields which are not exposed as IIIF.  

Before we merge this in, can we do a load test to see if importing a collection with 100 manifests kicks off 100 simultaneous concurrent rake processes?  Because that could clobber both the CONTENTdm servers and our own server.

jackweinbender:
Ok. I've looked into this more closely.

As far as I can discern, nothing runs asynchronously or in parallel when we import from IIIF or Cdm. The main "collection import" rake task deals with iterating through work-level manifests in the ruby code and not by firing off a rake task for each work. On our end, that means collection importing is a single, long-running process.

We do have a rake task for single-work imports, but, we only call it once and never in a loop.

As for the Cdm methods (those that go out and get extra metadata and ocr text), these are called within these respective rake tasks on particular works. Since all of our work-level iteration happens synchronously, this shouldn't be a problem.

All-in-all, I think we're in good shape re: not clobbering either our won servers, nor Cdm. We will still want to "nice"-up these tasks, I suspect, however (see #1456).

PR #1421, opened 2019-08-07
Join collection on signup

### LINKED ISSUE #1370: referral link to include owner/collection/work
USHMM asked for a way to link folks directly to a work/collection and to make sure they get there after signing up on FtP.

My idea was a new deed type "joined" that would show the material on their dashboard and show them in the collaborator list.

Ben's idea was a "next URL" parameter passed in to the registration link.

Third idea:  Save from the session what landing page they came from. (we have this with Ahoy.)  Perhaps add this to the "joined" deed.

jackweinbender:
Closes #1370 

This PR adds the `Collection Joined` Deed Type and automatically has new users "join" a collection based on where they entered the site.

Although the `Ahoy::Visit` does include a `landing_page` field, I couldn't figure out a good way to parse the path to determine whether it was, in fact, a path that included a `Collection`. As a way around this, I've opted to use the oldest `Ahoy::Event` from the `current_visit`. Ostensibly, this gives the same data, since the first `Event` of a `Visit` includes, by definition, the landing page. The advantage of using the `Event` is that it provides`Controller#method` as well as the URL parameters. So, we can check to see if this `Event` includes a `collection_id`, and if so, we can pass it to the `join_collection` method on the `User`, which makes the `Deed`.

The tests are basic, but cover the core functionality.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -26,10 +26,6 @@ class RegistrationsController < Devise::RegistrationsController
     else
       @user = build_resource(sign_up_params)
     end
-    
-    # Record the `joined` deed based on Ahoy Visit
-    join_collection = joined_from_collection(current_visit.id)
-    @user.join_collection(join_collection) unless join_collection.nil?
 
     #this is the default Devise code
     yield resource if block_given?
@@ -48,7 +44,11 @@ class RegistrationsController < Devise::RegistrationsController
       if session[:guest_user_id]
         session[:guest_user_id] = nil
       end
-
+      
+      # Record the `joined` deed based on Ahoy Visit
+      join_collection = joined_from_collection(current_visit.id)
+      @user.join_collection(join_collection) unless join_collection.nil?
+    
     else
       clean_up_passwords resource
       @validatable = devise_mapping.validatable?
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,5 +1,5 @@
 class RegistrationsController < Devise::RegistrationsController
-  
+
   def new
     super
   end
@@ -27,6 +27,10 @@ class RegistrationsController < Devise::RegistrationsController
       @user = build_resource(sign_up_params)
     end
     
+    # Record the `joined` deed based on Ahoy Visit
+    join_collection = joined_from_collection(current_visit.id)
+    @user.join_collection(join_collection) unless join_collection.nil?
+
     #this is the default Devise code
     yield resource if block_given?
       
@@ -75,4 +79,11 @@ class RegistrationsController < Devise::RegistrationsController
     true
   end
 
+  private
+
+  def joined_from_collection(visit_id)
+    first_event = Ahoy::Event.where(visit_id: visit_id).first
+    collection = first_event.properties["collection_id"] || nil
+    return collection
+  end
 end
--- a/app/models/deed_type.rb
+++ b/app/models/deed_type.rb
@@ -21,6 +21,7 @@ class DeedType
   WORK_ADDED = 'work_add'
   COLLECTION_ACTIVE = 'coll_act'
   COLLECTION_INACTIVE = 'coll_inact'
+  COLLECTION_JOINED = 'coll_join'
 
   # The TYPES hash houses all of the deed types and makes is easier to access
   # groups of deed types and also their human-readable names. Any new deed type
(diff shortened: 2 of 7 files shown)


saracarl:
So with this code I can't create a new user at all.  I've tried it from a collection landing page, an organization/owner landing page, and the app landing page.  Every time I get redirected to the collections list with a:

Bad record ID exception for params={"utf8"=>"✓", "authenticity_token"=>"xxx", "user"=>{"login"=>"landingpage", "email"=>"saracarl+landingpage@gmail.com", "password"=>"xxx", "password_confirmation"=>"xxx", "display_name"=>"landingpage", "activity_email"=>"1"}, "button"=>"", "action"=>"create", "controller"=>"registrations"}
Redirected to http://localhost:3000/dashboard

saracarl:
Any chance this depends on #1417 ?

jackweinbender:
I'm having a difficult time consistently reproducing this error. I *have* managed to reproduce it, but it has (seemingly) inexplicably resolved itself, and now I can't get it back again.

saracarl:
So now I can create my user successfully (yay) but I'm not seeing a deed in the activity stream or on my user dashboard (it doesn't show any deeds, and assumes I haven't done anything).  This happens if I start at a work or a collection.

jackweinbender:
Hmm. So, just to be sure we're on the same page on the functionality, when you say the deed isn't showing up if you "start" on a work or collection, are you starting a new `visit` or just clicking on `sign up` from a collection/work page? It will only log the `joined` deed if the landing page of your `visit` is a collection/work. So, if you start on the homepage, then navigate to a collection, it wont work because the landing page, technically, will have been the homepage.

I've been testing this by opening a "private" tab and going directly to a collection/work page then clicking `sign up` and it logs the deed—this is also how the feature test is written.

saracarl:
Ah, incognito is a great idea.  I was killing the server between attempts, but I don't think it worked to reset the session.

So that worked great for a collection, which is what this PR is for.  When I happened to be on a work page, though, I get an error after creating my user:

No route matches {:action=>"profile", :controller=>"user", :format=>nil, :user_id=>#<User id: nil, login: "aug193", display_name: "aug 193", print_name: nil, email: "saracarl+aug193@gmail.com", owner: false, admin: false, encrypted_password: "28b7695073799fc41cd487de556aea35668bc046", password_salt: "rxHFELXBguEsZeZgzzc6", created_at: nil, updated_at: nil, remember_token: nil, remember_token_expires_at: nil, location: nil, website: nil, about: nil, reset_password_token: nil, reset_password_sent_at: nil, remember_created_at: nil, sign_in_count: 0, current_sign_in_at: nil, last_sign_in_at: nil, current_sign_in_ip: nil, last_sign_in_ip: nil, account_type: nil, paid_date: nil, guest: nil, slug: nil, deleted: false, provider: nil, uid: nil, start_date: nil, orcid: nil, dictation_language: "en-US", activity_email: true>} missing required keys: [:user_id]

Seems to be a chicken and egg problem with line 7 of app/views/deed/_deed.html.slim

jackweinbender:
I've added several tests coming from collections, works, and pages that all behave as desired in the tests. I've gotten some weird errors in my manual testing, though. The apparent inconsistencies / non-reproducability here are making me uneasy, but I'm pretty sure most of the problems have to do with the ways that we have to reset cookies/use incognito to manually test. I think one of the new problems may have actually *caused by* incognito mode (no visit was being created; I restarted the server and that may have fixed it?). 

Whatever the case, I added some safety checks to make sure the signup form doesn't blow-up if there is no `current_visit` set.

saracarl:
Ok, I found another thing that needs to be fixed for this:

Screenshot from 2019-08-21 09-26-36

The deed list on the left side looks great, but the user dashboard just has the username, not a "username joined" or "username joined collection".  It looks off.

jackweinbender:
After some digging, I found that the problem was that this specific controller intentionally deleted the "collection" from the deed in order to prevent the collection name from displaying. The reasoning, I think, is that these deeds are displayed in the context of a particular collection, making the repetition of the collection ("X did Y *in Z collection*") redundant.

So, in those cases, I thought "joined this collection" seemed suitable:

image

### Bonus
I found one other bug in my implementation. I had the deed creation logic outside the `if/else` bock that tested if the user creation had succeeded, so, anytime a validation would fail, an attempt to create the deed would still be made---without a user being successfully made---and and things would blow up. So, I moved the deed logic into the `if save` block and now the validations work as expected and things don't explode.

saracarl:
Finally done!

PR #1429, opened 2019-08-12
Restrict document set routes.

### LINKED ISSUE #1425: document set edit screen visible without being logged in
If you go to a document set listing for a collection, without being logged in, you can see the "edit and delete" screen for the document sets, which is a problem.  (I haven't tried to see if you can actually do the edit, but you can see the view for sure):

https://fromthepage.com/document_sets?collection_id= (shortname)

A customer reported this -- we need to fix it soon.

jackweinbender:
Closes #1425

The DocSet controller seemed to not have any authentication logic whatsoever. I borrowed the auth logic from the `collection_controller` and restricted all but the index method.

### DIFF
--- a/app/controllers/document_sets_controller.rb
+++ b/app/controllers/document_sets_controller.rb
@@ -1,10 +1,21 @@
 class DocumentSetsController < ApplicationController
+  before_filter :authorized?, :except => [:index]
   before_action :set_document_set, only: [:show, :edit, :update, :destroy]
 
   respond_to :html
 
   # no layout if xhr request
   layout Proc.new { |controller| controller.request.xhr? ? false : nil }, :only => [:new, :create, :edit, :update]
+
+  def authorized?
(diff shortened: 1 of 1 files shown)


saracarl:
So the actual page that the customer is worried about isn't fixed here  -- it's
http://localhost:3000/document_sets?collection_id=doc-set-slug
Can we add that to this?

benwbrum:
@jackweinbender asked:

>  Looking at that URL, is it the `index` route? Thats the only one that doesn’t require auth. We must be doing something special with that route/view combo.

It is that one.  We have no feature that involves listing _all_ document sets across the site, so the `index` action listing document sets lists them for a single collection, within the context of that collection.  (And it usually lists them just for administrators, I think.)

saracarl:
Merging this, but we need to address the view part of it in a separate PR.

PR #1432, opened 2019-08-12
Better Cdm import errors

### LINKED ISSUE #1422: Better error message on bad CONTENTdm import
Currently, if a user tries to import a compound object from CONTENTdm that doesn't work (see examples in #1409), they get an unfriendly "Please enter a valid IIIF manifest URI" message, which totally confuses them.  In a lot of cases, they get the error message because CONTENTdm doesn't supply a IIIF manifest for the item; perhaps because it's a single item or because it's in a content type CONTENTdm hasn't implemented IIIF for.

We need to do a better job of presenting friendly error messages.  In particular, we should say "No IIIF manifest exists for CONTETNdm item #{original_cdm_uri}." if that's the case.

jackweinbender:
Closes #1422

Gives a better error message when the Cdm URL is well-formed but lacks a good IIIF manifest.

Because this error happens at the IIIF import stage (and not in the Cdm module) I had to pass the `source` and `source_uri` prams to the import route so that we knew that it was a Cdm import and not straight from a IIIF URL. Presumably these params could also be used for other import sources as well.

Since this built off of my previous Cdm work for collection importing (especially re: the test suite), I've added #1419 as a dependency for this PR.

Also: minor fix from a warning in one of the tests of the cdm translator


lib/contentdm_translator.rb
```diff
@@ -181,13 +181,18 @@ def self.cdm_record(at_id)
...
-    matches = url.match(/https?:\/\/(cdm\d+).*collection\/(\w+)\/id\/(\d+)/)
-    if matches
-      server = matches[1]
-      collection = matches[2]
-      record = matches[3]
-      
-      "https://#{server}.contentdm.oclc.org/digital/iiif-info/#{collection}/#{record}/manifest.json"
+    matches = url.match(/https?:\/\/(cdm\d+)(?:.*collection\/(\w+)(?:\/id\/(\d+))?)?/)
+    
+    server = matches[1]
```
benwbrum: Will this error if `matches` is false, say if the user puts in a non-URL or bad URL?

Issue #1437, opened 2019-08-13
Bizarre behavior of edit box+save button while transcribing long texts.


benwbrum:
In the Jeremiah White Graves diaries, our addition of a save button at the bottom of the screen may have caused some problems, or surfaced problems which had already existed.  Note that this document is long, and that it is imported from the Internet Archive.

See, for example this page: https://fromthepage.com/benwbrum/jeremiah-white-graves-diaries/jeremiah-white-graves-diary-volume-2-book-03/transcribe/428357#page/n0/mode/1up

* Scroll to the bottom of the transcription entry and observe the text stretching beyond the transcription form.
* Edit the text and attempt to press the bottom _Save_ button.  The button does not fire as it is somehow within the textarea, which has selection and prevents the button from receiving focus.

benwbrum:
See location of bottom save button, transcription text, and textarea border on this screenshot from [the Graves diary](https://fromthepage.com/benwbrum/jeremiah-white-graves-diaries/jeremiah-white-graves-diary-volume-2-book-03/transcribe/428360#page/n0/mode/1up)

![fromthepage_bad_save_button](https://user-images.githubusercontent.com/199961/62908688-e50d8000-bd3e-11e9-82cd-ce151dee2af4.png)

benwbrum:
When fixing this, please be sure to test:
* Over/under and side-by-side views of full-text transcripts
* Field-based transcription (as for Maryland State Archives and Indiana State Archives).

diegoviola:
@benwbrum can you please test if the proposed PR fixes the problem?

saracarl:
The scrollbar on the right side of the text entry box doe not extend all the way to the bottom:

![Screenshot from 2019-08-19 09-29-53](https://user-images.githubusercontent.com/708566/63273752-25e82600-c264-11e9-9206-f992e52d0365.png)

benwbrum:
This has been fixed for the side-by-side view but not for the over/under view.

benwbrum:
This is broken again on https://fromthepage.com/benwbrum/jeremiah-white-graves-diaries/diary-of-jeremiah-white-graves-volume-2-book-17/transcribe/6001#page/n0/mode/1up

It may have something to do with switching from "image on the left" and "image on top"

PR #1455, opened 2019-08-14
transcribe: fix save button bug

diegoviola:
margin-top causes the textarea to overlap with the "Save Changes" button
(when transcribing a lengthy document), making the button unclickable.

Removing the CSS property fixes the issue.

Fixes: #1437


saracarl:
The scrollbar on the right side of the text entry box doe not extend all the way to the bottom:

Screenshot from 2019-08-19 09-29-53

diegoviola:
@saracarl the latest commit should fix the issue, can you please test it?

Issue #1445, opened 2019-08-13
Suppress linebreak spans in XHTML export


benwbrum:
In some recent version the XHTML exports changed from 
```html
one line<br/>
another line
```
to
```html
one line<span><br/>
</span>another line
```
These additional `span` tags are breaking imports into the Biodiversity Heritage Library.  This might have happened as part of work on the `xml_source_processor` done several months ago (so they might be saved in `pages.xml_text`) or it might have happened due to changes in the export code.  They need to be removed during export.

We should also check the tei-xml export to see if the `span` tags are appearing there as well.

PR #1458, opened 2019-08-20
export: supress spans in XHTML export

diegoviola:
Fixes: #1445

### DIFF
--- a/app/helpers/abstract_xml_helper.rb
+++ b/app/helpers/abstract_xml_helper.rb
@@ -10,6 +10,11 @@ module AbstractXmlHelper
     return "" if xml_text.blank?
     xml_text.gsub!(/\n/, "")
     xml_text.gsub!('ISO-8859-15', 'UTF-8')
+
+    if preserve_lb
+      xml_text.gsub!("<lb break='no'/> ", "-<br />")
+    end
+
     doc = REXML::Document.new(xml_text)
     #unless subject linking is disabled, do this
     unless @collection.subjects_disabled
@@ -38,6 +43,7 @@ module AbstractXmlHelper
     end
     # get rid of line breaks within other html mark-up
     doc.elements.delete_all("//table//lb")
+
     # convert line breaks to br or nothing, depending
     doc.elements.each("//lb") do |e|
       lb = REXML::Element.new('span')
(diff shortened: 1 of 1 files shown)


diegoviola:
Unfortunately this creates a regression with #1090 which I still couldn't manage to fix. Suggestions?

benwbrum:
_If_ this addresses #1090, then I think we could implement it like this:
```
doc.elements.each("//lb") do |e|
      if preserve_lb
        if e.attributes['break'] == "no"
          # ADD "-" preceding e somehow.  Maybe e.previous_sibling.add_text("-") would work?
        end
        e.replace_with(REXML::Element.new('br'))
      else
        lb = REXML::Element.new('span')
        lb.add_text("")
        lb.add_attribute('class', 'line-break')

        unless e.attributes['break']=="no"
          lb.add_text(' ')
        end
         e.replace_with(lb)
       end      
    end
```

app/helpers/abstract_xml_helper.rb
```diff
@@ -38,23 +43,30 @@ def xml_to_html(xml_text, preserve_lb=true, flatten_links=false)
...
+          if e.attributes['break'] == "no"
+            lb.add_text('')
+          end
+        else
+          if e.attributes['break'] == "no"
+            lb.add_text('-')
+          end
         end
+
+        lb.add_element(REXML::Element.new('br'))
```
benwbrum: If `preserve_lb==false`, we need to stop adding `br` tags, so this line needs to be deleted.

benwbrum:
This gets rid of the `span` tags, but needs a single change to avoid inserting `br` when `preserve_lb==false`.

benwbrum:
We need to do the following:
* Make sure the XHTML export uses `preserve_lb=true`
* Add a single whitespace character when `preserve_lb=false` and the break is hard.

See the following screenshots:
## Original source
See source selected -- note the "hard break" text.
Screenshot from 2019-09-05 16-05-55

## Preserve_LB=*true*
Overview view for a page -- this is perfect.
Screenshot from 2019-09-05 16-06-25

## Preserve_LB=*false*
Multi-page/Read-work view.  See the lack of white space after the hard break.  We need to add some kind of space (probably just a `' '` space character, not an `&nbsp;`) here so that the appearance is `a hard` instead of `ahard`.

Screenshot from 2019-09-05 16-06-48

benwbrum:
This works in all cases now.

PR #1459, opened 2019-08-20
intercom integration

saracarl:


app/views/shared/_intercom.erb
```diff
@@ -0,0 +1,4 @@
+<script type="text/javascript">
```
benwbrum: I'm not sure it matters much in a partial, but this should probably be named `_intercom.html.erb`.

app/views/shared/_intercom.erb
```diff
@@ -0,0 +1,4 @@
...
+	window.intercomSettings = { app_id: "<%=INTERCOM_APP_ID%>" };
```
benwbrum: Where is INTERCOM_APP_ID defined?  Should that location be updated with a placeholder?

Issue #1435, opened 2019-08-12
Owner profile does not show private collections/sets to collaborators


benwbrum:
For customers that are running private collections or document sets, the owner profile page is not usable, since it only shows the organization's projects that are public.

We should mimic some of the logic in the collection_list page to show users the projects they can work on, rather than only public projects.

benwbrum:
This is related to #1196 -- the only difference being that in #1196 the user is added as a collaborator on the document set, while in this bug the user is added as a collaborator on the document set's parent collection.

jackweinbender:
closed by #1462

benwbrum:
This is not working for ADAH's use case because of the `is_public?` check in the view.

PR #1466, opened 2019-08-22
Proper handling of docset auth

jackweinbender:
I originally misunderstood how this route worked. The page under consideration should only be able to be edited by people with *collection*-level owner permissions. This is the right way to fix that page's permissions errors.

### DIFF
--- a/app/controllers/document_sets_controller.rb
+++ b/app/controllers/document_sets_controller.rb
@@ -1,5 +1,5 @@
 class DocumentSetsController < ApplicationController
-  before_filter :authorized?, :except => [:index]
+  before_filter :authorized?
   before_action :set_document_set, only: [:show, :edit, :update, :destroy]
 
   respond_to :html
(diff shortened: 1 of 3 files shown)


jackweinbender:
Although this PR does not depend on #1435, because DocSets are "shown" to users based on the `.show_to?` method, this will yield the proper restrictions only after #1435 is merged.

saracarl:
I'm testing this with development merged in, so we have #1462.

jackweinbender:
Ok. I think I figured it out. I had structured the `authorize` function as two consecutive `if` statements rather than an `if..elsif...else` and so the request was trying to call a method on a user that wasn't there. 

Since all of those failure conditions result in the same redirect, I moved them all in to one line of boolean logic.

This change revealed another bug, which was that we weren't passing the `collection` id to the `docset#create` method. It wasn't breaking before because we weren't actually filtering unauthorized requests. Now the `collection_id` is passed to the `#create` route as a query param and the operation is only successful if the user is authorized to act like an owner for the collection.

Issue #1063, opened 2018-05-09
Warn users if another user is editing this page.


benwbrum:
We're running into problems at Transcribe-a-thons and other high-volume projects in which transcribers have their page save fail because of stale database records.  We should track down what's going on and come up with a strategy for dealing with this.  (I'd thought that we implemented optimistic locking 11-13 years ago -- is that not working?  Is that the problem?)

Possible reproduction scenario:
- User A starts to edit a page.
- User B starts to edit the same page.
- User A saves the page.
- User B attempts to save the page and sees an error.

Once we come up with a reproduction scenario, we'll need to formulate how to present the error to the end user.  If the scenario I suggested is correct, should we attempt to merge B's changes into A's transcript?

Example error email:
====Page Save Failed on FromThePage====

=====Failure message=====
```
TRANSCRIPTION 2018-05-09 16:15:37 +0000 TRANSCRIPTION User ID: 49611 Email: staff02.archives@archives.alabama.gov	Display Name: ADAH student TRANSCRIPTION Collection ID: 69 Title:WWI service records for Jefferson County, Alabama Owner Email: digital.adah@gmail.com TRANSCRIPTION Work ID: 3974 Title: World War I military service cards for Jefferson County, Alabama (Scaife through Seymore). TRANSCRIPTION Page ID: 115435 Position: 135 Title:Marion Seals TRANSCRIPTION Source Text: BEGIN_SOURCE_TEXT END_SOURCE_TEXT
```

=====Error/Exception message=====
`Attempted to update a stale object: Page`

benwbrum:
Make sure works with field based projects like Alabama's.

benwbrum:
Maybe the right thing to do here is go ahead and over-write the previous version, and send a warning letter to the project owner with a direct link to the version comparison view.

benwbrum:
Looks like this can turn off optimistic locking:  ActiveRecord::Base.lock_optimistically = false 

However, we only want to disable it on pages.  However, if we disable optimistic locking we'll lose the exception that we could use to send error messages.

Please duplicate and make sure our theory is correct & we can reproduce.  Can we create a spec test???

Let's think through this some more.  Docs to review:

http://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html
[...]

PR #1468, opened 2019-08-23
transcribe: alert the user when a page is being transcribed

diegoviola:


benwbrum:
When I run this locally, I do not see the warning message.  Instead, I see this error in the Javascript console:
```
:3000/benwbrum/1/17/still_editing/6086:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
```

benwbrum:
The error message in the javascript console appears to be unrelated.  I can no longer duplicate it.

app/controllers/transcribe_controller.rb
```diff
@@ -29,6 +29,7 @@ def display_page
     @auto_fullscreen = cookies[:auto_fullscreen] || 'no';
     @layout_mode = cookies[:transcribe_layout_mode] || @collection.default_orientation
     session[:col_id] = @collection.slug
+    flash[:alert] = "This page is being edited by another user!" if @page.edit_started_at > Time.now + 1.minute unless @page.edit_started_at.nil?
```
benwbrum: This needs to be changed from `Time.now + 1.minute` to `Time.now - 1.minute`.

diegoviola:
@benwbrum I pushed a change that introduces the `edit_started_by_user` that you suggested in #1063 

Now if I click on the same transcribe link from another browser I get the warning as it's supposed to happen (with a different user).

Thanks for your help with this.

benwbrum:
I've tested this and found an odd edge case. 

* User A starts editing page 1
* User B clicks the transcribe tab on page 2
  * User B sees the warning message
  * User B's browser calls the `still_editing` route (unknown to them), updating the edit time and user_id
* User B clicks to page 2
* User B clicks back to page 1
* User B does _not_ see the warning message, even though user A is still editing the page

I _think_ this can be solved by stopping the Javascript from being called if the warning message was displayed.  I'm not sure whether it's better to do this client-side (by looking at the `flash`), or server-side (by looking at `@page.edit_started_by_user_id` and the time window).  It might be better to look in the flash since that keeps us from duplicating the time window logic.

Issue #1485, opened 2019-08-29
Rack Mini-Profiler for Admins


jackweinbender:
Display RMP info for Admins (enable/diable-able) on for production environments.

jackweinbender:
Closed by #1486

Issue #1430, opened 2019-08-12
Add DB instrumentation


benwbrum:
Oink has been invaluable in identifying large memory issues in production.  We'd like to add other libraries to check database performance and find actions that are firing off large numbers of queries.  

So far the recommendations have been
* `n+1`
* `rack-mini-profiler`
* `bullet`

How do these compare?

benwbrum:
We need to explore each of these three gems/approaches and figure out the pros/cons of each.  

* Currently the `findaproject` action is loading slowly.  How would we figure out why (or quantify it against different items) using each Gem?
* What kind of dependencies are introduced for each approach?  How much configuration will be needed, and will that be an undue burden for open-source installations?
* Does each run in development, production, or both?  Do they require a human to click on pages and watch what's happening, or--like Oink--can they gather data for later analysis, showing frequency of use in addition to cost of use for each page?

benwbrum:
## Rack Mini-Profiler

I'd like us to add RMP in a way that lets us see it by default in development environments, and lets us turn it on (or off) in production so that we can see RMP reports on our server.  

The constraints about the production use case are that 1) admin accounts should be able to see RMP reports (and perhaps should by default), but that 2) we often use admin accounts to demo the tool over screen-share, where we don't want any instrumentation visible in the UI.  My temptation is to add a new drop-down option to the Admin menu for "Enable Performance Tracking" which sets a session variable conditionalizing display of the RMP (and an inverse "Disable Performance Tracking").  I'm not sure whether admin users should see the RMP display by default or not -- while we don't want to interrupt a demo with RMP displays, it'd also be valuable--if unpleasant and distracting--to display RMP by default so that rarely-used actions are profiled.

## Bullet
This seems clearly useful and actionable.  Is there any reason not to deploy this immediately?

benwbrum:
@saracarl votes that we not display RMP by default.  We should also make the action that enables/disables performance tracking display public, so that we can monitor performance on production as site visitors (not logged in users) by hitting the URL.

jackweinbender:
Great. I think that's a good plan. My only concern about the last suggestion is that part of the profiling provided by RMP is to display the SQL query itself. So, we'd potentially be exposing some DB info (table names, ids, maybe some "where" info, etc.) to the public. I don't think it exposes full objects anywhere, but this seems like a potentially serious security issue. I don't know, do people still worry about SQL injection?

benwbrum:
[...]

PR #1486, opened 2019-08-29
Added Bullet and RMP

jackweinbender:
Currently this just adds the Bullet and Rack-Mini-Profiler gems to the gemfile with no configuration.

The RMP gem "just works" in the development environment, but the Bullet gem requires some config in the `development.rb` initializer. We `.gitignore` that initializer, however, so we'll need to either need to check it back in or just configure it for ourselves (I'm inclined toward checking it back in, personally).

I've opened  #1485 to address configuring RMP for use in production environments.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ gem 'omniauth-google-oauth2'
 
 gem 'rack-reverse-proxy', :require => 'rack/reverse_proxy'
 
+
 group :assets do
   gem 'therubyracer'
   gem 'uglifier'
@@ -67,9 +68,7 @@ group :development, :test do
   gem 'binding_of_caller' # goes with better_errors
   # Supporting gem for RailsPanel
   # https://github.com/dejan/rails_panel
-  gem 'meta_request'
   gem 'bullet'
-  gem 'rack-mini-profiler'
 end
 
 # Use SASS for stylesheets
@@ -96,3 +95,10 @@ gem 'iso-639'
 
 # Quiet asset lines in log files
 gem 'quiet_assets', '~> 1.1.0', group: :development
+
+# Profiling for use in prod
+gem 'flamegraph'
+gem 'memory_profiler'
+gem 'meta_request'
+gem 'rack-mini-profiler'
+gem 'stackprof'
--- a/Gemfile
+++ b/Gemfile
@@ -68,6 +68,8 @@ group :development, :test do
   # Supporting gem for RailsPanel
   # https://github.com/dejan/rails_panel
   gem 'meta_request'
+  gem 'bullet'
+  gem 'rack-mini-profiler'
 end
 
 # Use SASS for stylesheets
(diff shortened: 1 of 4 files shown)


benwbrum:
What are the configuration options that need to be added to `development.rb`?  I'd like to figure out where the conflicts are, and why we might have pulled it out of source control.

jackweinbender:
Bullet has lots of config options, but most of them have to do with third-party integrations. The only essential option is to enable bullet, then you need to picka method for logging. I don't like messing with log files if I can help it, so, I like to also include the browser console. My config looks like this (appended just before the last `end` in `development.rb`:

```
config.after_initialize do
  Bullet.enable = true
  Bullet.bullet_logger = true
  Bullet.console = true
end
```

But here's the whole list of options from the docs:
```
config.after_initialize do
  Bullet.enable = true
  Bullet.sentry = true
  Bullet.alert = true
  Bullet.bullet_logger = true
  Bullet.console = true
  Bullet.growl = true
  Bullet.xmpp = { :account  => 'bullets_account@jabber.org',
                  :password => 'bullets_password_for_jabber',
                  :receiver => 'your_account@jabber.org',
                  :show_online_status => true }
  Bullet.rails_logger = true
  Bullet.honeybadger = true
  Bullet.bugsnag = true
  Bullet.airbrake = true
  Bullet.rollbar = true
  Bullet.add_footer = true
  Bullet.stacktrace_includes = [ 'your_gem', 'your_middleware' ]
  Bullet.stacktrace_excludes = [ 'their_gem', 'their_middleware', ['my_file.rb', 'my_method'], ['my_file.rb', 16..20] ]
  Bullet.slack = { webhook_url: 'http://some.slack.url', channel: '#default', username: 'notifier' }
end
```
As for the `development.rb` initializer, I don't change anything locally on my machine (this would be the first exception to that), so I'm perfectly fine checking it in, personally. That said, theoretically, we all could have different development environments, and so, usually, (I think) you don't check them in.

A better approach, however, might be to check-in a `development.rb.example` file, then still keep the real `development.rb` out of VC. I run into this problem with my `database.rb` file because I have to change the socket path to something different for my docker environment. That means every time I `git stash` or `git reset --hard` my local changes get wiped out. It's a one-liner, so it's no big deal, but for something more substantial, I would need some other solution.

jackweinbender:
Or. I wonder if we could put bullet into its own initializer and only call it if we're in the `dev` or `test` environment? [Something like this, maybe?](https://stackoverflow.com/questions/4820987/rails-per-environment-initializers#answer-4821811)

jackweinbender:
I couldn't figure out a way to test this, because it is dependent on being in `production` env. When I try to flip my env variables, everything explodes because I don't have any SSL stuff configured for my local box.

jackweinbender:
I was able to get this working "in production."

Guest users:
image

Signed in non-admins:
image

Defaulted to "hidden" for admins:
image

Enable-able by admins by adding `?pp=enabled` or by pressing `Alt+p`
image

Confirmed that `Alt+p` and/or query params  do not work for other users, guests or otherwise.

benwbrum:
Addresses #1430

Issue #1428, opened 2019-08-11
Performance of `has_untranscribed_pages?`


benwbrum:
The collection list page iterates through collections and calls `has_untranscribed_works?` on each of them in order to determine whether to show the _Start Transcribing_ link.  This formerly called `detect` on `Work#has_untranscribed_pages?`, which was slow in part because the latter method iterated over records.  Now it's refactored to use work statistics, but still executes a query for each collection and document set.

We should fix this by:
* Refactoring `Work#has_untranscribed_pages?` so that it doesn't iterate over pages.
* Using collection-level or document-set-level statistics to make this determination.  This may require extensive work to roll up the work statistics on the collection or document set level if such code does not already exist.

jackweinbender:
I've been toying with a few ideas for this, but I think out best bet is going to be to add a field to either works or collections (or both) with a `has_untranscribed` flag that is part of the stats recalculation. 

I originally tried making a relationship for `collection` that returned `untranscribed_pages`. The advantage here is that we are then able to eager-load the relationship, which prevents the N+1 for collections. This did, in fact, work, however, the main query becomes dangerously large because we, in effect, are eager loading every single untranscribed page in every public collection. 

We could probably hack something together to with grouping/counts, but at the end of the day, I we're just using this for conditionally displaying a button. It seems to me that an added column to the `works_statistics` (or, again, perhaps also the `collection_statistics`) table would solve this querying quagmire by adding a single query to the collection and work `recalculate` methods.

Having just done work in the `works_statistics` area for #1490 I know this will be fairly trivial to add to the `work_statistic#recalculate` method. We'd then have the option of either doing an eager-loaded query for works with untranscribed pages, or saving this info in the collection statistics model when it recalculates.

One final option is if we can get all of of "find a page to translate" logic into a single query, returning a single page, we may be able to just eager-load that. This approach has the added benefit of making out "start_transcribing" route nearly obsolete.

But, again, I think the simplest solution is just to add the field/s and, effective. cache this info.

benwbrum:
I like the idea of adding the field, though here is another option that might kill two birds with one stone:

* Add `next_untranscribed_page_id` to `work_statistics`, populated with whatever logic we use for _find a page to transcribe_
* Add `def has_untranscribed_pages? return !next_untranscribed_page_id.nil? end

What do you think?

jackweinbender:
[...]

PR #1494, opened 2019-09-04
Improve dashboard#landing_page route perf

### LINKED ISSUE #1493: findaproject route performance

jackweinbender:
Closes #1493 

This addresses the slow `dashboard#landingpage` (where the "Find A Project" button takes you as a guest performance.

# Performance

## Initial Measurements
This route had two major problems:
  1. Querying for the relevant collections and docsets used a couple `.map` operations which fired off N+1 queries.
  2. Collection rendering was also exhibiting N+1
**Results**: ~240 queries per page load, total time to render ~2.5 seconds

## Improved Measurements
The Collection/Docset rendering was slow do to the fact that we were not only *not* eager loading, but, when references to a particular Collection's `owner` were used (e.g., in `link_to`), they were accessed through the inverse relationship that they were queried from. In other words, when we are looping through `@users as |user| , then looping through `user.collections as |collection|`, we would reference `collection.owner` rather than just `user` and this was firing off N+1 queries.

I was able to remove the `map` function from the Carousel logic and reduced this to a single query.

I added a couple convenience scopes and relationships in order to allow us to eager load a small sample of collections and document sets for each user. This was a [technique I got from here](https://www.justinweiss.com/articles/how-to-preload-rails-scopes/) and it's pretty slick.

There **is** exactly ONE N+1 case here. The way we find a link to an untranscribed page in a collection currently requires us to make an extra query to the `works` table for every project in the carousel. However, there can not be more than 8 items in the carousel, so, N will never be greater than 8. It's easily solvable for the Collection's case if we were to store the `collection_id` on the page object.  For obvious reasons, however, the case for Docsets would not be so clean. There may be a way to do something similar to what I did with `random_collections` in the link (above), but I couldn't figure out a way to make it work `:through` a third relationship.

Anyway, I thought I could live with the extra 8 queries; especially since I had removed nearly 200.
 
**Addendum**
In order to prevent further N+1s, we had to add `counter_caches` to `Collection`s and `DocumentSet`s; this meant that I had to make some small changes to the database (added columns) and to explicitly define the `has_and_belongs_to_many` relationship between `DocumentSet`s and `Work`s. That's why there's the new `DocumentSetWork` model.

**Results:** ~20-40 queries (I haven't figured out what causes this variance), total time to render ~500ms

# Behavior

## Former Behaviour (Non-search)
Two instance variables (`@owners` and `@collections`) are passed to the view.
  1. `@owners` is a list of distinct "true" owners (i.e., those listed in `owner_user_id`, rather than the many-to-many `owners`) who have *public` collections.
  2. `@collections` is actually the collection of `Collection`s (notably, not `DocumentSet`s!) for the Carousel. It maps over the `@owners` and selects the first `Collection`s that matches the `.carousel` scope (below) which is "truly" owned by the `|owner|`. It then `.compacts` and `.sample`s that result to get 8 "random" `Collection`s
  3. The `.carousel` scope is defined thusly:
```
    .where(pct_completed: [nil, 1..90])
    .where.not(picture: nil)
    .where.not(intro_block: [nil, ''])
    .where(restricted: false)
    .reorder("RAND()")}
```
I read this as a randomly sorted list of **public** `Collection`s  that **have a picture** and a **non-empty `intro_block`** and are either **empty** (though, no 0 pct?) or **up-to 90% complete**.

**But, it gets worse...**
The view uses two helper methods in the `dashboard_helper` to generate the "Start Transcribing" links and also to call out to for which `Collection`s and `DocumentSet`s to display.
  1. We've already talked about fixing the `start_transcribing` links, that'll be taken care of in the PR for #1428 (So, we'll ignore this one for now).
  2. The `dashboard_helper#owner_projects` method is the culprit here. It accepts an `owner` and returns a randomly sampled set of 3 **public** `Collection`s and `DocumentSet`s that are "truly" owned by the `owner` that are less than 90% complete, that have an `intro_block`, and also aren't empty (i.e., they have `Work`s).

The view itself will only display an `owner` if the `owner` actually has some `Collection`s or `DocumentSet`s to display.

## Problems:
1. It seems like we should be including collections/docsets that the current user is authorized to work on (either as a scribe or as an owner) as well as public ones
2. The Carousel should probably be `Collection`s and `DocumentSet`s.

Otherwise, I think the functionality is fine.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -54,10 +54,11 @@ class User < ActiveRecord::Base
   has_many :random_document_sets, -> { unrestricted.sample }, 
     class_name: "DocumentSet", :foreign_key => "owner_user_id"
 
-  scope :owners,         -> { where(owner: true) }
-  scope :trial_owners,   -> { owners.where(account_type: 'Trial') }
-  scope :paid_owners,    -> { owners.where('paid_date > ?', Time.now) }
-  scope :expired_owners, -> { owners.where('paid_date <= ?', Time.now) }
+  scope :owners,           -> { where(owner: true) }
+  scope :trial_owners,     -> { owners.where(account_type: 'Trial') }
+  scope :non_trial_owners, -> { owners.where.not(account_type: [nil, 'Trial']) }
+  scope :paid_owners,      -> { owners.where('paid_date > ?', Time.now) }
+  scope :expired_owners,   -> { owners.where('paid_date <= ?', Time.now) }
 
   validates :display_name, presence: true
   validates :login, presence: true, uniqueness: { case_sensitive: false }, format: { with: /\A[a-zA-Z0-9_\.]*\z/, message: "Invalid characters in username"}, exclusion: { in: %w(transcribe translate work collection deed), message: "Username is invalid"}
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,6 +49,16 @@ class User < ActiveRecord::Base
   has_many :notes, -> { order 'created_at DESC' }
   has_many :deeds
 
+  has_many :random_collections,   -> { unrestricted.sample },
+    class_name: "Collection",  :foreign_key => "owner_user_id"
+  has_many :random_document_sets, -> { unrestricted.sample }, 
+    class_name: "DocumentSet", :foreign_key => "owner_user_id"
+
+  scope :owners,         -> { where(owner: true) }
+  scope :trial_owners,   -> { owners.where(account_type: 'Trial') }
+  scope :paid_owners,    -> { owners.where('paid_date > ?', Time.now) }
+  scope :expired_owners, -> { owners.where('paid_date <= ?', Time.now) }
+
   validates :display_name, presence: true
   validates :login, presence: true, uniqueness: { case_sensitive: false }, format: { with: /\A[a-zA-Z0-9_\.]*\z/, message: "Invalid characters in username"}, exclusion: { in: %w(transcribe translate work collection deed), message: "Username is invalid"}
   validates :website, allow_blank: true, format: { with: URI.regexp }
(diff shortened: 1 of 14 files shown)


saracarl:
So this vastly improves the loading of the find a project page by paring down the number of projects we display to a small fraction of the projects we previously displayed.  (I'm sure there are real performance improvements, but it's an apples to oranges comparison.)  We need to get rid of the "paid owner" filter.  I'm horrible at keeping that data up to date, and there are certain projects we like to have visible even though they are not paid owners.  (i.e. the BL)

app/models/user.rb
```diff
@@ -49,6 +49,16 @@ class User < ActiveRecord::Base
...
   has_many :deeds
 
+  has_many :random_collections,   -> { unrestricted.sample },
+    class_name: "Collection",  :foreign_key => "owner_user_id"
+  has_many :random_document_sets, -> { unrestricted.sample }, 
+    class_name: "DocumentSet", :foreign_key => "owner_user_id"
+
+  scope :owners,         -> { where(owner: true) }
+  scope :trial_owners,   -> { owners.where(account_type: 'Trial') }
+  scope :paid_owners,    -> { owners.where('paid_date > ?', Time.now) }
```
saracarl: We can't scope projects by this.

app/models/user.rb
```diff
@@ -49,6 +49,16 @@ class User < ActiveRecord::Base
...
+  scope :expired_owners, -> { owners.where('paid_date <= ?', Time.now) }
+
```
saracarl: You could add these to the paid owners if you'd like.  I'd be OK with that.  (keeping out trial owners, but not others.)

app/models/user.rb
jackweinbender: I've added `non_trial_owner` to better match the older functionality

app/models/user.rb
jackweinbender: Changed these to `non_trial_owners.where(...)` so that we don't mistakenly suggest trial owners (or `nil` folks) are unpaid.

jackweinbender:
I also noticed that I had forgotten to scope the carousel items appropriately.

I replaced the `paid_owners` scope and with the more appropriate `non_trial_owners` scope; so that was a pretty straight-forward change.

But fundamentally, I think what I have here is functionally the same as what we had before (albeit, I misunderstood the purpose of the non-trial stuff). And we still have the performance gain:
image

Issue #1334, opened 2019-04-25
searching collections should search owner names 


saracarl:
So it's getting hard to find particular collections, because we remember the owner, i.e. Chicago Botanic, but not the name of the collection "The Scientific Notebooks of German Orchidologist Friedrich Wilhelm Ludwig Kränzlin".  We should search owner names as well as collection names when you search from the "Find a Project" page.

benwbrum:
So ideally a search for "Virginia" would find the Library of Virginia, the University of Virginia, and the University of Virginia Law Library.  This may require changes to whatever view we are using to display search results.

benwbrum:
Concrete reproduction scenario using recent production data:

* Search for "vertebrate" on the Find A Project screen.
* Expected result: Find projects by the Museum of Vertebrate Zoology
* Actual result: Find no results

PR #1523, opened 2019-09-18
dashboard: include owner names when searching

diegoviola:
Fixes: #1334

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -149,7 +149,9 @@ class Collection < ActiveRecord::Base
   end
 
   def self.search(search)
-    where("title like ? OR slug LIKE ? OR owner_user_id in (select id from users where owner=1 and display_name like ?)", "%#{search}%", "%#{search}%", "%#{search}%")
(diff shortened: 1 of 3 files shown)


benwbrum:
This now works, but the addition to the landing page seem to duplicate existing UI in a way that's not attractive.

If I search for "Indiana", I (correctly) get all collections owned by the `Indianapolis Public Library` and the `Indiana Archives`.  However, those collections are displayed twice: once in a normal owner/collection layout, and once above these in a flat list of links.  We should remove the flat list of links.

See screenshot.
Screenshot from 2019-10-08 19-11-47

saracarl:
1)  The list of collections at the top of the screenshot needs to "go away"
2)  We need to test this with a search string that matches a collection title but NOT an owner name.
AND
We need to test this with a search string that matches an owner name but NOT a collection title.
3)  If just a collection (not an owner) is returned, it should look like the lower half of the screenshot above; show the owner name & matching collection (but not other collections).

saracarl:
If a search string matches the owner name, it should return the list of collections (like the above for Indianapolis public library if we had searched for "Indianapolis")

Issue #1447, opened 2019-08-13
better usability for "Start a Project"


saracarl:
On the "start a new project page" we have a lot of options for getting material into FromThePage.  I'd like to improve the usability of that page.
http://localhost:3000/dashboard/startproject?collection_id=431

1)  reorder the list of options:
upload pdf or zip
import from a iiif repository
import from ContentDM
import from internet archive
import from Omeka
create an empty work

2)  If we could collapse each section with a pointed triangle widget to open it, then all of the options would be visible initially.  

3)  Change the word "Omeka" to "Omeka Classic"

4)  Get rid of "Search for IIIF Content" and "Explore IIIF Collections"

diegoviola:
@saracarl do we want to change to "Omeka Classic" just in the title or everywhere that Omeka is mentioned?

PR #1524, opened 2019-09-18
dashboard: make titles clickable

diegoviola:
This improves on #1447 and #1484 by actually making titles clickable.

Fixes: #1447.

### DIFF
--- a/app/views/dashboard/_empty.html.slim
+++ b/app/views/dashboard/_empty.html.slim
@@ -1,5 +1,5 @@
     h3
-      a(href="#" class="title" onClick="return false;") Create Empty Work
+      a(href="#" class="title" onClick="return false;" id="create-empty-work") Create Empty Work
 
     section
       p.big Use this option to create an empty work. You can then upload individual page images into the work.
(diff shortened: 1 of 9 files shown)


saracarl:
Can we collapse the sections?  This doesn't add much when the sections are expanded to start with.

Also, can we change "An IIIF Manifest" to "A IIIF Manifest", please?

saracarl:
So if you load this page it starts out "expanded" (all sections open) and then immediately flips to closed/hidden.  Is there any way to start with them all closed?  This will make customers say "what just happened?"

saracarl:
If we're starting with the default (open) style, then have JS that closes the divs when the page loads, that can cause a flicker.  If we can start with the closed style by default, that would avoid the flicker.

diegoviola:
> If we're starting with the default (open) style, then have JS that closes the divs when the page loads, that can cause a flicker. If we can start with the closed style by default, that would avoid the flicker.

I tested it and noticed the flicker, thanks for letting me know.

Hiding the sections with CSS seems to avoid the flicker somehow, please let me know if the latest commits fixes the issue for you as well.

saracarl:
This looks good, but it's no longer passing tests because the tests can't see the div to upload in (I think).  Let's fix those.

PR #1540, opened 2019-09-29
Next untranscribed page after completed work

### LINKED ISSUE #1331: Don't lose transcribers after last page of work
LVA says:  "Users are often able to work through a document, but get stuck at the last page going "now what?""

I think this also brings up an import use case "How do we direct a transcriber to more works when they transcribe the last page of a work?"  That one's going to take more thinking, but I think it's an important idea to keep volunteers in "flow".

jackweinbender:
Closes #1331 

This PR helps to redirect users to new transcribed pages after they complete a work.

## New Functionality
When users are on the final page of a work and are in the `transcribe` view (i.e., not in the "reading" view), when they click the "Right Arrow" to advance to the next page in the work, they are taken to a new untranscribed page.

**Case 1**: The current work is incomplete
- The user is taken to the work's `next_untranscribed_page`

**Case 2**: The current work is complete:
- sub-case: the current work is a part of a Document Set
  - The user is taken to an untranscribed page within the current DocSet
- sub-case: the current work is a part of a Collection
  - The user is taken to an untranscribed page within the current Collection

In all instances of Case 2, the `current_user` is checked to ensure that they are able to transcribe the work of which the `next_untranscribed_page` is a part. If the case fails, another work in the same collection/docset is selected. In the case where the docset has no more candidate pages, the logic is bumped up to the collection level. If the collection level fails: 

**Case 3**: Failure of the above
- In cases where there are no valid untranscribed pages in the cascade, the user is brought to the `find_a_project` page.

In all cases, the user is given an indication of why they were redirected, and what level they were redirected to (Work/DocSet/Collection, etc.).

## Notes on Style
I struggled with this PR in large part because I was trying to avoid writing a giant nested if/else block. I partially succeeded. The if/else is still big and nasty, but it's not **too** difficult to follow. I was able to move the authentication logic out to the models, which helped prevent another permutation of the if/else blocks.

## Tests
I did test this functionality fairly extensively. I think I've gotten most of the corner cases, too.


app/controllers/transcribe_controller.rb
```diff
@@ -312,6 +312,37 @@ def still_editing
     render nothing: true
   end
 
+  def goto_next_untranscribed_page
+    
+    next_page_path = landing_page_path
+    flash[:notice] = "There are no more pages to transcribe in this collection! Check out some of our other projects."
```
saracarl: I wonder if we should redirect them to the project owner's page instead?

app/controllers/transcribe_controller.rb
```diff
@@ -312,6 +312,37 @@ def still_editing
...
+
+    if @work.next_untranscribed_page
+      flash[:notice] = "Here's another page in this work, if you'd like to continue."
+      next_page_path = collection_transcribe_page_path(@work.collection.owner, @work.collection, @work, @work.next_untranscribed_page)
+    elsif @collection.class == DocumentSet
+      docset = @collection
+      next_page = docset.find_next_untranscribed_page_for_user(current_user)
+      if !next_page.nil?
+        flash[:notice] = "There are no more pages to transcribe in this work. Here's another page in this document set if you'd like to continue."
+        next_page_path = collection_transcribe_page_path(docset.owner, docset, next_page.work, next_page)
```
saracarl: Transcribers aren't aware of document sets.  We should use the same messaging as a collection here.

app/controllers/transcribe_controller.rb
```diff
@@ -312,6 +312,37 @@ def still_editing
...
   end
 
+  def goto_next_untranscribed_page
+    
+    next_page_path = landing_page_path
+    flash[:notice] = "There are no more pages to transcribe in this collection! Check out some of our other projects."
+
+    if @work.next_untranscribed_page
+      flash[:notice] = "Here's another page in this work, if you'd like to continue."
+      next_page_path = collection_transcribe_page_path(@work.collection.owner, @work.collection, @work, @work.next_untranscribed_page)
```
saracarl: This one worries me a bit.  When you get to the end, it makes sense to be sent "elsewhere" but when you are inside of a work and the next page is already transcribed...  What is the right behavior?  We don't want to lose them, but it would be jarring to be plopped elsewhere in the work.
You might also want to read/review the next page.
At a minimum we should think about changing the message to make it clearer.  At a maximum, I think we should skip this scenario.

app/controllers/transcribe_controller.rb
```diff
@@ -312,6 +312,37 @@ def still_editing
...
+    elsif @collection.class == DocumentSet
+      docset = @collection
+      next_page = docset.find_next_untranscribed_page_for_user(current_user)
+      if !next_page.nil?
+        flash[:notice] = "There are no more pages to transcribe in this work. Here's another page in this document set if you'd like to continue."
+        next_page_path = collection_transcribe_page_path(docset.owner, docset, next_page.work, next_page)
+      else # Docset has no more Untranscribed works, bump up to collection level
+        next_page = docset.collection.find_next_untranscribed_page_for_user(current_user)
+        unless next_page.nil?
+          flash[:notice] = "There are no more pages to transcribe in this work. Here's another page in this collection if you'd like to continue."
```
saracarl: Let's get rid of "if you'd like to continue".  I think ""There are no more pages to transcribe in this work. Here's another page in this collection." is fine.

app/controllers/transcribe_controller.rb
```diff
@@ -312,6 +312,37 @@ def still_editing
...
+          next_page_path = collection_transcribe_page_path(docset.collection.owner, docset.collection, next_page.work, next_page)
+        end
+      end
+    else
+      next_page = @collection.find_next_untranscribed_page_for_user(current_user)
+      unless next_page.nil?
+        flash[:notice] = "There are no more pages to transcribe in this work. Here's another page in this collection if you'd like to continue."
```
saracarl: same comment: 
Let's get rid of "if you'd like to continue". I think ""There are no more pages to transcribe in this work. Here's another page in this collection." is fine.

app/controllers/transcribe_controller.rb
saracarl: So change this to:
"There are no more pages to transcribe in this work. Here's another page in this collection."

app/controllers/transcribe_controller.rb
benwbrum: I agree.  Sending them to the owner's page would keep them closer to the organization.

app/controllers/transcribe_controller.rb
saracarl: I think we want to send them to the next page even though it has been transcribed.

jackweinbender: This only happens on the last page. So, if a user is mid-work and the next page is already transcribed, they still just go to the next page as expected. The only difference here is that on the last page they get taken to a page in the same work that is incomplete.

app/controllers/transcribe_controller.rb
jackweinbender: That makes sense.

app/controllers/transcribe_controller.rb
jackweinbender: Done.

saracarl:
So I found a scenario that isn't working as expected.  
The setup:
4 page work:  page 1 is transcribed, pages 2-3 aren't, and page 4 is marked blank.
The reproduce:
Start on page 1, page through using the arrows.  Each page takes you to the next.  When you go from page 3 to page 4 (the blank page) you are on the "overview" tab instead of the "transcribe" tab.  The overview tab has the next arrow greyed out.  The transcribe tab does not (and takes you back to page 2).

It only happens on the last page in the "overview".  I don't know if it's the right behavior or not.  

I will discuss with @benwbrum and we'll figure out if we want the last page over tab arrow to be active or inactive.

saracarl:
Ok.  @benwbrum and I decided the behavior as it stands is fine.  We don't think that many folks will be using "overview" and it might surprise them to be randomly taken elsewhere if they are paging through for reading purposes.  Stopping at the end makes sense.

PR #1550, opened 2019-10-21
Intercom trial user

saracarl:


app/views/shared/_intercom_customer.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+<% if controller_name=="dashboard" && action_name=="owner" %>
+	var anchor_value;
+	var stripped_url = document.location.toString().split("#");
+	if (stripped_url.length > 1) {
+	  anchor_value = stripped_url[1]; 
+	  if (anchor_value == "freetrial") {
+	   Intercom('trackEvent', 'trial-signup');
+     }
+    }
```
benwbrum: We should consider moving this logic to the ruby, right after the user creation and tagging.

PR #1567, opened 2019-12-05
Rails 6

diegoviola:


app/views/document_sets/settings.html.slim
```diff
@@ -75,7 +76,7 @@
         =button_tag 'Add'
         =label_tag 'user_id', "Add Collaborator", class: 'hidden'
       -@collaborators.each do |user|
-        -contributions = user.deeds.count :conditions => "collection_id = #{@collection.slug}"
+        -contributions = user.deeds.where(collection_id: @collection.slug).count
```
saracarl: Should this be passed `@collection.slug` or `@collection.id`?  (I realize that the code is testing the same thing; I just suspect the original code was wrong (or dependent on `friendly_id`).

config/locales/en.yml
```diff
@@ -16,29 +16,18 @@
...
-        one:   "<b>1</b> record"
-        other: "All <b>%{count}</b> records"
-      multi_page: "Records %{from} - %{to} of %{count}"
-      multi_page_html: "Records <b>%{from}</b>&nbsp;&ndash;&nbsp;<b>%{to}</b> of <b>%{count}</b>"
-
-  datetime:
-    distance_in_words:
-      x_days:
-        one: 24 hours
-        other: "%{count} days"
```
saracarl: Are we sure we want to delete all of these, with our internationalization effort coming up?  Does having them break things?

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
+# 'f' after migrating old data.
+# Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
+
+# Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
+# Rails.application.config.active_support.use_sha1_digests = true
+
+# Make `form_with` generate id attributes for any generated HTML tags.
+# Rails.application.config.action_view.form_with_generates_ids = true
```
saracarl: We should probably walk through all of these options together.

app/views/deed/_deeds.html.slim
```diff
@@ -12,8 +12,5 @@
 
 small.legend
   =link_to 'Show More',
-    { :controller => 'deed',
-      :action => 'list',
-      :collection_id => @collection ? @collection.id : nil,
-      :user_id => options[:user_id] },
-    class: 'button outline round'
\ No newline at end of file
+    deed_list_path(:user_id => options[:user_id]),
```
saracarl: This looks like it formerly worked both from the user context and from the collection context.  We should make sure we test (and understand how to test) the collection context.

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+# will be removed in Rails 6.1. This setting is not backwards compatible with earlier Rails versions.
+# If you send mail in the background, job workers need to have a copy of
+# MailDeliveryJob to ensure all delivery jobs are processed properly.
+# Make sure your entire app is migrated and stable on 6.0 before using this setting.
+# Rails.application.config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
+
+# Enable the same cache key to be reused when the object being cached of type
+# `ActiveRecord::Relation` changes by moving the volatile information (max updated at and count)
+# of the relation's cache key into the cache version to support recycling cache key.
+# Rails.application.config.active_record.collection_cache_versioning = true
```
saracarl: We should probably walk through all of these options together.

config/initializers/inflections.rb
```diff
@@ -3,18 +3,12 @@
...
-  # inflect.singular /^(ox)en/i, '\1'
-  # inflect.irregular 'person', 'people'
-  # inflect.uncountable %w( fish sheep )
-
-  inflect.plural /(lea)f$/i, '\1ves'
-  inflect.singular /(lea)ves$/i, '\1f'
-  
-  inflect.plural /(canvas)$/, '\1es'
-  inflect.singular /(canvas)es/, '\1'
-end
```
saracarl: Don't we want these inflections?  They look ftp-specific to me.

config/initializers/assets.rb
```diff
@@ -1,5 +1,14 @@
...
+
+# Add additional assets to the asset load path.
+# Rails.application.config.assets.paths << Emoji.images_path
+# Add Yarn node_modules folder to the asset load path.
+Rails.application.config.assets.paths << Rails.root.join('node_modules')
+
+# Precompile additional assets.
+# application.js, application.css, and all non-JS/CSS in the app/assets
+# folder are already added.
+# Rails.application.config.assets.precompile += %w( admin.js admin.css )
```
saracarl: This needs a joint review to make sure we are handling all of our js assets.  Can we load all of these as node_modules?  We had a hard time with the asset pipeline in our other (from scratch) rails 6 project.

I think this is going to take more work.

config/initializers/ahoy.rb
```diff
@@ -1,3 +1,7 @@
-class Ahoy::Store < Ahoy::Stores::ActiveRecordTokenStore
-  Ahoy.track_visits_immediately = true
+class Ahoy::Store < Ahoy::DatabaseStore
+  def visit_model
+    Visit
+  end
 end
+
+Ahoy.user_agent_parser = :legacy
```
saracarl: This is a big difference; let's make sure it works.  (I don't think we have any tests for Ahoy, but we might.)

config/environments/production.rb
```diff
@@ -16,71 +14,57 @@
...
-  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  # Store uploaded files on the local file system (see config/storage.yml for options).
+  config.active_storage.service = :local
 
-  # Set to :debug to see everything in the log.
-  config.log_level = :info
-  
-  #Ahoy logging should not geocode or you'll get errors
-  Ahoy.geocode = false
```
saracarl: We'll need this.

config/environments/production.rb
```diff
@@ -16,71 +14,57 @@
...
+  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # config.force_ssl = true
 
-  # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
```
saracarl: I don't think we want this to be debug by default.  info is probably good enough.

config/environments/production.rb
```diff
@@ -16,71 +14,57 @@
...
-    :user_name => "XXXXX",
-    :password  => "XXXXX",
-    :authentication => 'login',
-    :domain => 'yourdomain.com', # your domain to identify your server when connecting
-  }
-  config.action_mailer.default_url_options =  { host: 'hostname' } #change this to match your server URL, i.e. www.fromthepage.com
-  config.action_mailer.default_options
-  # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
-
```
saracarl: Where did action_mailer config move?  We need it.

config/environments/production.rb
```diff
@@ -89,19 +73,40 @@
...
+  # replica.
+  #
+  # By default Rails will store a last write timestamp in the session. The
+  # DatabaseSelector middleware is designed as such you can define your own
+  # strategy for connection switching and pass that into the middleware through
+  # these configuration options.
+  # config.active_record.database_selector = { delay: 2.seconds }
+  # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
+  # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 end
```
saracarl: This file needs a lot of review and thought.  We'll need to get it right for the development/master branch, and then we'll need to configure it for deploying to fromthepage.com.  (Staging server should come in useful there.)

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: test
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  channel_prefix: fromthepage_production
```
saracarl: Are we sure we're using redis?  Where or how is it configured?

config/environment.rb
```diff
@@ -1,8 +1,5 @@
 # Load the Rails application.
-require File.expand_path('../application', __FILE__)
-
-Encoding.default_external = Encoding::UTF_8
-Encoding.default_internal = Encoding::UTF_8
+require_relative 'application'
```
saracarl: Any idea why we don't need these?  (Or do we?)  Where did they move?

config/environments/development.rb
```diff
@@ -10,59 +9,72 @@
...
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
 
-  ## Config for MailCatcher ##
-  # Install mailcatcher locally on your machine 'gem install mailcatcher'
-  # Run 'mailcatcher' in the terminal to start the server
-  # Open 'http://localhost:1080/' in your browser to see mail sent
-  config.action_mailer.delivery_method = :smtp
-  config.action_mailer.smtp_settings = { :address => "localhost", :port => 1025 }
+  config.action_mailer.perform_caching = false
```
saracarl: what does this do?

config/environments/development.rb
```diff
@@ -10,59 +9,72 @@
...
 
-  # Adds additional error checking when serving assets at runtime.
-  # Checks for improperly declared sprockets dependencies.
-  # Raises helpful error messages.
-  config.assets.raise_runtime_errors = true
-  
-  # http://pontiiif.brumfieldlabs.com/api/v0.0/search/Irish
-  config.pontiiif_server = 'http://pontiiif.brumfieldlabs.com/'
-  
-  Ahoy.geocode = false
```
saracarl: suspect we need this.

config/environments/development.rb
```diff
@@ -10,59 +9,72 @@
...
+
+  config.action_mailer.default_url_options =  { host: 'localhost:3000' }
+
   # location of system calls on this machine
   NEATO = '/usr/bin/neato'
   RAKE = '/usr/bin/env rake'
-  TEX_PATH='/usr/local/texlive/2017/bin/x86_64-linux/'  
+  TEX_PATH='/usr/local/texlive/2017/bin/x86_64-linux/'
+
+  config.pontiiif_server = 'http://pontiiif.brumfieldlabs.com/'
```
saracarl: we can delete this.

config/environments/development.rb
```diff
@@ -10,59 +9,72 @@
...
 
+  ## Config for MailCatcher ##
+  # Install mailcatcher locally on your machine 'gem install mailcatcher'
+  # Run 'mailcatcher' in the terminal to start the server
+  # Open 'http://localhost:1080/' in your browser to see mail sent
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = { :address => "localhost", :port => 1025 }
```
saracarl: let's make sure all the action_mailer configs are next to each other.  (There's an action_mailer cache setting somewhere higher in this file.)

app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
```
saracarl: What is the story on this new `ApplicationRecord` abstract class?  This isn't a pattern I've seen before, although our one Rails 6 application does not use ActiveRecord, so we might not have encountered it.

app/models/category.rb
```diff
@@ -1,19 +1,18 @@
-class Category < ActiveRecord::Base
+class Category < ApplicationRecord
   extend ActsAsTree::TreeWalker
 
   acts_as_tree :order => 'title'
-  belongs_to :collection
-  has_and_belongs_to_many :articles, -> { order('title').uniq }
-  attr_accessible :collection_id, :title, :gis_enabled
+  belongs_to :collection, optional: true
+  has_and_belongs_to_many :articles, -> { order('title').distinct }
```
saracarl: Do we not need `attr_accessible` any more in Rails 6?

app/models/sc_manifest.rb
```diff
@@ -79,32 +78,32 @@ def convert_with_collection(user, collection)
...
     work
   end
 
   def cleanup_label(label)
     label = flatten_element(label)
     new_label = label.truncate(255, separator: ' ', omission: '')
-    new_label.gsub!("&quot;", "'")
-    new_label.gsub!("&amp;", "&")
-    new_label.gsub!("&apos;", "'")
+    new_label.dup.gsub!("&quot;", "'")
```
saracarl: I don't understand why we're duplicating a string and then modifying the duplication.  Shouldn't we be modifying the label itself?  I feel like this needs testing if the test suite doesn't already test it.  (Can we use Coveralls to find out whether this is being tested?)

app/models/xml_source_processor.rb
```diff
@@ -72,37 +72,35 @@ def validate_links(text)
...
+
     source_text = case text_type
-    when Page::TEXT_TYPE::TRANSCRIPTION
-      page.source_text
-    when Page::TEXT_TYPE::TRANSLATION
-      page.source_translation
-    else
-      ""
-    end
+                  when Page::TEXT_TYPE::TRANSCRIPTION
```
saracarl: These are either hard tabs or eight-space indentation and need to be replaced by two-space indentation.

spec/spec_helper.rb
```diff
@@ -23,7 +23,7 @@
 
 # Checks for pending migrations before tests are run.
 # If you are not using ActiveRecord, you can remove this line.
-ActiveRecord::Migration.maintain_test_schema!
+#ActiveRecord::Migration.maintain_test_schema!
```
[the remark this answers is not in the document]
diegoviola: @saracarl see the commit message here: https://github.com/benwbrum/fromthepage/commit/a66a0cb544b23e02b7cf3dab1cd66cba9d17ceaf

app/views/user_mailer/work_collaborator.text.erb
```diff
@@ -2,9 +2,9 @@ Hi <%=@user.display_name%> --
 
 We just wanted to let you know that <%=@work.owner.display_name%> added you as a collaborator on <%=@work.title%>.
 
-You can view #{@work.collection.title}: #{@work.title} at the URL: <%=collection_read_work_path(@work.owner, @work.collection, @work, only_path: false)%>
+You can view #{@work.collection.title}: #{@work.title} at the URL: <%=collection_read_work_url(@work.owner, @work.collection, @work, only_path: false)%>
```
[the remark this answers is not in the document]
diegoviola: @saracarl "path is relative while url is absolute".

But aside from that, `*_path` no longer works in email views after this commit: https://github.com/rails/rails/commit/d282125a18c1697a9b5bb775628a2db239142ac7

config/application.rb
```diff
@@ -8,41 +8,18 @@
...
 
     config.neato = '/usr/bin/env neato'
-    config.encoding = 'utf-8'
-
-    config.action_mailer.preview_path = "#{Rails.root}/spec/mailers/previews"
-
-    config.action_dispatch.default_headers = {
-        'Access-Control-Allow-Origin' => '*',
-        'Access-Control-Request-Method' => "GET"
-      }
```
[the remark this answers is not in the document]
diegoviola: @saracarl it got removed when doing the `rails app:update` when I was working on the rails 5 upgrade. I added it back.

config/locales/en.yml
diegoviola: @saracarl this might have been removed when doing the upgrade, having them doesn't break anything. I added it back.

app/models/xml_source_processor.rb
diegoviola: @saracarl Should we? My editor (Vim) disagrees and says the line is indented properly.

app/models/application_record.rb
diegoviola: This is the new default starting from Rails 5, see: https://github.com/rails/rails/pull/22567

app/models/category.rb
diegoviola: @saracarl No, `attr_accessible` is gone and should not be used anymore (the alternative is to use strong parameters).

config/initializers/inflections.rb
diegoviola: @saracarl I added those back.

config/environments/production.rb
diegoviola: @saracarl done.

config/environments/development.rb
diegoviola: @saracarl are you referring to the `TEX_PATH` variable?

That's used in `lib/tasks/config_test.rake` and `app/models/tex_figure.rb`.

config/environments/production.rb
diegoviola: @saracarl I added them back.

config/environments/development.rb
diegoviola: @saracarl those were added again.

app/models/sc_manifest.rb
diegoviola: @saracarl I can't remember exactly why I did this, I just reverted it and the tests are still passing locally.

config/cable.yml
diegoviola: @saracarl usually running `redis-server` is enough, but I think this can be safely ignored until/unless we start using Action Cable.

config/environments/development.rb
diegoviola: @saracarl done, thanks.

app/views/deed/_deeds.html.slim
diegoviola: @saracarl I just restored the previous functionality where we pass `collection_id`.

I removed that because I was experiencing breaks during the upgrade, and removing `collection_id` worked as a workaround.

It would still be nice to know how to trigger the collection context.

Good find, thanks.

config/initializers/ahoy.rb
diegoviola: @saracarl I added this to restore the original Ahoy behavior, some of our tests (`spec/features/ahoy_spec.rb`) break if we remove this option.

app/views/document_sets/settings.html.slim
diegoviola: @saracarl the query is basically the same, except the proposed change uses newer syntax.

config/environment.rb
diegoviola: @saracarl I don't think we need these lines, UTF-8 is the default.

config/environments/development.rb
diegoviola: @saracarl https://guides.rubyonrails.org/configuring.html#configuring-action-controller

config/environments/production.rb
diegoviola: @saracarl I just restored that config.

config/environments/development.rb
saracarl: Let's delete
` config.pontiiif_server = 'http://pontiiif.brumfieldlabs.com/'`

diegoviola: @saracarl do we really want to remove that? it's used in a lot of places.

diegoviola: @saracarl I just removed it.

PR #1571, opened 2019-12-10
Update transcriptions in CONTENTdm from FromThePage

benwbrum:


lib/contentdm_translator.rb
```diff
@@ -121,33 +121,40 @@ def self.fst_field_for_collection(collection, license_key, contentdm_user_name,
...
-      fts_field = doc.search("//field/type[text()='FTS']").first.parent.search('name').text
+      # good response
+      fts_field = doc.search("//field/type[text()='FTS']").first.parent.search('nickname').text
     end
 
     return error, fts_field
   end
 
   def self.export_work_to_cdm(work, username, password, license)
+    error, fieldname = fst_field_for_collection(work.collection, license, username, password)
```
saracarl: Should "fst_field" be "fts_field" (full text search)?

saracarl: I guess not -- maybe it was a typo the first time we wrote this?

saracarl: I'd like it updated to be fts_field instead of fst_field.  Better readability in the future.

PR #1575, opened 2019-12-12
collection: add individual researcher message

diegoviola:


app/views/collection/new.html.slim
```diff
@@ -1,5 +1,9 @@
 .litebox-embed(style="width:550px")
   h1 Create New Collection
+
+  - if current_user.account_type == "Individual Researcher"
+    p(style="color: #FF0000") Individual Researcher Accounts are limited to a single collection. To upgrade your account, contact support@fromthepage.com.
+
```
saracarl: So this should not show the form for creating a new collection if 
current_user.account_type == "Individual Researcher"
AND
current_user collection count => 1  

It should only show the message if 
current_user.account_type == "Individual Researcher"
AND
current_user collection count => 1 

In other words, one or the other.

saracarl:
Tests passed locally, though!

benwbrum:
We still need to hide the form if we display the warning message.

diegoviola:
@benwbrum done.

Note that I had to change the account_type to "Trial" for those specific tests, otherwise the form gets disabled and causes the tests to break.

PR #1579, opened 2019-12-19
1578 add formats export all works

diegoviola:


saracarl:
1.  Description in the UI currently says: "containing each work as an XHTML file with transcripts, user comments, subject articles, and internal HREFs linking subjects and pages."  and needs to be changed to say "containing each work as HTML and plain text.
2.  The old .xhtml file needs to be removed from the zip file -- the html directory serves that purpose.
3.  If no translation enabled for the work, do not include any _translation files or folders.
4.  If subject linking is disabled, do not include any _emended files or folders.
5.  The README should only contain a description of what each type of file is.  It should not list every file.  


HTML:
full -- full text of the transcription or translation, with versions, contributors, and comments.
text -- just the full text of the transcription or translation


Text:
verbatim -- full text of the transcription, with versions, contributors and comments.
searchable -- links removed and line breaks removed so searches can span line breaks.
emended -- replaces linked subjects' verbatim with canonical subject names.  Useful for text analysis.

saracarl:
I'm getting a:
```
No route matches {:action=>"export_work", :collection_id=>"iiif-showcase-2018", :controller=>"export", :format=>:zip, :user_slug=>"saracarl", :work_id=>15367}
```
For anything I try to export.

saracarl:
Let's also change the `emended_transcript` and `emended_translation` filenames to be `expanded_transcript` and `expanded_translation`.  The README should also explain them:
_Expanded transcripts and translations replace the verbatim linked subjects with the canonical name of the subject to make it suitable for text analysis._

diegoviola:
@saracarl I just added the changes you suggested, please give it a try.

saracarl:
Error when I tried to export all works:

```NameError at /export/export_all_works.zip
uninitialized constant ExportController::README
Did you mean?  Readline```

diegoviola:
> Error when I tried to export all works:
> 
> ```
> uninitialized constant ExportController::README
> Did you mean?  Readline```
> ```

@saracarl I can't reproduce this error, are you sure you're seeing this with the latest commit?

saracarl:
The html directories for each work have a translation.html even if translation is not enabled.  We need to remove those.

diegoviola:
> The html directories for each work have a translation.html even if translation is not enabled. We need to remove those.

@saracarl please try https://github.com/benwbrum/fromthepage/pull/1579/commits/a7c7556ac948d5a2c7b96abc955ba3dcd2cee4db.

Issue #1582, opened 2019-12-20
Downgrade owner account functionality


benwbrum:
When a trial user account expires, we need to revert the user back to a regular transcriber.

We need a `downgrade` method on the `User` model that
* Removes the `owner` bit.
* Loops through any collections they created, and for each collection 
  * Makes the collection inactive
  * Makes the collection private

The adminstator's Edit User functionality should include a Downgrade button for owners that will call this functionality.

saracarl:
This will need a decent amount of UI testing (and probably discussion) -- I have no idea what will happen if we have owned collections but not a owner bit.  Will they show up in that user's dashboard?  I doubt it.  That's probably OK.  (This does imply there's no way but exiting links to get to those collections, which means they are effectively orphaned.)  What if we flip the owner bit back?  I assume they'll show up.  This is probably the only way to unorphan the collections.  I'm curious how often we'll do that.

PR #1591, opened 2019-12-27
admin: add downgrade owner account

diegoviola:
Fixes: #1582.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -223,6 +223,12 @@ class AdminController < ApplicationController
     end
   end
 
+  def downgrade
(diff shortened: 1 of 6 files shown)


benwbrum:
I ran into a couple of problems when testing.

* If I downgrade `SSM_Archives` (id=215749 in the development dump file), they are marked as not an owner, which is correct.  However I'd also expect this:
  * Their collection still returns `true` when I call `active?`
  * Their collection still returns `false` when I call `restricted?`
  * Their `user.account_type` is still `"Trial"`.  It should be `nil`

I'm not sure what's wrong, since I thought the tests covered all of this (except for `account_type`, which I forgot to specify in the issue.)

PR #1595, opened 2020-01-06
collection: differentiate owner accounts

diegoviola:


saracarl:
So we missed a detail on this feature (sigh).  For the following reproduction:
```
create an owner. "Wakanda"
owner creates a collection "Letters from America"
create a non-owner user "Shuri"
Wakanda adds Shuri as a collection owner.
Confirm Shuri can see the "Letters from America" collection settings.
Create a collection, "Science Archives" as Shuri (menu -> "create new collection" or upload and create one there)
"Science Archives" should be listed under the Wakanda account as a collection. (It should look the same as "Letters from America")
```
Shuri needs to be able to see the "Science Archives" collection and can't.  (She should be automatically added as a secondary collection owner to "Science Archives" and she isn't.)

saracarl:
Can we also add some tests for this feature, please?

diegoviola:
@saracarl done.

spec/features/owner_actions_spec.rb
```diff
@@ -270,5 +269,62 @@
...
+        click_button "Add"
+      end
+      @user.reload
+      expect(@user.owner).to be(true)
+      expect(@user.account_type).to eq "Staff"
+    end
+
+    it "confirms that Shuri can read Wakanda's collection" do
+      logout
+      login_as @user
```
saracarl: I don't think this is logging in as Shuri, is it?

saracarl: How does it know to use user9 instead of user8?

saracarl: Got it.  Looks good.

Issue #1576, opened 2019-12-12
Extract existing text from PDFs for OCR correction


bencomp:
I would like to try correct the OCR text that is embedded in PDF files.Is there a way that FromThePage automatically extracts the text for each page on (or after) import?

It looks like the only way to start a OCR correction project currently is from importing from the Internet Archive or ContentDM, isn't it?

bencomp:
Alternatively, I could create OCR plain text files and images for each page and upload them in a zip file, but that doesn't seem to be currently supported either.

benwbrum:
That's an accurate description of the current functionality, however we received word on Monday that a new client will fund development of a feature to extract the text layer from uploaded PDFs and use it as raw OCR to be corrected.  Would you be willing to contribute example PDFs (to use as development/test data) and possibly do some early testing on the feature?  I suspect we will be building it in the next month.

PR #1600, opened 2020-01-10
Added OCR ingestion for #1576

benwbrum:


saracarl:
Running tests on this has not been done by Travis.  Several heisenbugs were found when this was merged into Rails 4, so we may need to check this carefully.

saracarl:
We need to merge from the Rails 4 version of this branch into this branch before it is merged into rails6

benwbrum:
This was handled when @diegoviola merged `development` into `rails6`.

PR #1601, opened 2020-01-11
admin: fix user search

### LINKED ISSUE #1594: Error when searching for a user (Administration -> Users)
The following error occurs when searching for a user:

```
 undefined method `visits' for nil:NilClass 
```
This can be reproduced on the `rails6` branch, works fine on `development`.

diegoviola:
Fixes #1594


benwbrum:
This works as expected.

PR #1609, opened 2020-01-16
Revert "Disable prettyfied html in development"

diegoviola:
This reverts commit 36ee86a6f662b496894574fc433f905f41616973.


diegoviola:
Reverting this commit because it breaks 10 other tests.

PR #1613, opened 2020-01-23
Handle filenames with spaces

benwbrum:


diegoviola:
@benwbrum looks good to me :+1:

Issue #1610, opened 2020-01-21
Attribution name for users in sign-up/edit forms and exports


saracarl:
FromThePage will expose a new "Attribution Name" field on both the sign-up and profile-edit forms.  * Additional wording will be added to explain that, if users fill out this field, they would like to be credited for their work (at the project owner's discretion) in exports and publications.  Additional clarifications to the wording on the form will explain the "display name" in greater detail.
* We'll review export formats from FromThePage.  When appropriate (i.e. not searchable plain text, but yes for TEI) exports will include the attribution name for contributions by users who have filled that in.  (Pseudonyms/logins will continue to appear where they currently do, as they are used by project owners for evaluating quality.)

We're shooting to have this done by the end of February, so that UT IT can deploy it to the UT FromThePage instance.

saracarl:
We can re-use the old `print_name` attribute of the User model for this.  It was designed for a very similar purpose, but was retired.

saracarl:
Sign Up Screen Changes:

![UpdatedSignUpScreen](https://user-images.githubusercontent.com/708566/73756948-09ff1080-472e-11ea-8c9b-55886eb18256.png)


- [ ] Remove the "Please fill in the following..."
- [ ] Change the label "Login" to "User Name" (same actual field)
- [ ] Add "You'll use this name to log in.  It will be shown publicly." in smaller font and italics under the User Name field.
- [ ] Change "Password confirmation" to "Confirm Password"
- [ ] Chance the label "Display name" to "Real Name"
- [ ] Add "Projects may use this when giving you credit.  Leave this blank if you do not want to be credited."

saracarl:
Edit Account Screen Changes:
![EditAccountScreen](https://user-images.githubusercontent.com/708566/73758646-81ce3a80-4730-11ea-9f5a-04e4052a8dd1.png)


- [ ] Change "Your login" to "User Name"
- [ ] Add a field "Real Name" (Use the User model attribute "print_name")
- [ ] Add  "Projects may use this when giving you credit. Leave this blank if you do not want to be credited."  in smaller font and italics under the User Name field.

saracarl:
We should migrate/rename that `print_name` attribute to `real_name` (probably before doing all of the UI work!)

saracarl:
The other half of this issue is to use `real_name` in any export formats that mention users.  Those are only HTML and TEI exports.  We believe that the current logic is similar to:
`print display_name then username`
This would change that logic to:
[...]

PR #1619, opened 2020-02-07
1610 attribution name for users

diegoviola:
#1610

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rvm: 2.3.7
 jobs:
   include:
     - stage: test for pry
-      rvm: 2.4.1
+      rvm: 2.7.0
       before_install:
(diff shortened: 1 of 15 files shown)


saracarl:
I still don't see any travis tests of this PR, @diegoviola

diegoviola:
> I still don't see any travis tests of this PR, @diegoviola

@saracarl I'm aware of that, but I need more information to be able to help with the Travis problem.

I had a look at the `.travis.yml` file (diffed it against other branches, etc) and couldn't find anything relevant.

benwbrum:
I couldn't see anything wrong with `.travis.yml`, but I updated it (increased the ruby version for our check for `pry` -- this should be a no-op) and pushed and it looks like it's running now.

I have no clue what went wrong before.

PR #1632, opened 2020-02-21
ahoy: stop tracking still_editing calls

diegoviola:


benwbrum:
This fixes our absurd-number-of-minutes problem!

Issue #1588, opened 2019-12-21
User export for mailing lists (2020 P4)


benwbrum:
Clients have requested the ability to download email addresses for all users who have worked on their projects, along with opt-out information and activity summaries, so that they can add them to monthly newsletters or other communications.

We've manually created a sample export like this for USHMM, who has approved it.  Now we should add it to the product.  The Owner Dashboard should have a new action, "Mailing List Export" which sends a spreadsheet to the user containing columns for User Login, User Name, Email, Opt-in, and columns for each of the owner's collections.  Each user should appear as a data row, with deed totals as the cell for each collection.

This is an edited listing of the code used to generate the sample spreadsheet, which may be useful:
```
header = ['User Login', 'User Name', 'Email', 'Opt-In']
owner = User.find 'ushmmarchives'
collection_ids = owner.collections.map { |c| c.id }.sort
deed_map = Deed.where(:collection_id => collection_ids).group(:user_id, :collection_id).count
user_ids = deed_map.keys.map {|e| e[0]}.uniq
Collection.where(:id => collection_ids).order(:id).each { |c| header << c.title }
csv << header
User.find(user_ids).each do |user|
  row = []
  row<<user.login
  row<<user.display_name
  row<<user.email
  row<<user.activity_email
  collection_ids.each do |collection_id|
    row << deed_map[[user.id,collection_id]] || 0
  end
print csv_out
File.write('/tmp/ushmm_users.csv', csv_out)
```

(This should be branched off of `development` and migrated to `rails6` if necessary.)

saracarl:
I'd like to add this to the "summary" tab of the owner dashboard, under "collaborators" but before the leader boards.

We should also add a link to the "collaborators" tab for a collection that says:
"A detailed spreadsheet export of all collaborators is available on your owner dashboard."

PR #1638, opened 2020-02-25
1588 user export mailing lists

diegoviola:


app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
     @all_indexers     = build_user_array(DeedType::PAGE_INDEXED)
   end
 
+  def export_csv
+    header = ['User Login', 'User Name', 'Email', 'Opt-In']
+    owner = User.find 'ushmmarchives'
```
benwbrum: Get the user.  (This will need to be changed to the current user, or some parameterized thing.)

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+    collection_ids = owner.collections.map { |c| c.id }.sort
```
benwbrum: Create an array of all the collections this owner owns, sorted by `id`

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+    deed_map = Deed.where(:collection_id => collection_ids).group(:user_id, :collection_id).count
```
benwbrum: Get all the deeds for this collection (because we want to know which users worked on which collections).
   We don't care what kind of deeds the user had, but want to know how many they had to indicate activity on the collection.

Then create an array with `[user_id, collection_id, deed_count]` from the users who worked on the collection.

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+    user_ids = deed_map.keys.map {|e| e[0]}.uniq
```
benwbrum: Now get the list of unique users who worked on those collections by plucking out the first element of the key `[user_id, collection_id]` -- this is the `user_id`, and we want the unique user IDs.

This creates an array of user IDs for all users who worked on this owner's collections.

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+    Collection.where(:id => collection_ids).order(:id).each { |c| header << c.title }
```
benwbrum: Now get all the titles of the collections (ordered by `id`) and add them to the header array.

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+
+    rows = []
```
benwbrum: Initialize the rows -- these should contain one row per user, with four standard columns (email, opt-in, etc) and one column per collection.

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+
+    User.find(user_ids).each do |user|
+      rows << [user.login, user.display_name, user.email, user.activity_email]
+    end
```
benwbrum: Add all the users to the rows array, but only with the first four columns.

(How are we going to figure out which user belongs to which set of deeds?  There is no `order` clause, so I worry that we're going to mix things up.)

benwbrum: Okay, the original sample code in #1588 includes this:
```

  collection_ids.each do |collection_id|
    row << deed_map[[user.id,collection_id]] || 0
  end
```
This loops through each of the collections for the user (preserving the order we used earlier)
Then it constructs a key for the `deed_map` hash out of the ID for this row's user and the ID for this column's collection.  
This `[user.id, collection.id]` key should return a deed count from `deed_map`, which is appended as a new cell onto this user's row.

app/controllers/statistics_controller.rb
```diff
@@ -12,6 +14,31 @@ def collection
...
+
+    csv_string = CSV.generate(headers: true) do |csv|
+      csv << header
+
+      rows.each do |row|
+        csv << row
+      end
+    end
+
+    send_data csv_string, filename: "ushmm_users.csv"
```
benwbrum: We'll need to rename the file to something like `fromthepage_user_emails.csv`; possibly with a date suffix on it.  (See what we do for other CSV export filenames.)

diegoviola: @benwbrum done.

app/controllers/statistics_controller.rb
diegoviola: @benwbrum done as well. :+1:

app/controllers/statistics_controller.rb
diegoviola: @benwbrum this was an oversight on my part, thanks for your help :+1:

diegoviola:
> "Spinny" never goes away when you download the csv file. Needs to.

@saracarl I just pushed a commit that should hide the spinner.

Issue #14, opened 2012-06-07
Works should not be transcribable without a collection.


benwbrum:
It's possible for works to exist without a collection, but only in an intermediate state.  We should force owners to add works to a collection so that they do not run into errors when they try to transcribe works in an invalid state..

PR #1641, opened 2020-03-03
1640 import metadata iiif manifest

diegoviola:
Screenshot_2020-06-02 Facets - test FromThePage


app/views/collection/edit.html.slim
```diff
@@ -186,6 +186,10 @@
     =link_to(collection_blank_collection_path(:collection_id => @collection.id), class: 'button')
       span Blank Collection
 
+    h3 Metadata
```
benwbrum: Can you move this above the _Blank Collection_ buttoninstead of below it?

benwbrum:
I think that the next step on this is to produce the exemplar CSV file.

We need a controller action that produces a CSV file (approximately) like this:

| work_id        | title           | your_metadata_field_one  | your_metadata_field_two |
| ------------- | ------------- | ----- | ----- |
| 2345      | Letter from Jonas Smith to James Smithers |  |  |
| 2346      | Letter from James Smithers to Jonas Smith     |   |  |
| 2653 | James Smithers Diary |   |  |

This should be created by code similar to this pseudocode:

```
csv = CSV.new(header_row:true)
csv.headers = ['work_id', 'title', 'your metadata_field_one', 'your_metadata_field_two']
@collection.works.each do |work|
  csv << [work.id, work.title]
end
render data: csv
```

app/models/metadata.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    csv_string
+  end
+
+  def self.create_example(collection)
+    csv_string = CSV.generate(headers: true) do |csv|
+      csv << ['work_id', 'title', 'your metadata_field_one', 'your_metadata_field_two']
+
+      collection.works.each do |work|
+        csv << [work.id, work.title]
```
benwbrum, line 97: At some point we'll want to pull out other fields here, but that's going to take some more thought.  This looks fine for now.

app/views/metadata/upload.html.slim
```diff
@@ -0,0 +1,12 @@
+.litebox-embed(style="width:550px")
+  h1 Upload Metadata
+  p To update metadata for several works within this collection, upload a spreadsheet with one row per work. See #{link_to "link", collection_metadata_example_path(params[:id])} for an example. (Note that multi-line/description fields cannot be updated.
+
+  =form_for(:metadata, url: collection_metadata_create_path) do |f|
+    table.form
+      tr.big
+        th =f.label :file, 'Metadata File'
+        td.w100.aright =f.file_field :file
```
benwbrum, line 9: Please make this UI match the file upload UI for uploading works.

app/controllers/metadata_controller.rb
```diff
@@ -0,0 +1,25 @@
+class MetadataController < ApplicationController
+  layout Proc.new { |controller| controller.request.xhr? ? false : nil }, :only => :upload
```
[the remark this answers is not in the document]
diegoviola: @benwbrum I wanted to suppress the layout for the `upload` action, I was not sure it would be needed for the other actions.

But it's clear that we don't need it now, I changed it to `layout false` as you suggested, thanks.

app/views/metadata/upload.html.slim
diegoviola: @benwbrum done, thanks.

app/models/metadata.rb
```diff
@@ -0,0 +1,130 @@
...
+      @new_metadata.each do |m|
+        @canonical_metadata << { m[:label] => 1 }
+      end
+    else
+      json_metadata = JSON.parse(work.collection.canonical_metadata)
+
+      unless json_metadata.empty?
+        json_metadata.each do |c|
+          c.each do |k, v|
+            c[k] = v + 1
```
benwbrum, line 82: It occurs to me that this will increment the metadata coverage count every time a work is updated if we do this in an `after_save` callback.  Other options include doing this in an `after_create` call-back or only when the metadata field has been updated.

benwbrum:
I've run into a problem importing metadata and seeing them in the facets, in that I import two IIIF manifests, then view the Facets tab for the collection and don't see any data there.

To reproduce:
* Create a new collection
* Enable Metadata Facets
* Add work the collection
* IIIF Import https://wellcomelibrary.org/iiif/b30352253/manifest
* Add another work to the collection
* IIIF Import https://data.ucd.ie/api/img/manifests/ivrla:2638
* View the Facets tab of the collection

diegoviola:
> I've run into a problem importing metadata and seeing them in the facets, in that I import two IIIF manifests, then view the Facets tab for the collection and don't see any data there.

@benwbrum we currently extract metadata only when the user manually uploads it (i.e. "Upload Metadata").

The existing Metadata work does not affect any of the "Start A Project" actions.

diegoviola:
@benwbrum should we change it to extract metadata for "Start A Project" actions as well?

benwbrum:
We definitely want to call the metadata code when importing IIIF manifests or CONTENTdm items.

diegoviola:
> We definitely want to call the metadata code when importing IIIF manifests or CONTENTdm items.

This should be working now.

benwbrum:
If I try to enable facets on the collection with ID `981`, I get an ActiveRecord error:

```
ActiveRecord::RecordNotSaved - You cannot call create unless the parent is saved:
  app/controllers/facets_controller.rb:21:in `block (2 levels) in enable'
  app/controllers/facets_controller.rb:12:in `block in enable'
  app/controllers/facets_controller.rb:9:in `enable'
  app/controllers/application_controller.rb:29:in `switch_locale'
```

diegoviola:
> If I try to enable facets on the collection with ID `981`, I get an ActiveRecord error:
> 
> ```
> ActiveRecord::RecordNotSaved - You cannot call create unless the parent is saved:
>   app/controllers/facets_controller.rb:21:in `block (2 levels) in enable'
>   app/controllers/facets_controller.rb:12:in `block in enable'
>   app/controllers/facets_controller.rb:9:in `enable'
>   app/controllers/application_controller.rb:29:in `switch_locale'
> ```

This should be fixed now with the commit: https://github.com/benwbrum/fromthepage/pull/1641/commits/a5a7304398de9fcb614abc9951e2511266afe1e7 -- please give it another try and let me know if you can still reproduce it. Thanks.

app/models/facet_config.rb
```diff
@@ -0,0 +1,48 @@
...
+        om = JSON.parse(w.original_metadata)
+
+        unless w.work_facet
+          w.create_work_facet
+        end
+
+        om.each do |o|
+          label = o['label']
+
+          if label == self['label']
```
benwbrum, line 22: I just realized that the `label` field which users can enter on `facet_config` refers to the label we plan to display to users searching the collection, not the label in the work's `original_metadata` that lets us look up fields. 

We have two separate ideas of "labels" on `facet_config`:
* `label` is a text string entered by the user configuring facets, and is stored within `facet_config` but *is not used to look up values in `work.original_metadata`*
* `original_label` is a text string from `metadata_coverage` which the user cannot edit.  It (currently) lives in `metadata_coverage.key` and it *should be used to look up values in `work.original_metadata`.

I think that changing line 22 to be `if label == self.metadata_coverage.key` will solve this issue.

saracarl:
I tested this using the following 3 documents imported via the "import from ContentDM" option (which pulls in metadata):
http://collections.msdiglib.org/digital/collection/mdah/id/360
http://collections.msdiglib.org/digital/collection/mdah/id/402
http://collections.msdiglib.org/digital/collection/mdah/id/724

And then I used the "Date" "Coverage" "Mississippi County" and "Geographic Location" metadata fields:
Screenshot from 2020-10-22 11-48-42
Screenshot from 2020-10-22 11-48-54

1. Let's remove this string from the collection settings UI: (You can configure the metadata facets in a separate step.)
2. After clicking the "Enable Facets" button it should redirect to the "Facets" tab.
3. We say "Configure metadata facets by reviewing the metadata currently present in your collection, and selecting fields to be displayed to users."  How do you select fields?  Does just adding a label do it?  If so, we should rewrite this "Configure metadata facets by reviewing the metadata currently present in your collection, and adding a label to fields to be displayed to users."

Once I did the above, this is what I see:
Screenshot from 2020-10-22 11-46-19 (sorry this is so big -- Github!)
4. The data ranges are 1500-2020.  The documents are all from 1860.  We should derive the dates for the data sliders based on the greatest and smallest dates.  (This will be weird for time period because everything is 1860 -- but that's OK.)
5.  When another facet is selected (i.e. "Hinds County Ms") the date ranges should adjust to just the ranges in the subset of documents that are Hinds County.  You can refer to the Curran Index code for how to do this.
6.  There are 3 random bullets.  They shouldn't be there.
7. "County" includes a value of "2020-10-22" but that's not in any of the data.
8.  After you select a facet it should have a little "x" next to it to remove it.  See the Curran Index UI for how this looks, and the code for how to implement.  (Ben just points out that the facet value itself is clickable to remove.  That's not a great UI -- we need to remove the hyperlink from the facet value and add an "x".)
9.  Geographic location is the 4th facet, but it's missing.  You may want to look at how we process field configuration code for a reference for that.
10.  The county filtering is working nicely.
11.  The "Date" data has years, months, and days.  The UI is just years.  It needs to have days and months as well.

The above is in the order I discovered them.  The priority order is:
9
7
6
4
11
5
8
then 1-3.

diegoviola:
> Let's remove this string from the collection settings UI: (You can configure the metadata facets in a separate step.)

https://github.com/benwbrum/fromthepage/pull/1641/commits/c4377ad18b4540d2f4ee33e645ed774c4fb9f8be

diegoviola:
> Let's remove this string from the collection settings UI: (You can configure the metadata facets in a separate step.)

https://github.com/benwbrum/fromthepage/pull/1641/commits/c4377ad18b4540d2f4ee33e645ed774c4fb9f8be

diegoviola:
> After clicking the "Enable Facets" button it should redirect to the "Facets" tab.

https://github.com/benwbrum/fromthepage/pull/1641/commits/4237bd43880598b15d0da4e46ada6a57765ed578

diegoviola:
> We say "Configure metadata facets by reviewing the metadata currently present in your collection, and selecting fields to be displayed to users." How do you select fields? Does just adding a label do it? If so, we should rewrite this "Configure metadata facets by reviewing the metadata currently present in your collection, and adding a label to fields to be displayed to users."

https://github.com/benwbrum/fromthepage/pull/1641/commits/7884d182229db6747298949311e8546f82163407

diegoviola:
>  6. There are 3 random bullets. They shouldn't be there.

Should be fixed with: https://github.com/benwbrum/fromthepage/pull/1641/commits/5de347722efcc15d7467c7310fd95897525c2aa0

diegoviola:
>  7. "County" includes a value of "2020-10-22" but that's not in any of the data.

Should be fixed with: https://github.com/benwbrum/fromthepage/pull/1641/commits/fa33e16165aff078c30aada8ddbcb1658b3dd60c and https://github.com/benwbrum/fromthepage/pull/1641/commits/5de347722efcc15d7467c7310fd95897525c2aa0

diegoviola:
>  9. Geographic location is the 4th facet, but it's missing. You may want to look at how we process field configuration code for a reference for that.

Should be fixed with: https://github.com/benwbrum/fromthepage/pull/1641/commits/fa33e16165aff078c30aada8ddbcb1658b3dd60c

saracarl:
Another review and checklist to make the "To Do"s clearer.  I'll update this checklist by editing the Github comment as things are completed.

- [x] 1. Let's remove this string from the collection settings UI: (You can configure the metadata facets in a separate step.)
- [x] 2. After clicking the "Enable Facets" button it should redirect to the "Facets" tab.
- [x] 3. We say "Configure metadata facets by reviewing the metadata currently present in your collection, and selecting fields to be displayed to users." How do you select fields? Does just adding a label do it? If so, we should rewrite this "Configure metadata facets by reviewing the metadata currently present in your collection, and adding a label to fields to be displayed to users."
- [ ] 4. The data ranges are 1500-2020. The documents are all from 1860. We should derive the dates for the data sliders based on the greatest and smallest dates. (This will be weird for time period because everything is 1860 -- but that's OK.)
- [ ] 5. When another facet is selected (i.e. "Hinds County Ms") the date ranges should adjust to just the ranges in the subset of documents that are Hinds County. You can refer to the Curran Index code for how to do this.
- [x] 6. There are 3 random bullets. They shouldn't be there.
- [x] 7. "County" includes a value of "2020-10-22" but that's not in any of the data.
- [ ] 8. After you select a facet it should have a little "x" next to it to remove it. See the Curran Index UI for how this looks, and the code for how to implement. (Ben just points out that the facet value itself is clickable to remove. That's not a great UI -- we need to remove the hyperlink from the facet value and add an "x".)
- [x] 9. Geographic location is the 4th facet, but it's missing. You may want to look at how we process field configuration code for a reference for that.
- [x] 10. The county filtering is working nicely.
- [ ] 11. The "Date" data has years, months, and days. The UI is just years. It needs to have days and months as well.


New changes:
Screenshot from 2020-10-22 14-11-19

- [ ] 12. Let's change "Configure metadata facets by reviewing the metadata currently present in your collection, and adding a label to fields to be displayed to users." to "Configure metadata facets by reviewing the metadata in your collection and labelling fields to be displayed to transcribers." (another change, sorry!)
- [ ] 13. "Time Period" and "Date" with 2020-10-22 shouldn't be there.
- [ ] 14. I imported 2 additional works to my collection (this manifest: https://cdm17313.contentdm.oclc.org/iiif/info/mdah/765/manifest.json twice, via IIIF) and while they are listed in the "works" tab for the collection, they are not in the main collection view.
Screenshot from 2020-10-22 14-53-43
Screenshot from 2020-10-22 14-54-08

- [ ] 15. When I go back and look at the Facets tab (after importing more works) the counts of documents are way off.  There are 5 documents in this collection.  The counts show 14, 11, 2, or 29.
Screenshot from 2020-10-22 14-42-19

- [ ] 16. Probably related to #14, but if I select "Marshall County" I only get 1 document instead of the 3 with that field.
- [ ] 17. When I use the date slider I get a `undefined` method `year' for nil:NilClass` on this line: https://github.com/benwbrum/fromthepage/pull/1641/files#diff-972315aae52083ccb2e16952bebefa4a3ff2532a51f5646ab327f30dcd472bb1R57
The same thing happens for the "Time Period" filter in my example.
- [ ] 18. For the Curran Index we have sliders for each category of metadata that open when you click on them, so the faceted select screen doesn't take up as much room.  (see the select section here: http://curranindex.org/articles).  We need to do that here, after everything else is working.

Current priority order:  4, 11, 5, 8, 13, 14, 16, 15, 17, 18, 12, 19

benwbrum:
@saracarl will look at 14 to see if it still happens in `development`.

saracarl:
14 does not happen in the `development` branch; it's only in this branch.

diegoviola:
>     * [ ]  17. When I use the date slider I get a `undefined` method `year' for nil:NilClass` on this line: https://github.com/benwbrum/fromthepage/pull/1641/files#diff-972315aae52083ccb2e16952bebefa4a3ff2532a51f5646ab327f30dcd472bb1R57

https://github.com/benwbrum/fromthepage/pull/1641/commits/c6460363cedcb21ff20478d58f77bfae2bc43e7a

diegoviola:
> 14 does not happen in the `development` branch; it's only in this branch.

Fixed by https://github.com/benwbrum/fromthepage/pull/1641/commits/4c23e5e5e3ecc9ddd19dcb5f8425a9d9b9e91a55

diegoviola:
> The `unless params[:search].is_a? String` means that the old search-for-works code takes precedence? I think that's OK.

Yes.

saracarl:
14 is still not fixed, even with this:
https://github.com/benwbrum/fromthepage/pull/1641#issuecomment-715497056

diegoviola:
> Current priority order: 4, 11, 5, 8, 13, 14, 16, 15, 17, 18, 12, 19

@saracarl 19 should be working now.

saracarl:
So testing this, we see that the search parameters are being passed before the date range action.  This means the AJAX never returns and you just see the spinner:

Started GET "/benwbrum/metadata-test?search%5Bs4%5D%5B%5D=Holly+Springs+%28Miss.%29/date_range"

diegoviola:
> So testing this, we see that the search parameters are being passed before the date range action. This means the AJAX never returns and you just see the spinner:
> 
> Started GET "/benwbrum/metadata-test?search%5Bs4%5D%5B%5D=Holly+Springs+%28Miss.%29/date_range"

This should be fixed now with commit: https://github.com/benwbrum/fromthepage/pull/1641/commits/35244839bca383f9bb916ae07c8c626ce4f3b0c0

saracarl:
Current UI of test data:

image

saracarl:
Newest commit resolves 5 and 19, but introduces 20.

diegoviola:
> Newest commit resolves 5, 13 and 19, but introduces 20.

19 have been regressed, the values won't remain at their current values after a request is made.

diegoviola:
> Newest commit resolves 5, 13 and 19, but introduces 20.

19 have been regressed, the values won't remain at their current range after a request is made.

diegoviola:
> Is there a good reason 19 can't work like the CurranIndex does? That would be ideal.:

@saracarl the functionality you are asking for was already implemented and working, see my comment here: https://github.com/benwbrum/fromthepage/pull/1641#issuecomment-719755403

I worked on that 12 days ago and spent many hours to get it to work right, but https://github.com/benwbrum/fromthepage/pull/1641/commits/0202350021f83919b8f330e2299eec83a977ec3f reverted that work.

diegoviola:
> Is there a good reason 19 can't work like the CurranIndex does? That would be ideal.:

@saracarl the functionality you are asking for was already implemented and working, see my comment here: https://github.com/benwbrum/fromthepage/pull/1641#issuecomment-719755403

Commit https://github.com/benwbrum/fromthepage/pull/1641/commits/0202350021f83919b8f330e2299eec83a977ec3f reverted that work.

benwbrum:
Interesting.  I didn't realize that leaving the full-range (rather than the range selected) was the desired behavior.  I'll take a look at matching the curran index's behavior, which is better.

PR #1649, opened 2020-03-10
collection: add confirm alert

diegoviola:
Blank Collection can take a long time on a large collection, let's alert
the user before we trigger it.


diegoviola:
@saracarl I was thinking about changing the message to something more informative, like "This might take a long time on a larger collection, are you sure?".

How do you feel about this, do you have any suggestions?

PR #1651, opened 2020-03-13
Change window size for headless chrome

benwbrum:
This modifies webdriver to run chrome in an HD window.


saracarl:
closing in favor of firefox + selenium.

Issue #1655, opened 2020-03-20
add "private collections you belong to" to the user dashboard


saracarl:
We have a lot of new collections and users coming online, and one of them pointed out that it is really hard to find a private collection.  The transcribers have to create an account, and then follow a link in an email or copy and paste the URL in.

We'd like to add "private collections you belong to" to the user dashboard.

Please add a query for "private collections you are a collaborator on" to the list of collections that show up on this page:
https://fromthepage.com/dashboard/watchlist

diegoviola:
Fixed-by https://github.com/benwbrum/fromthepage/pull/1595.

saracarl:
@diegoviola can you pull this into development as well?  I just had another user ask about it, which makes 2 in less than a week.  Because of so many folks using FromThePage for staff working at home, they are creating and collaborating on a lot of private collections.  Staff can't find them because of this.

diegoviola:
@saracarl I just pushed the changes to development.

saracarl:
@diegoviola I realized last night that #1595 doesn't resolve this.  That issue resolved "co-owner" issues.  We also have "collaborators", who don't have owner privileges, but can see and work on private collections.  Those are the users in question here.  We need collaborators to be able to see the private collections they can work on in their dashboard.

diegoviola:
@saracarl can you please try #1668 ? I'll add tests after you confirm this is working. Thanks.

PR #1668, opened 2020-03-25
dashboard: display private collections

diegoviola:
Fixes: #1655


diegoviola:
Just pushed some specs.

saracarl:
This works, but there is a UI issue that needs to be addressed.  
Screenshot from 2020-03-30 10-45-32

The "You haven't participated in any projects yet. You might want to try transcribing a page of Vita e morte, 1629-1809."  Needs to be removed if you have a private collection you belong to.

benwbrum:
This also has `fromthepage.com`-specific code and needs to be reset or cherry-pick.

PR #1708, opened 2020-04-18
Owner collection contribution time reporting

### LINKED ISSUE #1684: Hourly Report
For all collections owned by a user, create a report:

- user can select a date range
- for each transcriber in that date range, create a row
- for each day in that date range, create a column
- the intersection of each row and column should be the time the transcriber worked that day.  (Hours?  Minutes?  Hours & Minutes?)

The report and date range option should be on the owner dashboard under "summary".  (Please make sure it doesn't show up in the collaborator summary tab, which I think uses the same template.)

If we can do a draft report, we can send it to the client for review.

jackweinbender:
Closes #1684 

This PR adds several features dealing with how we track, store, summarize, and export user activity based on our Ahoy data.

1. **New Table**: The `AhoyActivitySummaries` table provides a place to store daily user activity (in minutes) for each collection that they work on. We store the user, collection, date, time in minutes, and the activity 'type'. This last column is in anticipation of storing data for other kinds of activity. I know. YAGNI. 

2. **Utility Module**: This module is supplied for use with the supplied Rake tasks (below) and the generation of the summary data put into the table. There are really just two methods of note: 1) the core logic for generating a particular User's activity summary for a given date, and 2) a method that takes a list of timestamps (from Ahoy Event logs) and return a length of "active" time, derived from clustering the timestamps together based on a given tolerance. By default, timestamps within 1 hour of each other are considered contiguous.

3. **Rake Tasks**: I've included three Rake tasks: 
    * An ad hoc task to populate the AhoyActivitySummary table based on all the Ahoy data that we already have. This method looks up what day the first Ahoy event is recorded on and processes AhoyActivitySummaries for each day and user crawling backwards from the present. When I ran this job, it took my computer 1-2 hours. The job intentionally takes small batches of Ahoy Events to prevent spikes in memory usage, since this is a long-running job.
    * A daily/nightly task that processes the summary data for the previous day
    * A utility task that will run/re-run the nightly job for the past N days. 
    All of these tasks are non-destructive (they don't delete Ahoy Events after using them) and idempotent (rerunning the job will not produce new records or duplicates).

3. **New Owner UI**: Using the new data supplied by the ahoyActivitySummary table, this PR adds allows owners to view a contributor **activity summary**, given a particular range of dates on their Owner `dashboard#summary` page. The data displayed shows all active users for the specified time frame and the amount of time in minutes that they were "active" (as defined by us).

4. **Export Data**: Additionally, we've included a button to e**xport more detailed data as a CSV;** this feature provides a breakdown of each active user's activity by day within the specified range.

image


lib/tasks/ahoy_activity_rollup.rake
```diff
@@ -0,0 +1,79 @@
...
+      date = n.days.ago
+
+      print "\n-- Ahoy Rollup #{date.strftime("%Y-%m-%d")} --\n"
+
+      # Perform the rollup for the day
+      rollup_transcribe_for_date(date)
+    end
+  end
+
+  def rollup_transcribe_for_date(day=1.day.ago)
```
benwbrum, line 26: I generally like to move business logic (even for recurring tasks) out of the body of rake tasks and into either static methods on a model or onto a separate library.   That lets the rake task focus on argument parsing and user notification, and encapsulates the logic to a place where it can be reused by the application or different rake tasks.  See `lib/contentdm_translator.rb` and its relationship to the two `lib/tasks/cdm_*` tasks for an example.

lib/tasks/ahoy_activity_rollup.rake
```diff
@@ -0,0 +1,79 @@
...
+                  begin 
+                    activity = AhoyActivitySummary.create({
+                      date: day.beginning_of_day,
+                      user_id: user,
+                      collection_id: cid,
+                      activity: 'transcribe',
+                      minutes: minutes
+                    })
+                  rescue ActiveRecord::RecordNotUnique => e
+                    puts e
```
benwbrum, line 70: I take it that this serves to avoid over-writing older data?  Is there another reason it's here that I'm missing?

lib/tasks/ahoy_activity_rollup.rake
saracarl, line 26: Change this to a parameter we can pass to the task.

lib/tasks/ahoy_activity_rollup.rake
```diff
@@ -0,0 +1,79 @@
+namespace :fromthepage do
+  desc "Daily Task to Rollup Ahoy Events into Transcribe minutes by Date, User, and Collection"
```
saracarl, line 2: Because we're only going to be running this task once a day, we need some slightly helpful indication in the UI that you can't include today's date.  

The defaults include today's date and probably shouldn't.

db/migrate/20200418194708_create_ahoy_activity_summaries.rb
```diff
@@ -0,0 +1,16 @@
...
+    # drop_table :ahoy_activity_summaries
+    create_table :ahoy_activity_summaries do |t|
+      t.datetime  :date
+      t.integer   :user_id
+      t.integer   :collection_id
+      t.string    :activity
+      t.integer   :minutes
+      t.timestamps
+    end
+    add_index :ahoy_activity_summaries, [:activity, :date, :collection_id, :user_id],
```
saracarl, line 12: Activity should be last here.

lib/tasks/ahoy_activity_rollup.rake
jackweinbender: We talked about this, but for the sake of the written record, yes, this is because we don't overwrite these data. Should we want to re-run this job, we simply delete  the bad records and re-run the job.

db/migrate/20200418194708_create_ahoy_activity_summaries.rb
jackweinbender: Done.

lib/tasks/ahoy_activity_rollup.rake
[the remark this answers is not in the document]
jackweinbender: I've abstracted the "crawls back to" date to look up the first Ahoy Event in the DB. I've also added a helper rake task that allows us to run this task "For the last N days" in case we miss a day or need to rerun the job. I thought this was better than trying to parse datestring parameters.

This ignores the case for a "date range," but it is easy enough to implement should we need it. I think these two main functions should get us 99% of the way there, though.

jackweinbender: Done.

lib/tasks/ahoy_activity_rollup.rake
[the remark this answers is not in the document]
jackweinbender: Dupe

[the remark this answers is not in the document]
jackweinbender: Dupe.

jackweinbender: This has been updated and at least partially addressed. There are some complications re: localization, but we're close.

jackweinbender:
I'd like a little guidance on the UI and what you'd like it to look like. Are there tables that I should try to match?

benwbrum:
I've just tested this using `2020_04_29_08_00_01.dmp`, running the `summarize_all_ahoy_activity` task, logging in as Wisconsin Historical Society and looking for data across all of April.

Issues:
* The rake task takes a _really_ long time to run.  I was able to process a months worth of data in about an hour, and leaving it running for three hours only got us back to February.  
* Although minutes appear for the summary web page, the spreadsheet contains 0 in each cell.
Screenshot from 2020-05-03 14-29-13
Screenshot from 2020-05-03 14-40-03

Other notes from testing:
* "There is no activity for this date range." should be changed to "There is no activity data for this time period."

lib/ahoy_activity_utils.rb
```diff
@@ -0,0 +1,67 @@
+module AhoyActivityUtils
+    def self.rollup_transcribe_for_date(day=1.day.ago)
+        # Active users are users who log "transcription-type" deeds on a given day.
+        # We only want to track these people for now
+        active_users = Deed
+        .where(deed_type: DeedType.transcriptions)
+        .where("created_at BETWEEN ? AND ?", day.beginning_of_day, day.end_of_day)
+        .distinct.pluck(:user_id)
```
benwbrum, line 8: Is there a chance this `distinct` is what's slowing us down?  At times, testing the uniqueness in SQL is really slow if there is no index (in which case it's blazingly fast).

lib/ahoy_activity_utils.rb
```diff
@@ -0,0 +1,67 @@
...
+        u = User.find_by(id: user) 
+        puts "\t#{u&.login || '[Deleted User]'}"
+
+        # We Define transcribe events as any even on the transcribe controller
+        # This could be expanded or resctricted
+        events = Ahoy::Event
+            .where(user_id: user)
+            .where("name LIKE 'transcribe#%'")
+            .where("time BETWEEN ? AND ?", day.beginning_of_day, day.end_of_day)
+            .select(:id, :user_id, :time, :properties)
```
benwbrum, line 24: I'd also be tempted to pull the generated SQL out of the logs and run `EXPLAIN #{sql statement}` in MySQL here.

benwbrum:
It looks like the problem is with the `to_time` conversion:

```
[17] pry(#<DashboardController>)> activity
=> {
    2020-04-09 00:00:00 UTC => 551,
    2020-04-10 00:00:00 UTC => 720,
    2020-04-11 00:00:00 UTC => 184,
    2020-04-16 00:00:00 UTC => 161,
    2020-04-17 00:00:00 UTC => 286
}
[18] pry(#<DashboardController>)> april_9_in_dates = dates.map { |d| d.to_time }[7]
=> 2020-04-09 00:00:00 -0500
[19] pry(#<DashboardController>)> activity.keys[0] == april_9_in_dates
=> false
```

It's possible that this works on other machines which are set to UTC, but on my box (CDT) fails.

jackweinbender:
Ahh! Of course! That makes perfect sense.

jackweinbender:
I've made two changes:

1. I've fixed the timezone issue by using the Date class, rather than the Time class. I was able to reproduce the reported error by changing the timezone on my docker container, then confirmed that my change, in fact, fixed the problem.

2. Regarding the speed of the Rake task; I realized that I included a small Db query to look up each user in the inner loop of each day. Removing this query made a significant difference on my local machine---I estimate about 30-40% speed increase. 

I did try a couple other things (batching/grouping queries, etc.), but frankly, none of them seemed to make any difference or were a bit slower. None of the current set of queries were particularly slow, though, perhaps the `LIKE` is slower than we'd like. I don't know of a way to get around that.

benwbrum:
This looks good.  We can merge after testing.

Issue #1627, opened 2020-02-19
Subject export hardcodes 'http://' in links


bencomp:
The subject export controller hardcodes 'http://' links, instead of using URL helper functions that should automatically use the configured scheme.

diegoviola:
@bencomp @saracarl can you please provide some steps to reproduce for this issue? I looked at this tonight and couldn't reproduce it. Thanks in advance.

bencomp:
Sure, @diegoviola. On the export page of a collection, near the bottom, there is a button to export the Subjects as CSV. When you use that to download a CSV file, I notice that all the URLs in the file start with `http://`, even though our FromThePage is configured to only use HTTPS.

The issue is in https://github.com/benwbrum/fromthepage/blob/35caa2a2eb82bbe4f4aff0d92c3290ea8246b42f/lib/subject_exporter.rb#L21 and https://github.com/benwbrum/fromthepage/blob/35caa2a2eb82bbe4f4aff0d92c3290ea8246b42f/lib/subject_exporter.rb#L30 – though the latter may be a Rails configuration issue. I think Rails 4 needs extra configuration to tell it that we are using HTTPS (which is handled by a proxy server).

It's not the biggest issue, of course, but hopefully it's easy to fix.

diegoviola:
@bencomp can you please try https://github.com/benwbrum/fromthepage/pull/1719?

bencomp:
I'm afraid I can't easily test any fixes, as we don't have a test environment 👼 
I did comment on your PR. Thanks for your work!

Issue #1460, opened 2019-08-21
Activity links link to fromthepage.com addresses rather than relative paths


jackweinbender:
I noticed that the links to deeds on both the user profile ("recent activity") and dashboard (owner dash and collaborator dash under "your activity") link to the fromthepage.com host in my local environment. I'm guessing this has to do with a change I made a while back to use the `_url` helper methods (rather than `_path`) in the deed partial. We did this so that links in the mailers (which use these partials) would still work.

I don't *think* this is a serious problem, since local installs should configure their hostname in such a way that the right URLs will be generated. That said, I'm pretty sure the deed partial has some logic in it that is supposed to prevent this, so, we should try to figure it out.

atomotic:
same behaviour for me
i've deployed fromthepage using this [docker](https://github.com/ryanfb/docker_fromthepage) behind nginx

these links are rendered with only_path="true"
is **app/views/deed/_deed.html.slim** the template to look?


```<a href="http://localhost:3000/andrea" only_path="true">andrea</a> transcribed
page
<a
  href="http://localhost:3000/raffaele/biblioteca-digitale-estense/trentasei-lettere-inedite-di-l-a-muratori/transcribe/6"
  only_path="true"
  >0006 - A.M.48.25</a
>
in
<a
  href="http://localhost:3000/raffaele/biblioteca-digitale-estense"
  only_path="true"
  >biblioteca digitale estense</a
>
collection
```

bencomp:
I agree that installs should configure the hostname in the configuration. I guess @jackweinbender did not change the hostname?

I also see what @atomotic sees: `only_path="true"` in HTML elements. I don't understand most of Rails' magic, but I think <https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to> may provide hints, e.g. that the `only_path: !mailer` part of the `link_to` call should move inside the `..._url()` call.

A different issue that I see is that HTTP URLs are generated instead of HTTPS. I changed the `config.default_url_options` already to include `protocol: "https"` (as to me it looks like the automatic use of https when `config.force_ssl = true` only becomes available in Rails 5, whereas we currently use Rails 4.1.2), but it doesn't help.

bencomp:
I'm learning a little about Rails routing magic :)

[...]

PR #1719, opened 2020-04-24
export: fix export subjects url

diegoviola:
by passing the request.base_url to our export method, rather than
hardcode http in the URL.

Fixes: #1627

### DIFF
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -63,7 +63,7 @@ class ExportController < ApplicationController
   end
 
   def subject_csv
-    send_data(@collection.export_subjects_as_csv(base_url: request.base_url),
+    send_data(@collection.export_subjects_as_csv,
               :filename => "fromthepage_subjects_export_#{@collection.id}_#{Time.now.utc.iso8601}.csv",
               :type => "application/csv")
     cookies['download_finished'] = 'true'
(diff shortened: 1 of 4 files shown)


bencomp:
Interesting approach. I am not a Rails developer, but my impression has been that URLs for resources should be generated using `url_for()` or the URL helper `*_url()`. See https://stackoverflow.com/questions/11939865/named-routes-path-vs-url and #1460 for another issue involving URL helpers.

This change appears to only fix one of two links, the page URL. The URL of the article itself may need a different fix (it is likely that the Rails upgrade helps).

diegoviola:
@bencomp you're right, see commit https://github.com/benwbrum/fromthepage/pull/1719/commits/8c0669c3ac0cec99561f91a5c75455310bafae6a. Thanks for the suggestion. :+1:

bencomp:
I think we should looked at existing uses of `url_for` in the codebase 😊 
In the export controller I just found `page_url=url_for({:controller=>'display',:action => 'display_page', :page_id => page.id, :only_path => false})` which is I believe the exact call that we need.

So no need for an additional route or a nested `_path` within `url_for`.

diegoviola:
@bencomp I just pushed a commit with your suggestion, thanks again.

PR #1745, opened 2020-05-19
1739 i18n

diegoviola:


diegoviola:
@JoshuaGOB can you please add the missing strings for the dashboard locale?

https://github.com/benwbrum/fromthepage/blob/1739-i18n/config/locales/dashboard/dashboard-en.yml

The missing strings are in the Spanish and Portuguese versions, see:

https://github.com/benwbrum/fromthepage/blob/1739-i18n/config/locales/dashboard/dashboard-es.yml

https://github.com/benwbrum/fromthepage/blob/1739-i18n/config/locales/dashboard/dashboard-pt.yml

Thanks!

PR #1760, opened 2020-06-06
Fixed #1758 by reading FTS field from CDM REST API

### LINKED ISSUE #1758: Use FTS column for OCR ingestion when importing from CONTENTdm
Currently, when importing works from CONTENTdm, we assume that the OCR will be stored in a field with the CDM nickname `transcr` or `full`.  We should instead look for the field marked Full Text Search, as we do when syncing to CDM.

Currently we fetch the FTS field using the Catcher SOAP API, which requires a log-in.  According to CDM, there is a REST API requiring no login that will give the same information:
https://sandbox.contentdm.oclc.org/digital/bl/dmwebservices/index.php?q=dmGetCollectionFieldInfo/p10010coll1/json

Let's switch this code to use the right API.

benwbrum:


lib/contentdm_translator.rb
```diff
@@ -67,14 +80,11 @@ def self.metadata_from_cdm_info(info)
...
 
-  def self.ocr_from_cdm_info(info)
-    # Hard-coding field names is an ugly hack, but the FTS configuration
-    # is only accessible from the SOAP API, not the REST API and we don't
-    # want to prompt for username/password/license key at this point in
-    # the work flow
-    transcript = info['transc'] || info['full']
+  def self.ocr_from_cdm_info(info, fts_field)
+    transcript = info[fts_field]
+    binding.pry
```
saracarl, line 85: You left a binding.pry here.

PR #1762, opened 2020-06-07
Update article_category route

jcreedon:
This fixes an issue I was running into running on the rails6 branch. I'm not sure if it is the correct fix but it works as far as I can tell. Previous to this fix, adding a category to a subject would fail with: ActionController::RoutingError (No route matches [POST] "/article/article_category"):


saracarl:
Thanks!

PR #1775, opened 2020-06-14
1761 upload subjects

benwbrum:


app/controllers/article_controller.rb
```diff
@@ -185,6 +186,57 @@ def show
...
+      flash[:error] = "CSV file must contain headers for HEADING, ARTICLE, URI, and CATEGORY"
+      redirect_to article_upload_form_path(@collection)    
+    end
+  end
+
+
+  def upload_example
+    example = File.read(File.join(Rails.root, 'app', 'views', 'static', 'subject_example.csv'))
+    send_data example, filename: "subject_example.csv"
+#    render :file => 'static/subject_example.csv', :layout => false, :content_type => "text/csv"
```
saracarl, line 237: Let's clean up this commented out line.

app/views/article/upload_form.html.slim
```diff
@@ -0,0 +1,16 @@
+=javascript_include_tag 'application'
+.litebox-embed(style="width:550px")
+  h1 Upload Subjects
+  p To create subjects from an external authority file, upload a spreadsheet with one row per subject. See #{link_to "link", article_upload_example_path, {target: "_blank"}} for an example. (Note that multi-line/article cannot be updated.)
```
saracarl, line 4: You didn't include upload_example/subject_example.csv in this PR.

saracarl: Does the spreadsheet support comments?  If it does, can we move this to it?:
"(Note that multi-line/article cannot be updated.)" 
If not, we should document it in documentation.  I wouldn't include it here.

saracarl: Instead of "See #{link_to "link", article_upload_example_path, {target: "_blank"}} for an example."  Just link the "spreadsheet with one row per subject" words.  That will mirror what we did for metadata.yml.

app/views/transcribe/josie.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Hi I am not Josie</h1>
+
+Here is a new line
+
```
saracarl, line 4: Let's remove this file.

config/routes.rb
```diff
@@ -72,6 +72,9 @@
   put   'article/article_category', :to => 'article#article_category'
   patch 'category/update', :to => 'category#update'
   patch 'user/update', :to => 'user#update'
+  get   ':collection_id/article/upload', :to => 'article#upload_form', :as => 'article_upload_form'
+  post  'article/upload', :to => 'article#subject_upload', :as => 'article_subject_upload'
+  get   '/article/upload_example', :to => 'article#upload_example', :as => 'article_upload_example'
```
saracarl, line 77: The equivalent metadata.yml doesn't seem to have a route, and just goes to static/metadata.yml.  (I don't care that much, but it seems simpler.)

benwbrum: I think we're going to have to have specific routes for everything; `rails6` seems to expect them.

app/views/article/upload_form.html.slim
benwbrum: Looks like our `.gitignore` prevented me from adding the example csv.

app/views/static/subject_example.csv
```diff
@@ -0,0 +1,4 @@
+HEADING,URI,ARTICLE,CATEGORY
+Sally Joseph Carr Brumfield,P001001,Daughter of Rufus Bascom Carr and Fannie Edmonia Hobson,People
```
saracarl, line 2: So P001001 breaks the UI a bit, I don't think we have to address it here but I opened issue #1778 to address it later.

app/views/article/upload_form.html.slim
```diff
@@ -0,0 +1,16 @@
+=javascript_include_tag 'application'
+.litebox-embed(style="width:550px")
+  h1 Upload Subjects
+  p To create subjects from an external authority file, upload a CSV file with one row per subject.  (See #{link_to "example file", article_upload_example_path, {target: "_blank"}}.)
```
saracarl, line 4: Instead of "See #{link_to "link", article_upload_example_path, {target: "_blank"}} for an example." Just link the "spreadsheet with one row per subject" words. That will mirror what we did for metadata.yml.

benwbrum:
Obviated by #1787.

PR #1784, opened 2020-06-20
Update user_params filter to use real_name

jcreedon:
This fixes a bug where neither users nor admins could update the real name
of a user profile. The `user_params` permit list had `:display_name`, however,
per the model that is an auto-generated field and `:real_name` is the one
used in the form.

This has been tested and verified to work correctly now.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -232,6 +232,6 @@ class AdminController < ApplicationController
   private
 
   def user_params
-    params.require(:user).permit(:display_name, :real_name, :login, :email, :account_type, :start_date, :paid_date, :user, :owner)
+    params.require(:user).permit(:real_name, :login, :email, :account_type, :start_date, :paid_date, :user, :owner)
   end
 end
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -98,7 +98,7 @@ class UserController < ApplicationController
   private
 
   def user_params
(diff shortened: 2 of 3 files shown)


app/controllers/user_controller.rb
```diff
@@ -98,7 +98,7 @@ def record_deed
   private
 
   def user_params
-    params.require(:user).permit(:display_name, :orcid, :slug, :website, :location, :about, notifications: [:user_activity, :owner_stats, :add_as_collaborator, :add_as_owner, :note_added])
```
benwbrum, line 101: Looks like this does the trick!

app/controllers/admin_controller.rb
```diff
@@ -232,6 +232,6 @@ def downgrade
   private
 
   def user_params
-    params.require(:user).permit(:display_name, :real_name, :login, :email, :account_type, :start_date, :paid_date, :user, :owner)
```
benwbrum, line 235: Having two `:real_name` parameters to `permit` seems a bit off here.  I also wonder if we're actually showing a `display_name` field to admins; if so--and if it'll be obliterated by the callbacks--we should probably remove it from the admin UI.

jcreedon: Yep, I should have looked closer, that is totally what is happening, and why it wasn't working for me earlier. I'll submit another commit to address that as well.

jcreedon:
I have addressed the changes requested and added in a little cleanup of the admin side as suggested.

I am totally fine if you want to mention me in the release notes.

benwbrum:
TODO: merge into into `development` branch.

PR #1793, opened 2020-06-22
1645 redirect to owner tab

isabelabarton:
when the admin edits an owner they are redirected to the owner tab instead of the user tab


saracarl:
Great!  This is going to save me a lot of annoyance!

PR #1814, opened 2020-06-25
translation of views

JoshuaGOB:
First pass Spanish only


benwbrum:
To test: error messages from bad XML in transcriptions and translations.

PR #1827, opened 2020-07-07
Finished adding footers to work tabs

JosieBrum:


app/views/display/list_pages.html.slim
```diff
@@ -2,7 +2,7 @@
 
 -if @work.pages.empty?
   .nodata
-    h4.nodata_title No pages found
+    h4.nodata_title No pages found, you lose.
```
saracarl, line 5: Please change this back to the original text.

benwbrum, line 5: This looks like my fault.

PR #1830, opened 2020-07-08
Fixes #1617 by handling real_name/display_name in TEI

### LINKED ISSUE #1617: TEI export responsibility statements
Responsibility statements in the TEI export should include all the contributors with their real_name, if it exists. @benwbrum needs to review the TEI exports because they have entity IDs for contributors, but not responsibility statements. See https://fromthepage.com/export/tei?work_id=14647 @benwbrum please add the responsibility statements to this issue. (And we're going to need this for CWGM.)

benwbrum:


app/controllers/export_controller.rb
```diff
@@ -33,6 +33,8 @@ def tei
     @user_contributions =
       User.find_by_sql("SELECT  user_id user_id,
                                 users.real_name real_name,
+                                users.display_name display_name,
+                                users.guest guest,
```
saracarl, line 37: Why do we have guests here?  I suspect that would leave us with some "guest" IDs in the attributions, which I'm not sure we want.

PR #1841, opened 2020-07-13
owner content will not be flagged for abuse

isabelabarton:
Users' edits will still be redirected to the abuse tab if the qualify, but owners' will not.


benwbrum:
This is failing in tests, probably because we need to add `.owner` to the page versions in the fixtures or mocks.

benwbrum:
@diegoviola says he'll fix the tests for this one.

PR #1849, opened 2020-07-20
1804 other user alert fix

JosieBrum:


JosieBrum:
This seems to include all of the files from the pull request I put in about longer flash messages as well as changing the one line. I am going to try to change that so the changes are not repetitive when they are merged.

saracarl:
@JosieBrum let's talk about this on the call today.

PR #1860, opened 2020-07-21
preview and autolink button moved to above and below the text box

isabelabarton:


isabelabarton:
Screen Shot 2020-07-21 at 3 54 42 PM
Screen Shot 2020-07-21 at 3 54 48 PM

This is what it looks like for me but the checks are failing

PR #1870, opened 2020-07-23
autosplit mention removed from upload screen

isabelabarton:


app/views/collection/upload.html.slim
```diff
@@ -8,5 +8,5 @@
     .fglight
       p PNG, GIF, and JPG files are all acceptable.
       p Images should be oriented so that they are right-side-up.
-      p Images should be split down the spine, so that only one page is visible on an image.<br> (See the <a href="https://github.com/benwbrum/autosplit">autosplit</a> tool if you need to split two-page images.)
```
saracarl, line 11: Let's actually remove this whole paragraph, starting with the
```p Images should be split down the spine, so that only one page is visible on an image.```

File: app/views/dashboard/_new_work.html.slim [hunk 2]
```diff
@@ -66,5 +66,5 @@
     .fglight
       p PNG, GIF, and JPG files are all acceptable.
       p Images should be oriented so that they are right-side-up.
-      p Images should be split down the spine, so that only one page is visible on an image.<br> (See the <a href="https://github.com/benwbrum/autosplit">autosplit</a> tool if you need to split two-page images.)
```
saracarl, line 69: Same here.

app/views/dashboard/upload.html.slim
```diff
@@ -10,5 +10,5 @@
     .fglight
       p PNG, GIF, and JPG files are all acceptable.
       p Images should be oriented so that they are right-side-up.
-      p Images should be split down the spine, so that only one page is visible on an image.<br> (See the <a href="https://github.com/benwbrum/autosplit">autosplit</a> tool if you need to split two-page images.)
```
saracarl, line 13: and here.  ;)

isabelabarton: i think i fixed all of them

PR #1907, opened 2020-07-29
render_anywhere gem removed

isabelabarton:


diegoviola:
render_anywhere was removed on the rails6 branch, development no longer uses this gem.

Issue #1890, opened 2020-07-29
Lost owner in profile pages  [regression]


saracarl:
If you go to a project owner profile page, the "Owned by" is blank (for at least some users, but probably all)

For instance:
https://fromthepage.com/adah  

This user has a name in their profile.

PR #1912, opened 2020-07-30
dashboard: fix regression due to missing locale

diegoviola:
Fixes: #1890


saracarl:
@diegoviola this doesn't address the issue.  (The issue has a bad subject -- all stack traces now start with a locale message, which misled me the first couple of times I saw it), but please click through the link provided in the issue and view the owned by in the UI.

diegoviola:
> the owner name string is missing, not the "owned by".

Works for me. Please re-test.

PR #1935, opened 2020-08-03
1854 inclusive language "blacklist" to "denylist"

isabelabarton:


saracarl (changes requested):
@diegoviola Could you please rename the view from flag_blacklist to flag_denylist?  That will get rid of the last instance of blacklist in the code.

PR #1937, opened 2020-08-04
Replace non-XML 1.0 characters with space

### LINKED ISSUE #1875: Form feed character causes XML processing error; saving transcription fails
One of our users quite consistently tries to save transcriptions with a [form feed character][ff] in them. This causes the transcription to not be saved and errors to be produced.

[ff]: https://en.wikipedia.org/wiki/Page_break#Form_feed

### Start of request
```
I, [2020-07-25T08:14:50.259046 #8333]  INFO -- : Started PATCH "/transcribe/save_transcription" for 86.93.153.185 at 2020-07-25 08:14:50 +0000
I, [2020-07-25T08:14:50.260513 #8333]  INFO -- : Processing by TranscribeController#save_transcription as HTML
I, [2020-07-25T08:14:50.260936 #8333]  INFO -- :   Parameters: {"utf8"=>"✓", "authenticity_token"=>"I5Ju4956DSQfFWvAgU/o0495mJvfjIj89BK4Z5/oT7k=", "page_id"=>"4169", "page"=>{"mark_blank"=>"0", "needs_review"=>"0", "source_text"=>"\fIl
 tappeto
```

### Start of raw text
```
TRANSCRIPTION   Source Text:
BEGIN_SOURCE_TEXT
^LIl tappeto
```

### Start of stacktrace
```
E, [2020-07-25T08:14:50.292635 #8333] ERROR -- : TRANSCRIPTION  2020-07-25 08:14:50 +0000       ERROR   EXCEPTION       #<RuntimeError: Illegal character "\f" in raw string "^LIl tappeto">
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/text.rb:139:in `block in check'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/text.rb:135:in `each'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/text.rb:135:in `check'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/text.rb:121:in `initialize'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/parsers/treeparser.rb:47:in `new'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/parsers/treeparser.rb:47:in `parse'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/document.rb:288:in `build'
/opt/rh/rh-ruby23/root/usr/share/ruby/rexml/document.rb:45:in `initialize'
/opt/fromthepage/app/models/xml_source_processor.rb:343:in `new'
/opt/fromthepage/app/models/xml_source_processor.rb:343:in `update_links_and_xml'
/opt/fromthepage/app/models/xml_source_processor.rb:114:in `wiki_to_xml'
/opt/fromthepage/app/models/xml_source_processor.rb:86:in `process_source'
/home/ubgit/.gem/ruby/gems/activesupport-4.1.2/lib/active_support/callbacks.rb:424:in `block in make_lambda'
```

## Analysis
I do not know for sure how this user gets the form feed character into the text, though I assume that they copy and paste the transcription from software that automatically inserts the character at the start of the text. I would like to tell them to clean the text before submitting, but the character is invisible. I think FromThePage should be more accepting of users and more strict in its handling of user-submitted content.

FromThePage internally (and when exporting) uses XML 1.0 to store transcriptions and translations. XML 1.0 doesn't allow control characters at all (XML 1.1 allows most control characters when they are character-encoded), but FromThePage doesn't check for or filter such characters before passing the constructed XML string to the XML processing library.
The method `valid_xml_from_source(source)` does convert `&` to `&amp;`, but should in my opinion also remove control characters (and perhaps do some more checking) before saying the return value is valid XML.

I will create a PR to at least remove the form feed character.

bencomp:
This fixes #1875 (and potential other bugs caused by illegal characters) by replacing characters that are outside the [valid range for XML 1.0](https://www.w3.org/TR/xml/#charsets) with a space. I chose to replace with space rather than remove because someone might use some characters as spacing – and it could signal that there used to be an invisible character where there are unexpected spaces.

I ran a simple test to see that form feeds are replaced, as well as `NULL` characters, using:

```ruby
puts "blabla\fdljlkdf\tcheck\0".gsub! /[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u10000-\u10FFFF]/, 'REPLACED'
#=> "blablaREPLACEDdljlkdf   checkREPLACED"
```

I didn't dare to touch the test spec `spec/models/xml_source_processor_spec.rb`, as I don't really know how to run it, though that is where I would add a new source string and expected XML return string.
However, in our production environment, I did add a similar line:

```ruby
safe.gsub! /\f/, ''
```

and the user who ran into the issue reported that the errors dissappeared. This line of course doesn't check for other illegal characters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,9 +44,7 @@ script:
 
 branches:
   only:
+    - main
     - master
-    - ui-design
     - development
     - fromthepage.com
-    - indianapolis
-    - rails6
--- a/app/models/xml_source_processor.rb
+++ b/app/models/xml_source_processor.rb
@@ -323,6 +323,7 @@ module XmlSourceProcessor
     source = source || ""
     safe = source.gsub /\&/, '&amp;'
     safe.gsub! /\&amp;amp;/, '&amp;'
+    safe.gsub! /[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u10000-\u10FFFF]/, ' '
 
     string = <<EOF
     <?xml version="1.0" encoding="UTF-8"?>
--- a/spec/features/collection_spec.rb
+++ b/spec/features/collection_spec.rb
@@ -177,7 +177,7 @@ describe "collection settings js tasks", :order => :defined do
 
     it "transcribing doesn't work for inactive collections" do
       visit collection_display_page_path(@collection.owner, @collection, @page.work, @page.id)
-      expect(page).not_to have_link('Transcribe')
+      #expect(page).not_to have_link('Transcribe') #for FromThePage there is a "Sign Up to Transcribe" link
     end
 
     it "toggles collection active" do
--- a/spec/features/collection_spec.rb
+++ b/spec/features/collection_spec.rb
@@ -177,7 +177,7 @@ describe "collection settings js tasks", :order => :defined do
 
     it "transcribing doesn't work for inactive collections" do
       visit collection_display_page_path(@collection.owner, @collection, @page.work, @page.id)
-      #expect(page).not_to have_link('Transcribe') #for FromThePage there is a "Sign Up to Transcribe" link
+      expect(page).not_to have_link('Transcribe')
     end
 
     it "toggles collection active" do
--- a/spec/models/xml_source_processor_spec.rb
+++ b/spec/models/xml_source_processor_spec.rb
@@ -15,7 +15,7 @@ EXPECTED_XML_DISABLED = <<EOF
       </page>
 EOF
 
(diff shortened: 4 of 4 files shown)


bencomp:
So I did add a test, but I think the control characters used in the test fixture may have caused GitHub to refuse the commit over HTTPS. I then updated the test in the GitHub editor and again in my local repository and now the branch history is a bit messy.

benwbrum:
Brilliant!  I was working on a fix, but this is far more comprehensive.  I am editing the PR and changing to base branch to `benwbrum:development`.  (We need to document this in `contributing.md`.)

Issue #1972, opened 2020-08-13
Some domains missing on IIIF exports [regression]


benwbrum:
After the upgrade, some IIIF endpoints are missing domain names, and showing relative URLs instead.
See https://fromthepage.com/iiif/12/manifest for an example.

The missing elements seem to be
* Related (manifest)
  * Read work
  * Table of Contents
* seeAlso
  * Verbatim Plaintext
  * Emended Plaintext
  * Searchable Plaintext
* Sequence Rendering
  * Verbatim Plaintext
* Canvas seeAlso
  * all plaintext links

PR #2021, opened 2020-08-24
iiif: replace _path with _url for endpoints

diegoviola:
Fixes: #1972


benwbrum:
I've tested this and it works.

PR #2041, opened 2020-08-31
Added seperation of name and time in notes

JosieBrum:
Looks like this:
Screenshot (11)
If it is not okay, I can figure something else out.

### DIFF
--- a/app/views/notes/_note.html.slim
+++ b/app/views/notes/_note.html.slim
@@ -14,6 +14,7 @@
           =svg_symbol '#icon-remove-sign', class: 'icon', title: t('.delete')
     =time_tag note.created_at
       =link_to note.user.display_name, user_profile_path(:user_id => note.user.id)
+      p 
       =time_ago_in_words(note.created_at)
     .user-bubble_note ==simple_format(sanitize(note.body, tags: %w(strong b em i a), attributes: %w(href)))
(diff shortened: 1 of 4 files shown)


benwbrum:
@diegoviola said he'd fix the tests for this.

app/views/notes/_note.html.slim
```diff
@@ -14,9 +14,8 @@
           =svg_symbol '#icon-remove-sign', class: 'icon', title: t('.delete')
     =time_tag note.created_at
       =link_to note.user.display_name, user_profile_path(:user_id => note.user.id)
-      =" "
+      |&nbsp;
       =time_ago_in_words(note.created_at)
-      =" ago"
```
saracarl: We seem to have lost the "ago".  Can we add it back in, please?  Translated, if possible.  (Or maybe it's blank in other languages, but it only makes sense in with the "ago" in English.)

saracarl (changes requested):

benwbrum:
As @saracarl said, we want the display to say "about 3 years ago" instead of "about 3 years".  We'd like to add `" ago"` and the equivalents in Spanish and Portuguese.

If this requires a new message in the dictionary, we can add a comment saying `# needs translation`

PR #2075, opened 2020-09-21
Improve document set experience

benwbrum:


benwbrum:
Interesting.  Let's think about this in #3298

bencomp:
When I login as this other user, I see the document set in Projects, but cannot access it. I feel this could be an issue of its own, although it is very useful to know that my (admin) account should probably not be a staff account.

Issue #1993, opened 2020-08-19
opt-out for project owner collaborators list


saracarl:
You can opt-out of emails when you sign up for an account, but there's no way to opt-out in your profile.  (The profile just has unsubscribes for specific emails, but not for the whole list of contributors that project owners use for newsletters -- that's in the UI and as a spreadsheet export.)

saracarl:
The user needs to be able to opt-out and be removed from what the owner sees in two places:

![image](https://user-images.githubusercontent.com/708566/95255827-92671100-07e7-11eb-8dcc-59ba8d10ade7.png)
![image](https://user-images.githubusercontent.com/708566/95255893-ae6ab280-07e7-11eb-8c32-6f4acbe9d86f.png)

saracarl:
The goal is to give transcribers an option to opt-out of collection emails in their user profile, then if they have opted-out remove just their email address from *both* the list of collaborators in that view and in the owner dashboard spreadsheet download.

PR #2089, opened 2020-10-06
collection: filter collaborator emails

diegoviola:
Fixes: #1993

### DIFF
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -337,7 +337,7 @@ class CollectionController < ApplicationController
         time_proportional = (@user_time_proportional[user.id] / 60 + 1).floor
       end
 
-      user.email = "opted out" unless user.notification.user_activity?
+      user.email = "opted out" unless user.notification.owner_contact?
 
       id_data = [user.display_name, user.email]
       time_data = [time_total, time_proportional]
@@ -399,7 +399,7 @@ class CollectionController < ApplicationController
     note = ''
     note += d.note.title if d.deed_type == DeedType::NOTE_ADDED && !d.note.nil?
 
-    d.user.email = "opt out" unless d.user.notification.user_activity?
+    d.user.email = "opt out" unless d.user.notification.owner_contact?
 
       record = [
         d.created_at,
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -399,6 +399,8 @@ class CollectionController < ApplicationController
     note = ''
     note += d.note.title if d.deed_type == DeedType::NOTE_ADDED && !d.note.nil?
 
+    d.user.email = "opt out" unless d.user.notification.user_activity?
+
       record = [
         d.created_at,
         d.user.display_name,
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -326,41 +326,37 @@ class CollectionController < ApplicationController
     ]
 
     stats = @active_transcribers.map do |user|
-      if user.notification.user_activity?
-        time_total = 0
-        time_proportional = 0
+      time_total = 0
+      time_proportional = 0
 
-        if @user_time[user.id]
-          time_total = (@user_time[user.id] / 60 + 1).floor
-        end
-
-        if @user_time_proportional[user.id]
-          time_proportional = (@user_time_proportional[user.id] / 60 + 1).floor
-        end
-
-        id_data = [user.display_name, user.email]
-        time_data = [time_total, time_proportional]
-
-        user_deeds = @collection_deeds.select { |d| d.user_id == user.id }
-
-        user_stats = [
-          user_deeds.count { |d| d.deed_type == DeedType::PAGE_TRANSCRIPTION },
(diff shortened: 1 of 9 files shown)


diegoviola:
This will disable the "All Collaborator Emails" part of the Collaborators tab when toggling on the "Nightly emails with owned collection activity" of a user profile.

saracarl:
This does not accomplish the goal of the issue which is to give transcribers an option to opt-out of collection emails in their user profile, then if they have opted-out remove just their email address from *both* the list of collaborators in that view and in the owner dashboard spreadsheet download.

app/controllers/collection_controller.rb
```diff
@@ -326,37 +326,41 @@ def contributors_download
     ]
 
     stats = @active_transcribers.map do |user|
-      time_total = 0
-      time_proportional = 0
+      if user.notification.user_activity?
```
saracarl, line 329: This doesn't work.  Project owners need to know how much work is happening.  The intent of this fix is just to keep project owners from emailing folks.  We should just skip printing their email address in this report, but include their username and time contributed.

app/views/collection/_contributors_body.html.slim
```diff
@@ -30,7 +31,7 @@
     -unless @active_transcribers.empty?
       p
       -@active_transcribers.each do |user|
-          =link_to user.display_name, user_profile_path(:user_id => user.id, only_path: false)
+          =link_to user.display_name, user_profile_path(:user_id => user.id, only_path: false) if user.notification.user_activity?
```
saracarl, line 34: Same issue here.  Only skip email address.

app/views/collection/_contributors_body.html.slim
```diff
@@ -49,7 +50,7 @@ h3 =t('.all_collaborator_emails')
 p=link_to "Find recommendations for project newsletters.", "https://content.fromthepage.com/crowdsourcing-project-newsletter/"
 p#collaborators
   -unless @all_collaborators.empty?
-    =@all_collaborators + ", support@fromthepage.com"
+    =format_collaborators
```
saracarl, line 53: This is the only place that needs support@fromthepage.com.

saracarl (changes requested):

benwbrum:
If user Tom Jones has opted out, he should still appear in the statistics and in the spreadsheet download, but his email should be blank.  His email should also be blank in the contributor email list.

benwbrum:
Users who have opted out should still appear in the statistics and the spreadsheet, but their email should be blank.

app/controllers/collection_controller.rb
diegoviola: @saracarl OK, I changed the code to show emails as "opt out" for users who have their user_activity set to false. Please let me know if this works.

app/helpers/contributor_helper.rb
```diff
@@ -68,7 +68,20 @@ def new_contributors(collection_id, start_date, end_date)
...
+
+  def format_collaborators
+    @collaborators = @all_collaborators.split
+    last_char = @collaborators.last[-1]
+
+    if last_char == ","
+      @all_collaborators + " support@fromthepage.com"
+    else
+      @all_collaborators + ", support@fromthepage.com"
+    end
```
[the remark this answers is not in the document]
diegoviola: @saracarl this is just a helper method used in the view to remove a comma when the comma is already included in the string. It shouldn't affect the CSV at all.

app/views/collection/_contributors_body.html.slim
```diff
@@ -6,10 +6,11 @@
     h3.nomargin =t('.new_collaborators')
     p
       - @new_transcribers.each do |t|
-        =link_to t.display_name, user_profile_path(:user_id => t.id, only_path: false)
-        |
-        =t.email
-        | <br>
+        -if t.notification.user_activity?
```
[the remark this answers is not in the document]
diegoviola: @saracarl done.

app/views/collection/_contributors_body.html.slim
diegoviola: @saracarl right, format_collaborators won't affect the CSV output, it's only there to remove a comma from the view when the string already includes one.

app/controllers/collection_controller.rb
```diff
@@ -337,6 +337,8 @@ def contributors_download
         time_proportional = (@user_time_proportional[user.id] / 60 + 1).floor
       end
 
+      user.email = "opted out" unless user.notification.user_activity?
+
```
saracarl, line 341: Clever.  I like this solution.

saracarl (changes requested):
This is looking good, but there's still 1 piece missing.  Instead of using user.notification.user_activity we need to create a new field user.notification.owner_contact.  It should default to to the user.activity_email (not user.notification.user_activity). 

In the user profile, we should add a "Share my email with project administrators" to set/unset it.

image


And then in this code it should be based on user.notification.owner_contact.

saracarl (changes requested):
Testing this, I get a:
"undefined method `owner_contact?' for nil:NilClass" when I go to the collaborators tab for a collection.  It's in contributor_helper.rb line 72.  The data is an existing user.  The migration defaults to "true", so I'm not sure why this isn't set.

Issue #1712, opened 2020-04-21
Searchable plaintext export mishandling some whitespace


mulcare:
The "searchable plaintext" that a IIIF export points to is mishandling some whitespace. Two carriage returns between text, specifically, behaves oddly. Compare the [first few lines of this page](https://fromthepage.com/harvardlibrary/colonial-north-america-harvard-university-archives/bentley-william-1759-1819-papers-of-william-bentley-1783-1815-an-inventory-letter-from-james-winthrop-to-william-bentley-1812-may-4-hug-1203-5-box-1-harvard-university-archives/transcribe/1050407) to the [searchable plaintext export](https://fromthepage.com/iiif/31971/export/plaintext/searchable).

Returned:
`Cambridge 4 May 1812{Reverend} Sir,Your very learned...`

Expected:
`Cambridge 4 May 1812 {Reverend} Sir, Your very learned...`

saracarl:
A similar report from another customer:

_One of our staff just reported this. I confirmed that it is different than what we see in the Browse display.
On the search results display, there are no spaces between words separated by line breaks. In Browse, the spaces are shown.

Search Example: https://fromthepage.com/display/search?utf8=%E2%9C%93&collection_id=631&search_string=telescope&button=
Browse Example: https://fromthepage.com/whslamc/increase-lapham/box-10-folder-2-correspondence-june-december-1865?page=19
Page Display: https://fromthepage.com/whslamc/increase-lapham/box-10-folder-2-correspondence-june-december-1865/display/889997

I have not verified what happens with the text that is re-imported to Contentdm._

benwbrum:
This is a problem in `SearchTranslator.search_text_from_xml` which needs similar logic to this (from AbstractXmlController):
```ruby

      if preserve_lb
        e.replace_with(REXML::Element.new('br'))
      else
        if params[:action] == "read_work" || params[:action] == 'needs_review_pages' || params[:action] == 'page_plaintext_searchable' || params[:action] == 'work_plaintext_searchable'
          if e.attributes['break'] == "no"
            lb.add_text('')
          else
            lb.add_text(' ')
            lb.add_attribute('class', 'line-break')
          end
        else
          if e.attributes['break'] == "no"
            lb.add_text('-')
          end
        end
      end
```

benwbrum:
Note above the addition of the `' '` to the element if there is no `break="no"` attribute in the `lb` element.

benwbrum:
[...]

PR #2106, opened 2020-10-19
1712 fix newline whitespace plaintext

benwbrum:
Fixes newline problems in plaintext searchable exports.


benwbrum:
Closes #1712

PR #2178, opened 2020-11-29
1926 locale selection

### LINKED ISSUE #1926: Locale detection and selection
Currently, the user's locale is based off of their selected dictation language, which works for testing but is basically a stop-gap.

We'd like to do the following:
* Add an "Interface Language" in the user profile, as well as an `interface_locale` on the `User` model to drive selections.  This should default to `nil`.
* Read the browser's accept language headers to set the user interface.  If the `current_user.interface_locale` is `nil`, the browser configuration should determine the locale, and should be recorded in `current_user.interface_locale`.  If the `current_user.interface_locale` is not `nil`, we should use `current_user.interface_locale` to determine the locale.  Use cases:
  1. A new user has their browser locale set to `es_MX` and logs on to the site.  They should see the interface in Spanish, and their `interface_locale` should be used to send them notifications or emails in Spanish.
  2. The same user updates their user profile to set their interface language to `pt_BR`, without changing their browser locale.  They should see the interface in Portuguese, and should receive email in Portuguese.
* Read a URL parameter to set the interface.  This should never update `interface_locale`.

benwbrum:
Closes #1926.  This implements an interface language on the user profile, which takes priority.  If no language is set for a user, we use the user's browser settings.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,7 @@ gem 'active_link_to'
 gem 'warning'
 
 gem 'rails-i18n', '~> 6.0.0'
+gem 'http_accept_language'
 
 gem 'edtf'
 gem 'edtf-humanize'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -196,6 +196,7 @@ GEM
     hashie (4.1.0)
     highline (2.0.3)
     hodel_3000_compliant_logger (0.1.1)
+    http_accept_language (2.1.1)
     httpi (2.4.5)
       rack
       socksify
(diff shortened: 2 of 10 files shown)


saracarl (changes requested):
When you edit your profile, it looks like the language is the first in the list, "English", which will cause 2 problems:
1)  Someone whose browser language is Spanish, but wants English for FtP won't know they have to save their profile.
2)  Someone who edits their other profile settings and hits "save" will inadvertently change their profile language to English.

Issue #1977, opened 2020-08-15
FindAProject Page Performance [3s]


benwbrum:
Two queries appear to be taking more than a second on the FindAProject page.  Why are we even executing them in the first place?

1675.1 ms

```
app/views/dashboard/landing_page.html.slim:10:in `block in _app_views_dashboard_landing_page_html_slim___3098914021548539018_1592060'
app/views/dashboard/landing_page.html.slim:9:in `each'
app/views/dashboard/landing_page.html.slim:9:in `_app_views_dashboard_landing_page_html_slim___3098914021548539018_1592060'
app/controllers/application_controller.rb:29:in `switch_locale'
[...]

PR #2192, opened 2020-12-02
Address #1977 by tuning queries and caching partials

benwbrum:


benwbrum:
This passes tests locally

PR #2201, opened 2020-12-07
2198 tei editor

benwbrum:
This adds the following URLs
* `/feature/codemirror/enable` enables the TEI editor
* `/feature/codemirror/disable` disables the TEI editor
* `/feature/codemirror` displays current status of the TEI editor
* `/feature/line_numbers/enable` enables line number display in the TEI editor
* `/feature/line_numbers/enable` disables line number display in the TEI editor
* `/feature/line_numbers` displays current status of line numbers within the editor

It also adds display support for tags used within the [CWG-K stylesheet](https://github.com/CivilWarGovernorsOfKentucky/Omeka/blob/fe414acc822f411bbd21c4bc897e2e349b37132a/files/xslt/content/testCWG.xsl)


app/assets/config/manifest.js
```diff
@@ -1,5 +1,13 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
 //= link_directory ../stylesheets .css
+//= link_directory ../stylesheets/codemirror/lib .css
+//= link_directory ../stylesheets/codemirror-buttons .css
```
saracarl, line 5: This is weird -- the space before the .css -- maybe it's just github?

app/controllers/user_controller.rb
```diff
@@ -8,6 +8,30 @@ def demo
...
+      end
+      return
+    end
+    redirect_back :fallback_location => dashboard_role_path
+  end
+
+  def disable_codemirror
+    Flipper.disable(:codemirror)
+    redirect_back :fallback_location => dashboard_role_path
+  end
```
saracarl, line 33: I don't see this method being called anywhere.

config/initializers/flipper.rb
```diff
@@ -0,0 +1,11 @@
+require 'flipper'
+
```
saracarl, line 2: We either need an updated Gemfile with flipper in it or this file shouldn't be here.

saracarl (changes requested):

app/assets/config/manifest.js
saracarl: nope, definitely is a problem on my filesystem as well.

benwbrum: I don't think this is a problem.

saracarl: Discussed and decided this was OK.

PR #2205, opened 2020-12-08
2204 spam mgmt

saracarl:


config/locales/admin/admin-en.yml
```diff
@@ -71,6 +71,8 @@ en:
       delete_the_abusive_content: "Delete the abusive content, leaving other contributions by the user untouched."
       expunge_user: "Expunge User"
       delete_the_user_and_remove_all_contributions: "Delete the user and remove all of their contributions."
+      ok_user: "OK User"
+      mark_this_users_content_as_not_abusive: "Mark this user's content as not abusive."
```
benwbrum, line 75: Should we add these keys to `es` and `pt`?

benwbrum (changes requested):

config/locales/admin/admin-en.yml
saracarl: We don't translate admin side views, so I don't think this is useful/necessary.

Issue #2288, opened 2021-01-21
UI for "Export All Works" screen


saracarl:
In the 1449-bulk-export branch, please style (restyle?)
app/views/bulk_export/new

It's in progress code, so it's pretty rough!

https://github.com/benwbrum/fromthepage/tree/1449-bulk-export

kolking:
@saracarl wrote:

> Ignore the first/old button -- or remove it!

![image](https://user-images.githubusercontent.com/4656448/105410705-cefe3180-5c4b-11eb-8206-0e0650c00ccc.png)

PR #2292, opened 2021-01-22
Bulk export UI

kolking:
Re #2288 

image


benwbrum:
If I click the download link on a previous export, the file is downloaded to my browser, but the javascript spinner runs with no way to exit it.  Can we disable the JS spinner on this page, or disable it when a download happens?

Issue #2266, opened 2021-01-11
New tags and transformations for Folger Shakespeare Library


benwbrum:
This is a place-holder issue for new tags and transformations for the Folger.

benwbrum:
We need to add `catchword` and `supplied`.  Details on how they will be rendered to follow.

benwbrum:
If we allow users to use `lb break="no"` at the end of a line, they can use any kind of word-continuation mark they want (or none at all) instead of being restricted to a line-ending hyphen.

Sample encoding:
`br<lb break="no">=</lb>eak`

Line-preserving rendering:
```
br=
eak
```

Word-wrapping rendering:
`break`

TEI-XML (which does not support child elements of `lb` export:
`br=<lb break="no"/>eak`

benwbrum:
Results of today's discussion of marginal notes:

Add a `marginal` or `marginalia` tag to indicate marginal notes, epitomes, or labels.  Render this (somehow) in the text.  Export TEI-XML as `note`, adding a `type` attribute to differentiate contemporary marginalia from the transcriber comments we also export as `note` elements.

PR #2297, opened 2021-01-25
New abbr tag and better handling of abbreviation display for #2266

benwbrum:
This is part of (but does not resolve) #2266


saracarl (changes requested):
<reg orig=”original text”>regularized text</reg>  needs support

saracarl (approved):

PR #2317, opened 2021-01-30
2316 splash page performance

### LINKED ISSUE #2316: Splash page performance (no index on page.status)
The splash page is loading slowly because we calculate the number of pages transcribed by querying for pages with a non-null status.  The status column has no index.

benwbrum:
Closes #2316


benwbrum:
This PR has bad commits

PR #2358, opened 2021-02-09
Added catchword and marginalia tags plus softbreak handling for #2266

benwbrum:
Adds soft-break handling, and `marginalia` and `catchword` tags.  We should copy the documentation from the buttons or the Folger's docs into our own advanced editor docs.


saracarl:
This branch is stripping out the <lb break="no"/> from the TEI export.

Issue #2195, opened 2020-12-03
resizing pages with carousel skews images


saracarl:
If you go to a page like this one:
https://www.fromthepage.com/unr
Or the Find A Project page:  https://www.fromthepage.com/findaproject

and increase or decrease your font size (control-shift-plus or control-shift-minus) the images in the carousel at the top of the page grow and shrink in a way that skews their proportions.

Issue #2083, opened 2020-09-30
carousel:  words break and images are stretched thinner


saracarl:
> Also, I noticed on my mac (not on my PC) that the images in the slide show get stretched out and some of the line breaks for titles happen in the middle of words (pc or mac). Is there a way to make adjustments? The stretching isn’t extreme, but I think it’s noticeable. What is more important is the width of the text, or the size of the titles, so that words don’t break. I’m attaching screen shots.

<img width="1380" alt="divorce slide image on a mac - stretched" src="https://user-images.githubusercontent.com/708566/94733266-1b380580-032d-11eb-92e4-34a14060f41a.png">
<img width="1318" alt="basque image on a mac - stretched" src="https://user-images.githubusercontent.com/708566/94733287-225f1380-032d-11eb-8188-e97acec9332b.png">
<img width="1325" alt="doten slide image on a mac - stretched" src="https://user-images.githubusercontent.com/708566/94733298-2a1eb800-032d-11eb-8d64-2205c80ed073.png">

PR #2393, opened 2021-02-19
Rework image carousel

kolking:
Re #2195, #2083

Here I've changed the carousel plugin because jCarouselLite is quite old and not maintained.
I've switched to the Slick carousel - https://github.com/kenwheeler/slick/

Both issues reported in #2195 and #2083 should be fixed now.


saracarl:
This is exciting -- I can't wait to test it!

saracarl (changes requested):
This looks great, but it needs some small changes for accessibility.  See in line comments.

(Our customers and open source deployers are large public institutions who are often targeted with lawsuits if their software is not accessible.  If we don't fix this, one of the universities will ask us to in the next 3 months or so.)

kolking:
@saracarl sure, and done. Thanks for pointing me!

Also I wasn't sure about the image size, right now it's limited by 300px (width or height) but if you think the image should be bigger / smaller please let me know.

saracarl:
Thanks!  I like this size.

PR #2396, opened 2021-02-23
1216 deeds and document sets

### LINKED ISSUE #1216: Activity Stream ("show more") doesn't work for document sets
Here's a document set:  https://fromthepage.com/lva/richmond-va

Click on the "Show More" for the Notes and you are taken here: 
https://fromthepage.com/deed/list?collection_id=492

Which is blank.  It's blank because the "collection" is a document set.  An actual collection works (i.e. https://fromthepage.com/deed/list?collection_id=1)

We need deed/list to work for a document set as well as a collection.

benwbrum:
If approved, this resolves #1216.

It requires some pretty extensive testing, however, as the previous functionality listing deeds for a document set showed deeds for works currently in the set, and our new functionality shows deeds on works within a particular document set context.  This means that if a work is moved from one set to another, the old set may still show the old deeds, which may be absent from the new set.


saracarl:
Testing:

- test 1:  a public collection, a private document set (this is an unusual set up).  Add a note.  Note does not show up on the document set deeds list but does on the collection.    DOES NOT WORK AS EXPECTED

- test 2:  a public collection, a public document set (this is how LVA has their county questionnaires set up).  Add a note.  Note does not show up on the document set deeds list but does on the collection.  DOES NOT WORK AS EXPECTED

- test 3: a private collection, a public document set (this is very common, i.e. CWRGM).  Add a note.  The note shows up in the deed list for the document set and in the deed list for the collection, and clicking through keeps the context for both.  Works as expected.

- test 4:  a private collection, a private document set.  If I have access, then I see the deed on both.  If I don't, then I don't see any of the dollection or document set.  Works as expected.

- test 5:  make the private document set in test 4 public.  Note is now visible on the document set deed list in public.  Works. as expected.  

- test 6:  put a note on a public collection.  make collection private.  Put work that note is on in a public document set.  Log out.  Note is visible in the public document set.  Works as expected.

- test 7:  put a note on a private collection.  create a public document set containing that document.  Note is visible on the document set.  Works as expected.

benwbrum:
The most recent commit should fix 1 and 2.

benwbrum:
Replaced by #2445

PR #2406, opened 2021-02-26
check activity setting before sending owner email

saracarl:


lib/tasks/send_owner_emails.rake
```diff
@@ -2,13 +2,11 @@ namespace :fromthepage do
...
   task :collection_stats_by_owner => :environment do
 
-    OWNER_EMAIL_DENYLIST = []
-
     #set variables
     owners = User.where(owner: true).joins(:notification).where(notifications: {owner_stats: true})
 
     owners.each do |owner|
-      unless OWNER_EMAIL_DENYLIST.include?(owner.email)
+      if owner.notification.user_activity
```
benwbrum, line 9: I thought that the flag in question would have been `owner.activity_email`, which is more global?

```
2.7.1 :002 > User.where(:email => 'xxxxx@g.harvard.edu').first.activity_email
 => false 
```

benwbrum: Okay, it looks like the user in question has also turned those off, so this is fine.
```
 => #<Notification id: 191882, add_as_owner: false, add_as_collaborator: false, note_added: false, owner_stats: true, user_activity: false, user_id: 229286, created_at: "2021-02-09 17:56:17", updated_at: "2021-02-23 03:12:38"> 
```

benwbrum (approved):

PR #2445, opened 2021-03-10
Refactored deed display for #1216

### LINKED ISSUE #1216: Activity Stream ("show more") doesn't work for document sets

benwbrum:
Closes #1216, we hope.


app/models/deed.rb
```diff
@@ -31,6 +31,18 @@ def calculate_public
     return true # don't fail validation when is_public==false!
   end
 
+  def show_to?(user)
```
benwbrum, line 34: This appears to be dead code.

Issue #1526, opened 2019-09-18
more features for viewing images


saracarl:
A transcriber asks:

I’d like to see more features for viewing the images. I’ve seen other sites, eg Ancestry.com where you can invert the image colors, rotate the image, and change the contrast.

saracarl:
Can we investigate if openseadragon supports image manipulation options like invert and changing the contrast?  

Related to #1531

saracarl:
Here's an extension for OSD that does a lot of different image enhancements:
https://pages.nist.gov/OpenSeadragonFiltering/

Here's a page that a user tried to enhance themselves:
https://fromthepage.com/display/display_page?page_id=1195334

And their attempt:
https://twitter.com/2021funnies/status/1344267160781250560?s=21

We should:
1)  decide which of the image enhancement options makes the most sense for documents, based on how useful they are on this document.
2)  figure out how to add buttons to OSD to offer this/these image enhancement options.  (Contributing back to the extension)

mulcare:
I'm very interested in this as well. I've had several collaborators ask if it was possible to adjust the brightness and/or contrast of a page, allowing them to more easily work with faded or otherwise difficult to see words and letters.

(As an aside, this is kind of a funny/interesting feature: isn't this essentially what OCR programs do for printed texts as part of the process of identifying letterforms?)

PR #2453, opened 2021-03-22
Image filters

kolking:
Re #1526

NO FILTERS | FILTERS APPLIED
---|---
image | image


benwbrum:
This resolves #1526

benwbrum:
This is really fantastic!

We have one UI change to request -- can you move the controls for filtering into the OSD viewer, near the buttons for zoom, rotation, etc.?  That will be more consistent with the other image controls, and will also be visible if the user uses OSD's full-screen/pop-out option.  (It will probably require adding new buttons to match the existing OSD zoom/fullscreen/etc. buttons.)

kolking:
I have reworked the image toolbar and moved the image filters control into the toolbar. Now they can use the filters in fullpage mode as well.

NEW TOOLBAR | IMAGE FILTERS
---|---
image | image

saracarl (approved):

Issue #2457, opened 2021-03-24
Initial load of an image should fit to image width, not height


benwbrum:
When a user has panned or zoomed the OpenSeaDragon image viewer, we use the same coordinates whenever we display a page image.  If the user has not zoomed or panned images in the past, however, we attempt to fit the new image within the OSD viewer.  This is often not desirable, especially in "Image on Top"/"Image on Bottom" configurations, since the image is usually reduced to something very small.
![Screenshot from 2021-03-24 09-53-06](https://user-images.githubusercontent.com/199961/112331556-02c6f900-8c87-11eb-99c6-5677866ce142.png)

We should instead try to fit the image to the width of the viewer, panned to the top of the image:
![Screenshot from 2021-03-24 09-53-52](https://user-images.githubusercontent.com/199961/112331584-09ee0700-8c87-11eb-900f-1e098246d64f.png)

If the user has already panned/zoomed, we should continue to respect their settings.

PR #2461, opened 2021-03-28
Fit image to viewport width on initial load

kolking:
Re #2457 

On initial load we are checking for the previously stored `json_bounds`. If there are saved bounds, we will apply them. If no saved bounds found, we will position the image to fit viewport width, panned to the top.

VERTICAL | HORIZONTAL
---|---
image | image

However, this will happen only once - the very first time they open the Transcribe page in a browser. And after that there always will be `json_bounds` object saved in the local storage, for any other page opened.

I think that maybe we should go further and re-position the image to fit viewport width every time the layout mode changed? So, every time they switch from "Image at the left / right" to "Image at the top / bottom", the image will be re-positioned using the same behavior.

What do you think?


kolking:
Or, maybe we should save `json_bounds` for a particular page, so zoom + pan applied on the current page won't affect any other page.

kolking:
Or, maybe both - re-position image on every layout change and make each page to have its own `json_bounds` saved in the local storage after user has panned or zoomed.

benwbrum:
Because some documents are scanned with too much empty space, users may zoom in to the useful part of a page, transcribe it, and then move to the next (identically scanned) document.  In that case we don't want them to need to zoom and pan again, so we want `json_bounds` to be kept from page to page.

However, if the user is repositioning their layout, I think it's fair to say that they are modifying the zoom/pan experience and to go to the home/re-positioned image.

So let's ignore `json_bounds` when the layout is changed, but keep it from page to page.

kolking:
I've added the behavior to ignore `json_bounds` and re-position image in the viewport when they change layout mode.

saracarl (approved):

Issue #2343, opened 2021-02-04
Ledger style transcription


saracarl:
> In addition to the current full-text transcription/OCR correction mode, the one-record-per-page code funded by CoSA, and the many-pages-per-record enhancements contributed by LVA, we propose to add support for multiple records per page.  Since many of the tabular documents you shared with us include additional, non-ledger fields at the top or bottom of the page--e.g. county/registrar for a voter list or regiment/unit/commander for a muster roll-- we are integrating the spreadsheet data entry into the field-based transcription system.

> The field configuration screen will allow project owners to add a spreadsheet control for a project, and to configure the columns of the spreadsheet for data type, controlled vocabulary, and width.  Configuring a spreadsheet column will be almost identical to configuring a field in the current system -- "line" and "page" will not apply, but the other options are likely to be the same.  Projects will be able to mix one (and only one) spreadsheet control with regular fields above or below the spreadsheet in order to support fixed header or footer fields. 
> 
> Transcribers will see a web control similar to a spreadsheet on their transcription screen.  Any non-spreadsheet fields will appear above (or below) the spreadsheet.  Drop-downs will appear for controlled vocabularies.  (See attached mock-up.)
> 
> Transcripts will be displayed as HTML tables, which will also be embedded in the full-text export formats.  The table export will export one row per spreadsheet entry (with existing links to the page URL, page and work titles, and imported metadata), and any header or footer fields will be added as additional columns onto the rows from the page they were transcribed from.
> 
> 
> Mock-up of spreadsheet control below header fields for 19th-c Welsh census forms
> 
> Image-Spreadsheet Linking
> It is very easy for the human eye to skip from the first half of one line to the second half of another--or to miss lines entirely--while transcribing, and it requires extra effort to avoid this problem, which can slow down or frustrate volunteers.  We think that highlighting the line a volunteer is transcribing will improve transcriber productivity and enjoyment while reducing transcription errors.  For documents that are consistently formatted and which have been consistently imaged, we have designed a system for automatically highlighting (and panning to) the portion of the page image that corresponds to the row a user is transcribing in the spreadsheet control.  (See image above.)  To enable this, the project owner configuring the system will need to specify the location on a representative page at which the first row begins, the location at which the last row ends, and the number of rows on the table.

saracarl:
Release of HandsOnTable that is available under the MIT license:
https://github.com/handsontable/handsontable/releases/tag/6.2.2

saracarl:
Design notes:

[...]

PR #2463, opened 2021-03-30
Spreadsheet editor UI fixes

kolking:
Re #2343 - Please test it very carefully!

I've updated markup / styles for the Fields tab - forms to configure fields and spreadsheet options. Here everything should be fine or at least better (I hope). But the handsontable plugin is very complicated, so I just tuned it a little. I would not surprise if I did something wrong or something got worse, so once again - test it carefully please.

### DIFF
--- a/app/assets/stylesheets/components/shared.scss
+++ b/app/assets/stylesheets/components/shared.scss
@@ -361,6 +361,49 @@
   table.form + & { margin: $gapSize 0 0 0; }
 }
 
+.flex-toolbar {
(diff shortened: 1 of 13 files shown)


saracarl (approved):
This worked.  I was able to configure a spreadsheet, transcribe a spreadsheet, and export it with no (new) problems.  I did have to merge in 2 changes to get saves to work and tests to (hopefully) pass, but they weren't limited to this branch.

Issue #2478, opened 2021-04-06
add new row to spreadsheet when use tabs out of the last row


benwbrum:
Right now users must bring up a context menu to add a row to the spreadsheet.  We should automatically add a row when the user approaches the end of the spreadsheet.

PR #2487, opened 2021-04-07
NOT READY TO MERGE Add new row to spreadsheet when user selects last row for #2478

benwbrum:
This branch has code that adds a new row to the spreadsheet when a user reaches the last of the starting rows.  

To duplicate:
* Configure a spreadsheet with 4 initial rows
* Transcribe a page (you should see four rows in the spreadsheet)
* Move the cursor to the last row
* A fifth row should be added to the spreadsheet.

### DIFF
--- a/app/views/shared/_handsontable.html.erb
+++ b/app/views/shared/_handsontable.html.erb
@@ -83,7 +83,7 @@
     rowOffsetFromPageTop = topOffset + (rowHeightAsPagePct*(rowNum));
     fuzzyRowOffsetFromPageTop = rowOffsetFromPageTop - margin;
 
-    if(window.viewer.world.getItemCount() > 0) {
+    if (window.viewer.world.getItemCount() > 0) {
       // now transpose for viewport coordinates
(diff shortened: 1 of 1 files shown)


benwbrum:
The functionality works, but the user interface needs to re-size to fit the longer spreadsheet with the added rows.

Here is what I see if I start with 4 rows and move my cursor to rows 4 and 5
Screenshot from 2021-04-07 16-31-48

kolking:
@benwbrum in the first commit I've added hooks to recalculate spreadsheet height when a row inserted or removed. This should solve the initial issue you've mentioned in the PR.

But during testing, I thought that maybe adding a new row every time the very last row selected is not the best behavior. Thus, in the second commit I changed the behavior to insert a new row only when something changed in the last row. So, when user just put the cursor to the last row, it won't insert a new one, but once they enter something in a last row cell, it will add a new row below.

Please test this new behavior, I hope you'll find it reasonable, but if not - we'll rollback to the previous commit.

benwbrum:
I really like this version!

Issue #2483, opened 2021-04-07
Field configuration UI doesn't create new lines in configurator


saracarl:
If you click "add additional line" is doesn't show a new line; it adds a field to line 1 (even if you have a line 2).  
If you click "add additional field" it adds a field to line 1 (even if you have a line 2).

![image](https://user-images.githubusercontent.com/708566/113897343-645f8b00-9790-11eb-9446-ff50023e8267.png)

When you save the above, you end up with a line 3 with 2 new fields, which is sort-of what you want.

![image](https://user-images.githubusercontent.com/708566/113897433-78a38800-9790-11eb-8295-54e79228d878.png)

PR #2495, opened 2021-04-14
Remove extra thead for #2483

benwbrum:


saracarl:
So this fixes #2483, I think, but it doesn't fix the problem that Rice and Maryland are reporting.  

Notice how line 3 has 4 fields on it:
image

But the transcription (and preview) screen breaks those 3 fields across 3 lines:
image

Also -- didn't we use to have up and down arrows on the configuration screen before you saved the initial configuration?  We don't know.

benwbrum:
It looks like the newly added field elements have a margin that expands across the screen.

See the "Bride first name" field at [this example from production](https://fromthepage.com/msa/1973-marriage-certificate/mdsa-s1864-1906/transcribe/1503889) -- perhaps the `field-wrapper` style is conflicting with the inline style?

kolking:
Fixed and should display the rows correctly.
No conflicts with inline styles, the selector was wrong.

Issue #2465, opened 2021-03-30
Drag and drop field/spreadsheet field order


saracarl:
So when I was testing a spreadsheet configuration, I left out the first field on the spreadsheet.  

In order to insert it, I have to 
1) add it to the end of the list of fields
2) Save
3) click the up button
4)  wait for the reload
5) find the field
6)  go to 3
repeat for the number of colums in the spreadsheet, in this case 25.

This is a humongus pain.  Can we do drag and droppable fields?

Once we get it working for spreadsheets, we should also make it work for field based.

kolking:
Yeah, this is definitely something we need to improve 👍

PR #2496, opened 2021-04-18
Sortable spreadsheet columns

kolking:
Ref #2465 - Reorder spreadsheet columns using jQuery sortable plugin.
Depends on #2495


saracarl:
Amazing!  Love this one.  (Can we do it for the fields in the field based configuration in addition to the spreadsheet based one?)

PR #2499, opened 2021-04-19
updated example metadata file to include document sets

saracarl:


app/views/static/metadata.yml
```diff
@@ -1,5 +1,5 @@
 # Title of a work -- this brief string is displayed prominently to end users
-title: '1918'
+title: 'doc set experiment'
```
benwbrum, line 2: I'd prefer '1918' here, as it seemed to fit the hypothetical example better.  Could we restore it and make the document sets line consistent?  (Something like `['Diaries', 'Public`]`?)

benwbrum (changes requested):
See comment inline

app/views/static/metadata.yml
saracarl: fixed.

benwbrum (approved):

PR #2521, opened 2021-04-26
Sortable fields

kolking:
Ref #2465 - Reorder fields using jQuery sortable similar to #2496
The logic here is more complicated because of grouping by line, so please test it carefully.

### DIFF
--- a/app/assets/stylesheets/sections/collection.scss
+++ b/app/assets/stylesheets/sections/collection.scss
@@ -524,6 +524,7 @@
 
 .field-form {
   width: 100%;
+
   margin-bottom: 15px;
 
   th, td {
@@ -543,6 +544,12 @@
     padding-top: 15px;
   }
 
+  &_line {
+    font-weight: bold;
+    font-size: $fontSizeBig;
+    padding: 20px 0 0 0 !important;
+  }
+
   &_label {
     width: 35%;
   }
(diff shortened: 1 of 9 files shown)


saracarl:
So this both isn't passing tests and isn't working when testing manually.  

image

The functionality is great for moving a field within a line.

The UI works great for moving a field from line to line, but when you hit "save" the field reverts to the former line.
image

Additionally the tests fail:
```
1) collection settings js tasks uses page arrows with unsaved transcription
     Failure/Error: page.fill_in('fields_1_First_field', with: "Field one")
     
     Capybara::ElementNotFound:
       Unable to find field "fields_1_First_field" that is not disabled
     # ./spec/features/field_based_spec.rb:105:in `block (2 levels) in <top (required)>'
```

saracarl (changes requested):
@kolking -- please see the comments on the PR.  (If you need help, please let us know.)

kolking:
> The UI works great for moving a field from line to line, but when you hit "save" the field reverts to the former line.

This is a bug, probably I missed something or something changed later. It should save the new position of a field (and the new  line number as well) as soon as you drop the field into a new place. So hitting Save after this _should not_ change the position, because it already updated in the DB. This is how it should work, will figure out what's wrong.

> Additionally the tests fail:

Yes, I saw it and it left me confused, because there should be the `fields_1_First_field` field. As far as I understood the spec file code, this field created a few lines above. So I'm not quite sure why it fails to find the element.

saracarl:
Re:  The test failure.  When I drop a pry into the tests, I see that there isn't a Field 1 on this page:
image
That implies to me that we aren't saving the field configuration change.  If we solve the "save" problem then there is a good chance this test will pass.  (Or at least it isn't worth fixing until we fix the functionality.)

@kolking let us know if you want to fix the "save" or if you'd like us to.

benwbrum:
Closing in favor of #2572

Issue #2202, opened 2020-12-08
Better use of screenspace


saracarl:
Our interface hasn't kept up with increasing screen real estate; can we use the width of the screen better?

saracarl:
After the EMROC transcribathon, the biggest complaint was how little screen space the document images get on when you are transcribing.    Here's a page from one of their projects:  https://fromthepage.com/folger/emroc-transcribathon-2021/ayscough-lady/transcribe/1780372
Specifically, they complained that if you zoomed in on a single word you lost the context of the line because the image window was too narrow.  (This was with with images on the left.)  
Moving images to the top helped some.  

But Ben and I feel like the margins on either side of the actual page content -- while it makes sense for 2013 when we built this -- doesn't make as much sense for a 2021 context where screens and screen resolution is so much higher.

PR #2537, opened 2021-05-02
Transcribe page layout

kolking:
This PR aimed to improve the transcribe page layout and partially address the problem described in https://github.com/benwbrum/fromthepage/issues/2482

The main idea - get rid of the inner scrollers in the edit area, and use `position: sticky` for important UI elements instead. This will change the UI in the following way:
1. The toolbar (Fullscreen / Preview / Save buttons) will stick to the top of viewport and remains visible when they scroll down. Thus, we can remove the button duplicates from the bottom toolbar - because buttons at the top are always visible.
2. When the edit area becomes taller than the height of the image preview at the left, the edit area can be scrolled down by scrolling down the entire page. This means, if there's a spreadsheet with many rows, it won't _completely_ block the edit area from scrolling down. They can scroll by keyboard arrows or just move mouse pointer out of the spreadsheet and scroll using mouse wheel.
3. The image preview area position is also sticky, so it will remain visible until they scroll to the very bottom of the edit area. This means that no matter how long the edit area will be, they will always see the image during transcribing.
4. Also, to address https://github.com/benwbrum/fromthepage/issues/2202 I've increased the maximum content width to 1440px, so the columns will be wider on a big screens.

I've made this video (animated GIF) to demonstrate how it works for a long edit area content:

Kapture 2021-05-02 at 19 40 49

Please test it and let me know what do you think.


saracarl:
This is super slick -- we're really impressed!  

The only concern we have is that the buttons scroll off the screen if you are in over/under view:
losingbuttons3

This doesn't happen if you're in side-by-side view.

saracarl (changes requested):
see comments in the PR

kolking:
Yeah, I see the problem, thanks for making the video. Need to think how to fix that.
Ignore the conflicts in this PR, I will resolve.

saracarl:
@kolking I'd recommend pulling `development` into this before moving forward.  It includes test fixes for your previous changes among other things.

kolking:
@saracarl the issue with the buttons scrolling off the screen should be fixed now. Please test again.

PR #2548, opened 2021-05-07
Save checkbox values as true for #2509

### LINKED ISSUE #2509: ledger checkboxes all result in "false"
If you check a checkbox, the exported spreadsheet will still say "false" for the output.

benwbrum:
Closes #2509


saracarl:
I'm not sure how to handle this, but:
1)  If you check the checkbox it reports "true"
2)  If you check and then uncheck and then save it reports "blank"
3)  If you check and save and then uncheck and save it reports "false"
image
I think we need a better way to handle the unchecked boxes -- they should probably be false.  All the way down?  Only for rows that have other data?  (I like that.

saracarl (changes requested):
Needs some more design work.  See comments in PR.

benwbrum:
For rows that have data, we could display "true" for `true`, and "false" for `false || nil`

saracarl:
Yes, I think that is the right approach.

saracarl:
Better, but if you start with a "true" and then uncheck it and save, you end up with a blank:
image

PR #2549, opened 2021-05-07
Switch ledger to autocomplete for #2493

### LINKED ISSUE #2493: autocomplete for controlled vocabularies in ledger
For controlled vocabulary fields, the ledger steering committee feels autocomplete would be better than a drop down.  It would be better for end users because they don't have to lose the typing flow and switch to the mouse.  

MO is experiencing challenges with users who use the arrow keys to move between cells -- when they get to a dropdown, the cursor keys start selecting items from the dropdown, which frustrates the user.  Consider exploring autocomplete instead.

FL has done similar work with Google Sheets and has been pleased with keyboard-only entry there.

benwbrum:
Resolves #2493


saracarl:
Autocomplete looks great, but I don't think the data field works correctly.
1)  A 4 digit year looks good.
2)  A text month-year is highlighted in red.
3)  A EDTF 01-03-1989, which I'm pretty sure is correct EDTF, is also highlighted in red.
image

saracarl (changes requested):
EDTF needs better pattern matching.  See comments in PR.

benwbrum:
EDTF is only `YYYY-MM-DD` -- year is always at the beginning, since they built EDTF off of ISO-8601.  We need to figure out whether we can create tool-tips or something else that will inform users and owners of what "date" implies here.

benwbrum:
I've added a tooltip functionality for the date fields.  One interesting thing about the HOT plugin I'm using is that we could let project owners define tooltips to be displayed for columns.

saracarl:
I can't get the tooltips to show up, even after a save/reload/etc.  Tried both hovering and clicking on the column header and the cell.  Do I need to enable them somehow?

So if you have a date that is flagged in red upon entry -- like 01-03-1889 -- once you save it isn't in red anymore.  Is this good?  Bad?  I don't know.

saracarl:
I asked Josh at FL to weigh in on the date formatting.  Here's what he said:

> We catalog all dates in EDTF, but so far we haven’t trusted volunteers (or even all staff) to input dates that way for indexing projects without some guardrails. In our data input templates, we have broken date fields up into three subfields and set up data validation/formatting in Google Sheets so the volunteer is forced to enter the parts of the date one part at a time… YYYY, MM, DD. It’s sort of like the fields online where you have to enter a date of birth or credit card expiration date… effectively the dashes are already provided—except in our case each part of the date is a different column. So far this has proven to be the best policy for us.

 

> I see from the GitHub discussion that you are aiming to provide a single date field where the user is expected to input dates in EDTF format, with some red flags if the data is put in incorrectly. Here are my thoughts:

 

> - I think an option for an EDTF date field would be great, but it will work best if the field displays with those dashes or slashes already in there like online forms do for entering a date of birth. People encounter those a lot these days and I think that will help reinforce the expectation. Maybe YYYY MM DD could even be ghosted in the field in the right spots? You often see that in online forms as well. Here is an example of that, although it’s using the MM/DD/YYYY style.
image

 
> - A tool tip or something similar will definitely be necessary, since it doesn’t generally occur to folks that “January 1981” can be written as “1981-01” or that year-only dates are OK. We put all that information in a “Read Me First” file for volunteers and then repeat it in a comment that appears anytime the volunteer mouses over the column header.

> -  Another thing I do to remind the volunteers about the EDTF format is to name the date field “Date (YYYY-MM-DD)” instead of just “Date”

> I hope this helps… let me know what other information I can provide about how we have handled this part of data capture for our indexing projects.

benwbrum:
The library we use allows us to specify `placeholder: 'YYYY-MM-DD'` for these columns, which we should do before deployment.

Issue #2395, opened 2021-02-21
Export directly from Work-level interface


mulcare:
As far as I can tell, currently you can only export transcriptions of Works from the Collection-level interface (or via the IIIF API). Having an Export tab similar to the Collection-level one appear at the Work level would be very helpful. In large collections it can be a bear to find a particular Work to export. 

Even exposing the Work ID somewhere in the Work-level interface would help, as it would allow for easily copying and pasting at the end of a basic in-browser call to one of the export methods (e.g. https://fromthepage.com/export/work_plaintext_verbatim?work_id=38228).

benwbrum:
That seems pretty straightforward to add, and might fit well into some of the researcher-focused export format development work we've got planned.  

I can imagine a new tab, Export, on the work level (along with Read, About, Versions, and Help).  This would show up to any work owner, and to non-owners based on a collection setting (let users download works/let users download works they've contributed to/don't let users download works).  An alternative UI would present a Download link instead of an Export tab.

Clicking this would bring the user to menu of our ever-growing export options, including all existing work-level exports.

mulcare:
Exactly how I could imagine it working, too! It'd be a really valuable addition, I think, both for those of us managing projects and for collaborators themselves.

benwbrum:
Even if volunteers have access to export a work, we should be careful to only show formats that are volunteer-friendly, while still displaying all formats to project owners.

benwbrum:
cf. #700

benwbrum:
Hey, @mulcare, we just deployed new export functionality which should let you export from a new "Download" tab on the work level.  We'd be interested in any feedback on this.

mulcare:
Nice! The new download tab is great. I was actually imagining just being able to quickly get right to the various /export/ endpoint outputs (like I linked in the original post), but understand why dumping most users directly into a plaintext API response isn't ideal. The PDF export looks really nice and clean with the LaTeX-style typesetting. And the Voyant and Word Trees auto-export is a good idea -- I can imagine popping some texts into those on a whim and exploring them, when I otherwise probably wouldn't manually grab all of the content, go to those sites, and paste in the text.

PR #2576, opened 2021-05-20
2395 work export UI

benwbrum:


benwbrum:
This adds a new work-level export screen for #2395 but does not address all the requested functionality there, since some long-running exports will need to be backgrounded.

Issue #2606, opened 2021-06-02
Add field above spreadsheet doesn't work


saracarl:
If you add a field after adding a spreadsheet, you can't drag the field above the spreadsheet (even if it's in a separate line).

Video:
https://app.searchie.io/watch/l7qL6wBMD9

benwbrum:
The problem here appears to be that reordering sends an array of field IDs, which we use to re-order the position of fields within a line, but it does not send the line attribute.  As a result, we don't change the line number of a field that has been reordered.

PR #2608, opened 2021-06-06
Preserve new fields order after save

kolking:
Re #2606 - Interesting bug, took some time to figure it out :)
Reordering was working correct, so if you _refresh the page_ after reordering everything will be fine. But reordering didn't update the line number value in the hidden input, so if you _submit the form_ right after reordering, it will post the wrong (old) line numbers and update. The small fix should resolve the issue.


saracarl:
looks good.  I found one wonky thing where if you put a new field on the same line as a spreadsheet and before the spreadsheet, after you save the field will be the 2nd thing on the line.
I'm merging this anyway because I personally think a spreadsheet should always be on a line by itself.  If that's ever a problem, then we can address this.

Issue #2551, opened 2021-05-10
export link fro email wierdness


saracarl:
> That reminds me, I was going to mention one thing about exports. Previously when I received an email saying that an export was ready, I could click on the link in the email and it would take me to the export page where I could download my requested export. The past few weeks, when I've gotten the email saying the export is ready (we haven't set up the automatic backup yet), I've clicked on the link in the email and it opens a tab that then immediately closes. If I manually go to the export page, I can see the link to download the export and do so, but wanted to mention the link in the email was having trouble for me (and another project owner). Not a big issue at this point, but thought you'd be interested to know. Thanks again for all you do!

(Sherilyn)

Issue #2436, opened 2021-03-08
Facing-page PDF export


benwbrum:
Small institutions would like to export documents as PDFs, ideally in facing-page layout with transcription and image side-by-side.

benwbrum:
This could use better formatting and review.

Issue #2426, opened 2021-03-08
No next action after firing off a bulk export


benwbrum:
If a user starts a bulk export, how do they start a new one?  How do they refresh the export list?  What can they do next?

PR #2640, opened 2021-06-24
2395 bulk export work download

benwbrum:
This PR attempts the following:
* Provide an easier way for project owners to launch exports (issue #2395)
* Improve the usability for bulk exports (issue #2426)
* Make work-specific field/table CSV exports run in the background (TNA)
* Fix problems with email clients that won't download zip files (issue #2551)
* Deliver the facing edition export feature by backgrounding it (issue #2436)

To test:
* The download tab should be available to project owners regardless of whether the feature flag is enabled
* If the feature flag is enabled and the collection allows, it, non-owners should see the download tab but not see the new zip export flags
* The bulk export UI should be more usable; launching a bulk export should send the user to a new "exports" tab.
* Work-specific bulk exports launched from the download tab should work
* Emails about work-specific bulk exports and collection exports should send the user to their exports tab.

### DIFF
--- a/app/controllers/bulk_export_controller.rb
+++ b/app/controllers/bulk_export_controller.rb
@@ -81,6 +81,7 @@ class BulkExportController < ApplicationController
         :subject_csv_collection, 
         :table_csv_work, 
         :table_csv_collection,
-        :work_metadata_csv)
+        :work_metadata_csv,
+        :facing_edition_work)
     end
 end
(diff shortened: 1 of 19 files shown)


saracarl:
We should add some text on the Exports tab to say: `Exports are initiated from each collection.`

So the result should be:
```
Exports are initiated from each collection.

This lists all bulk exports you have run. Exports are cleaned after a few days, and are no longer available for download

Reload this page to update the list.
```

saracarl:
There is no export or download tab showing for works I own.

PR #2647, opened 2021-06-30
2128 reviewer role

benwbrum:


saracarl:
Testing results:

- If I'm logged in as a project owner with review required (but not restricted), I can transcribe and complete a page.  The page then has an "approve" button and  shows up when I click "pages needing review" from the work or collection.  HOWEVER, the review tab is empty.  (note:  there's only 1 work in this collection, but it definitely needs review.)
- Same problem is previous one if I'm logged in as a transcriber rather than owner.
- If you restrict reviewers, it auto-saves and scrolls you back up to the top of the settings page.  We really need a better UI that points them to the next step of adding reviewers.  (They'll miss it otherwise.)
- We should probably list project owners as default "authorized reviewers"
- No users are available in the "add reviewer" pop up.  I can get the arrow to point up or down (and only sometimes), but that's it.  I can't type in the lookup box; it's just a pointer icon.

I'll need the last issue here fixed before I can finish testing.

benwbrum:
I've fixed this blocking issue:  _No users are available in the "add reviewer" pop up. I can get the arrow to point up or down (and only sometimes), but that's it. I can't type in the lookup box; it's just a pointer icon._

benwbrum:
I've fixed the top two issues with the reviewer dashboard hiding works.

PR #2667, opened 2021-07-11
Bulk export pdf docx

benwbrum:
This was requested by LVA, so we should notify Sonya when it is deployed


benwbrum:
Closes #2589

Issue #2648, opened 2021-07-01
Project Owner emails should link to your profile


saracarl:
The `Recent Activity in Your Collections` project owner emails have the following line at the end:
`If you would prefer not to receive these emails, please contact support@fromthepage.com to opt out.`

Instead we should send them to edit their profile.  This is the link, substituting their username for `chowe` here:
`https://fromthepage.com/chowe#update-profile`

benwbrum:
We already do this for user emails, [here](https://github.com/benwbrum/fromthepage/blob/main/app/views/user_mailer/nightly_user_activity.html.slim#L25-L26) and [here](https://github.com/benwbrum/fromthepage/blob/main/app/views/user_mailer/added_note.html.slim#L14-L15)

Here is the problem email which needs to be fixed:
https://github.com/benwbrum/fromthepage/blob/fb125e13ba0c816036ed2b931e7a653194ff7267/app/views/admin_mailer/collection_stats_by_owner.html.slim#L40-L41

PR #2674, opened 2021-07-14
link to profile for #2648

IsaacArneson:
This is in emails so it needs testing.


saracarl:
This needs work.  The link works great, but all it says is "click here", which is confusing, and definitely doesn't help them figure out what to do:

image

Previously, this email said:

`If you would prefer not to receive these emails, please contact support@fromthepage.com to opt out.`

Let's combine these to:

`If you would prefer not to receive these emails, please click here to change your notification settings.`
where "click here" is the link.

PR #2675, opened 2021-07-15
check intercom variable not empty

marcolarosa:
When INTERCOM_ACCESS_TOKEN = '' this code fails. This check ensures there's actually value set.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -294,7 +294,7 @@ class Work < ApplicationRecord
   end
 
   def alert_intercom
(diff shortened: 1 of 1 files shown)


benwbrum (approved):

saracarl:
Travis has stalled twice running automated tests for this.  I think the intercom change may be responsible.  We need to try this locally and see.

benwbrum:
Looks like I need to cherry-pick this into a branch off of `development`, then create a new PR into that branch.

Issue #2575, opened 2021-05-20
performance of settings page 


saracarl:
The settings page does not load quickly because we pre-load users who you *might* want to add as staff/associate project owners (or collaborators on private projects) and that list is very long these days.

@benwbrum has suggested that we show the existing owners and have someone click an "add" button to add more.  The "add" would pop-up a screen that *then* would smart-load some set of users to be added.

We could use this same widget for 

- [ ] adding owners
- [ ] adding collaborators to private projects
- [ ] adding collaborators to restricted works
- [ ] (later) adding reviewers.

Issue #2128, opened 2020-10-26
Reviewer Role


saracarl:
Add a “reviewer” role for users within FromThePage.  These users would be indicated on a collection by collection basis by the project owners.  They have more power than the typical contributor, but less than the project owner.  

Collections start out open to the public for transcription.  As works within the collection are completed, the project owner closes them and they are removed from public transcription.  Then the reviewer comes in and continues to edit after it’s been removed from the public.  
We’ll add a new “review” tab to a collection, visible to the project owner and reviewers.  That tab will have a card per work in the collection.  The card displays the names of the work, the number of pages, the names of the transcribers, and the number of pages with notes.  This data helps the reviewer pinpoint specific works that are higher priority to review (a large number of new transcribers or a lot of notes, for instance, would be a higher priority work to review).  The work card will have two buttons “approve all”, which will mark all pages in a work as approved (useful for works with a single dependable transcriber) and an “approve each page” button which drops the reviewer into the standard work view.  Each page in the standard work view will have an “approve” button, visible only to the reviewer and project owner.  We’ll add “approved” status percentages to the project owner export views.

See also #1749

PR #2714, opened 2021-08-08
Move user search to ajax for #2575 and #2128

benwbrum:


saracarl:
I can only get this to work if I do a shift-reload once the dialog window for adding reviewers is up.

Issue #2239, opened 2020-12-29
don't let project owners save field based projects with no fields


saracarl:
If you have field based transcription configured, but no fields we should either throw a big warning when you save the collection settings and/or put a big red alert at the top of the collection settings page.

benwbrum:
The thing to test for here is `@collection.transcription_fields.present?`.

This should probably go in the `collection_controller.rb` in the `update` method, but it might make more sense to add a validation to the `collection` model to test that the collection is not both field based and missing fields.

benwbrum:
Correction: we can't put this in a model validation, since it is a valid (if short-lived) state for a collection to have no fields configured yet.

saracarl:
The warning you display should say "Please configure fields or choose document based transcription under Transcription Type."

saracarl:
Here's an example of a alert message displayed at the top of the screen:
https://github.com/benwbrum/fromthepage/blob/0cb70e5f471bc0bd6b55fa02b0e4bb4e0e549ce2/app/controllers/transcribe_controller.rb#L40

IsaacArneson:
collection.controller.edit.action

PR #2719, opened 2021-08-11
added warning for #2239

IsaacArneson:


IsaacArneson:
ok this one should finally work

PR #2722, opened 2021-08-16
Hide 'help transcribe' wording for inactive projects

benwbrum:


saracarl:
There's another "start transcribing" button on the owner profile page, i.e. http://localhost:3000/benwbrum, that needs to be conditionalized.

Issue #2437, opened 2021-03-08
Static site export


benwbrum:
From the NEH grant application:

> A popular approach for both preservation and access to digital cultural heritage sites is “minimal computing” -- “computing done under some set of significant constraints of hardware, software, education, network capacity, power, or other factors.”  Exporting work to a static site using the static site generator Jekyll provides a structured set of HTML pages that can be viewed on any web browser without dependencies beyond a simple web browser.  For instance, sites can be published to github pages or even be shared on a thumb drive.  These internally consistent static sites are also ideal for digital preservation.  This project will fund the addition of a static site generation interface to FromThePage.

saracarl:
Input from Joshua:

> 1. Originally, one of the concerns with the html exports was that they wouldn't be easy to move to other platforms like Jekyll. This doesn't appear to be the case anymore from what I can see using the "full" exports of the html. The file maintains the formatting and the structure but it is very difficult to do any editing work on it. If someone wanted to split the file or add additional things to it, it would require extensive knowledge of html. A strip down version of the html would be a much better export for someone wanting to keep the visual layout if they are hosting their own pages. Drupal doesn't complain about the current export either.  

> 2. We had also talked about the image addresses or links being included within the html but as I remember, FtP was not going to keep the objects up long term. One of the ideas was to have the iiif address for each page object within the export so that once an exported page was rendered locally on another computer it would pull the image without needing to go to FtP or UTL. This could be useful in a classroom setting for example when unreliable or low access to bandwidth could limit how much browsing could be done. 

> 3. Ideally, users would be able to export from the collection level or document level and be able to navigate from one item to the next with or without internet access. I haven't implemented this but from what I have done it requires a manifest file with the list of pages that host the collection, document and/or page. If the internet is working fine, then the links point to the online version and creates cache copies locally. If you are offline, then opening the links points to the local copy on your machine and you can keep working. The Brumfields may have a more elegant implementation for this but an automatically updating manifest file is one of the examples I have seen. 

> 4. The last point ties in to a conversation we had about pagination. With the pdfs I believe the suggestion was to have the option of downloading specific sets in order to avoid infinite scrolls. With the html exports it would be useful to have an internal link in the export that can connect one doc to the next if the export is going to be page level. FtP already has this in their menu options and it follows a numbered sequence. 

[...]

PR #2742, opened 2021-08-27
Static site export for #2437

benwbrum:


saracarl:
Changes from Review:

- [ ] if no subjects, can we drop the "subjects" in the left nav?
- [ ] if "about" just has contributors, we should rename it "contributors" (there is about text, but it's on the homepage)
- [ ] get rid of Recent Posts on the homepage
- [ ] get rid of "FEED" in the footer
- [ ] think about what this line in the footer should be for our projects: © 2021 CoSA Ledger Demo. Powered by Jekyll & Minimal Mistakes.
- [ ] remove bullets from the work & subject listings

saracarl:
problems building on github:
`The tag include_cached on line 15 in /_layouts/default.html is not a recognized Liquid tag.`
We need to add 
`  - jekyll-include-cache`
to the list of plugins.

saracarl (changes requested):
needs one more jekyll plugin, then we should be good to merge.

saracarl:
Subject links are not working.  They are as such in the source:
`<a href='/pages/subjects/20' title='Benjamin Franklin Brumfield, Jr.'>Franklin</a>`
and if you edit it by hand to this they will work:
`<a href='../subjects/7203' title='Long Island, Virginia'>Long Island Va.</a>`

saracarl:
On the subject pages, the links to the work/pages that contain the subject also needs to be relativized.

PR #2759, opened 2021-09-10
added More link in work list for collection export.  resolves #2737

### LINKED ISSUE #2737: Add a "more formats" link to the export tab
A new "more" or "more formats" link should appear by each work listed on the collection export screen, taking the user to the Download tab for the work.  In addition, we should add more explanatory text for the Export All Works link.

saracarl:


benwbrum:
I realized we could eliminate the code the old "Zip" link called, then discovered more dead code to eliminate from the old "Export all works" inline zip export generator.

Also internationalized and added a tooltip.

PR #2761, opened 2021-09-10
2524 include metadata contributors in pdf

benwbrum:


saracarl:
If there is no metadata, this looks weird:
Screenshot from 2021-09-13 10-48-46
I think if we moved "Metadata" above the "exported from FromThePage" section it would look fine.  Alternatively, we shouldn't have the "metadata" title when there is no metadata.

saracarl:
Metadata from a Portal to Texas History IIIF manifest is now showing up in the PDF export:
this document: http://localhost:3000/saracarl/dhsi-2020/animal-tracking-journal/about

PR #2762, opened 2021-09-12
2738 dictation codemirror reprise

benwbrum:
Improve codemirror and dictation interoperation.


saracarl:
1)  This is ugly:

Screenshot from 2021-09-13 10-56-16

2)  The fact that this only prints the dictated text after you click the microphone off is not good enough.  It's not a user pattern that shows up elsewhere, and folks are going to expect it to work like voice dictation on your phone works.

saracarl:
After discussion with @benwbrum, we're going to solve this problem by swapping out the code mirror div for a plain text div when voice dictation is enabled.

saracarl:
In the translation view, the text area is too small -- only 2 lines tall.

PR #2781, opened 2021-09-17
WAIT Update export-pt.yml

saracarl:


config/locales/export/export-pt.yml
```diff
@@ -2,3 +2,100 @@ pt:
   export:
     update_contentdm_credentials:
       updating_contentdm_message: "Atualizando CONTENTdm.  Você deve receber um e-mail quando a sincronização terminar, e nesse momento teremos que reconstruir o seu índice para as alterações aparecerem."
+index:
```
saracarl, line 5: Can we add a tab here so it lines up with "updated_contentdm_credentials"?

Issue #2857, opened 2021-10-28
Let users view old metadata while describing new fields


benwbrum:
In order to support the staff-centered use cases of renaming/redescribing/translating existing metadata, we should add a "show metadata" or "original metadata" button to the metadata description screen, the same way we have "show transcription" as a button.  This should render a div identical to the `about` tab.  

The button should not be visible if `work.original_metadata.blank?`

Issue #2849, opened 2021-10-27
advancing from metadata description to the next work for an incomplete document


saracarl:
When working in the transcribe-describe configuration, I’m not able to advance to the next work in a collection until I mark the transcription pages as Done or Needs Review. This may be intentional, but if a user only wanted to work on metadata and not transcription, they’d have to back out to the top layer to get to the next work.

from Meredith

PR #2861, opened 2021-10-28
28xx metadata UI issues

benwbrum:


benwbrum:
Resolves #2849 (probably)

benwbrum:
Resolves #2857

Issue #2504, opened 2021-04-20
Bulk Export of Document Sets


benwbrum:
LVA would really like a bulk export functionality for document sets, allowing them to export a single county at a time.  This seems valuable in lots of other contexts as well.

PR #2866, opened 2021-11-02
Export document sets for #2504

benwbrum:
This adds the export tab to document sets.  It does not implement set-specific coocurrence or CONTENTdm exports, but should support everything else.

### DIFF
--- a/app/views/shared/_collection_tabs.html.slim
+++ b/app/views/shared/_collection_tabs.html.slim
@@ -34,11 +34,7 @@
       :name => t('.settings'),
       :selected => 5,
       :path => "#{edit_collection_path(@collection.owner, @collection)}",
-    },{\
-        :name => t('.export'),
-        :selected => 6,
-        :path => "#{collection_export_path(@collection.owner, @collection)}",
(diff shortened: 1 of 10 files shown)


app/views/shared/_collection_tabs.html.slim
```diff
@@ -34,18 +34,20 @@
...
+    }]
      
   -else
     -tabs += [{\
     :name => t('.settings'),
     :selected => 5,
     :path => "#{collection_settings_path(@collection.owner, @collection)}",
     }]
+
+  -tabs += [{\
```
saracarl, line 46: I'm not sure if the problem is here or elsewhere, but I am not seeing an "export" tab on a document set.

saracarl (changes requested):
Export tab doesn't show up

app/views/shared/_collection_tabs.html.slim
benwbrum: I just tested this locally on CWRGM, and see the Export tab fine if I'm a project owner.

Issue #1953, opened 2020-08-05
"Pages Needing Transcription" doesn't work (potential [regression])


saracarl:
There are two places were the list of needs transcription pages aren't working.  One is in the "Pages Needing Transcription" within a work (or possible .  The other is when you finish a work and click the "next" arrow -- it's supposed to take you to the next untranscribed page.  Instead it says there are no untranscribed pages (even though there are).

PR #2885, opened 2021-11-16
Remove incomplete pages from untranscribed listings for #2869 #1953 #…

benwbrum:
This makes "untranscribed pages" elements in the UI find pages that have not been started.


benwbrum:
Closes #1953 and #2885

Issue #2883, opened 2021-11-16
Add "state" to the sets tab/list for each work


benwbrum:
Make it easier to move documents between sets by visually showing their state.

PR #2888, opened 2021-11-16
#2883 display status instead of collaboration in set work list

benwbrum:


saracarl:
This is just a colored status bar, which isn't particularly clear.  We should add a popup or a textual status in some way to make this more usuable.

saracarl (changes requested):
Make status more usable.

Issue #2686, opened 2021-07-19
"move all" from one document set to another


saracarl:
CWRGM request.  

See notes for how to do this manually in the console.

Issue #2868, opened 2021-11-05
Save vs Done -- done needs to be clearer


saracarl:
Transcribers are not marking things as done.

Maybe "Save and Done" or "Save and Mark Complete"

PR #2889, opened 2021-11-17
#2686 Make done button green

benwbrum:


benwbrum:
This is actually issue #2868.

saracarl:
I'm going to merge this in, but I have concerns it does not pass accessibility guidelines.  (It's hard for me to read; I suspect the "contrast" will be a problem.)

Issue #1389, opened 2019-07-09
collection lost when importing via IIIF


saracarl:
If you go through a collection "add work" tab, and choose to import a IIIIF manifest, you have to *reselect* the collection on the 2nd page.

PR #2890, opened 2021-11-17
Don't lose collection during import #1389

benwbrum:
This also alphabetizes the list of collections in places where that was not the case.


benwbrum:
Resolves #1389.

saracarl:
This does not resolve #1389 .  If I start in the Ann Smith letters, once I'm prompted to select manifests for import I still have to re-choose the collection:

image

saracarl (changes requested):
nil referent and doesn't solve original issue.

saracarl (changes requested):
Not passing tests:

```
1) import data CONTENTdm browses a single record
     Failure/Error: expect(page).to have_content("Manifest: Letter with envelope from Virginia Shewalter")
       expected to find text "Manifest: Letter with envelope from Virginia Shewalter" in "undefined method `slug' for nil:NilClass\n×\nProjectsDashboard\nSigned In As\nGeorge\nEN\nOwner Dashboard\nActions\nCollectionsWorks\nStart A Project\nYour Collections\nSummary\nExports\nUpload PDF or ZIP File\nImport a IIIF Manifest or Collection\nImport from CONTENTdm\nImport from the Internet Archive\nCreate Empty Work\nPage Image Guidelines\nPNG, GIF, and JPG files are all acceptable.\nImages should be oriented so that they are right-side-up.\nImages should be named so that an alphabetical sort will result in the correct page order.\n(This may require \"zero-padding\" for any page numbers: page_09.jpg, page_10.jpg will sort correctly, but page_9.jpg, page_10.jpg will not.)\nDocumentation\nBlog\nAbout\nTerms & Conditions\nPrivacy Policy\nContact Us\n© 2021 FromThePage. All rights reserved."
     # ./spec/features/import_data_spec.rb:31:in `block (3 levels) in <top (required)>'
```

(and downstream tests)

PR #2907, opened 2021-12-03
Added review status to metadata description

benwbrum:


saracarl:
Buttons aren't on the right side of the screen:

image


```
NoMethodError at /josie/ann-smith-letters/11-04-1943-waac/save_description
undefined method `[]' for nil:NilClass

Hint: Something is `nil` when it probably shouldn't be.
```

work_controller on line 37.

saracarl (changes requested):
see notes

Issue #2897, opened 2021-11-29
Make typed-in metadata searchable


benwbrum:
In order for typed-in work metadata to be searchable, we need to add it to the original_metadata fields, then prune it from the exports which add `work.attributes` to `work.original_metadata`

PR #2908, opened 2021-12-07
Make metadata searchable for #2897

benwbrum:
Testing this will require a (fairly) long-running migration.


benwbrum:
The migration will run for about 8 minutes

saracarl:
This works, but I think we should change the text in the search bar to "search by title or metadata"?

saracarl (changes requested):
Change the search bar text

PR #2936, opened 2022-01-13
Quality control workflows

benwbrum:


saracarl:
Failing tests when deleting a collection due to:
`NameError at /collection/delete
uninitialized constant Collection::QualitySamplings`

saracarl:
Testing:
1)  If you "save" a page in the review flow (after making a change), it takes you to the next page, but that page is probably not approved.  Actually, further examination seems to show that it is, so either we remained on the page to move it to completion or it was reviewed.  Will check on the next review.
2)  When you get to the end of a review set, we take you back to the collection and show an error "Couldn't find Page without an ID" (in red as an alert across the top of the page).
3)  If I start a sampling, review anything in the sample (which in this case is everything that needed review in the collection), and go back to the review tab, I see "start sampling" again.  I'm not sure what I should see, but not that.  Maybe "everything has been reviewed"?
4)  "Start Transcribing" button on a collection with pages to transcribe leads to:
`ActionController::UrlGenerationError at /indianaarchives/jeffersonville-land-office-receipts/jeffersonville-land-office-book-9-receipts-12447-to-12996/transcribe/32182134
No route matches {:action=>"display_page", :collection_id=>"jeffersonville-land-office-receipts", :controller=>"transcribe", :page_id=>"32182134", :user_slug=>"indianaarchives", :work_id=>"jeffersonville-land-office-book-9-receipts-12447-to-12996"}, missing required keys: [:quality_sampling_id]`

saracarl:
- [ ] There's no way to get back to the main review dashboard once you're in the sampling list.
- [ ] We should doublecheck the behavior when you are in the review flow, correct a page, and "save" instead of "approve" (or uncheck in Indiana's case).   It moved on, but I'm not sure what the state of that page was.
- [ ] If you only do ~5 pages, every result is "high" or "very high" -- desired or not?

Issue #2949, opened 2022-01-24
UI Tweaks for Metadata


benwbrum:
- Change 'Describe Metadata' to 'Create Metadata'
- Change 'Describe' tab to 'Metadata'
- Move general instructions to the top of the form.

PR #2950, opened 2022-01-24
UI tweaks on metadata for #2949

benwbrum:


saracarl:
I merged this before testing -- oops -- but there are some changes needed:

Instructions need to go below the field in the settings preview.  I think they should also go above the fields in the configuration section.

PR #2971, opened 2022-02-02
Added Staff account type to admin user edit screen #2970

### LINKED ISSUE #2970: Add "Staff" as an account type to the user edit screen

benwbrum:
Closes #2970

### DIFF
--- a/config/locales/admin/admin-en.yml
+++ b/config/locales/admin/admin-en.yml
@@ -49,7 +49,7 @@ en:
       large_institution: "Large Institution"
       trial: "Trial"
       legacy: "Legacy"
-      legacy: "Staff"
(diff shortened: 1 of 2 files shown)


saracarl:
I think it's entering the wrong string into the dB:

image


Let's check on your dev system, because my message catalogs might be out of date.

config/locales/admin/admin-en.yml
```diff
@@ -49,6 +49,7 @@ en:
       large_institution: "Large Institution"
       trial: "Trial"
       legacy: "Legacy"
+      staff: "Staff"
```
benwbrum, line 52: I was missing this key.

PR #2975, opened 2022-02-02
changed the export links to open in a new tab for #2960

sylvieed:
Currently, analyzing a work via the Download page opens links on the same page, which is not what we want. 

This uses `target: :_blank` to open the links on a new page instead.


saracarl:
looks good.

Issue #2939, opened 2022-01-18
ampersands in find a project page not escaped


saracarl:
Ampersands in the Find A Project page aren't escaped, i.e.:

![image](https://user-images.githubusercontent.com/708566/149850157-f731cc38-1a02-4634-86e2-f2eda2daae09.png)

But the same description is on the collection homepage:

![image](https://user-images.githubusercontent.com/708566/149850267-10014301-3645-47e2-b2d4-b4409beef773.png)

sylvieed:
The [Collection homepage](https://fromthepage.com/tngenewhiz/tennessee-legislative-petitions), where the & aren't escaped, [corresponds to](https://github.com/benwbrum/fromthepage/blob/main/config/routes.rb#L65) the [`collection#show` ](https://github.com/benwbrum/fromthepage/blob/main/app/controllers/collection_controller.rb#L129-L197) action. The [view for the page](https://github.com/benwbrum/fromthepage/blob/main/app/views/collection/show.html.slim#L10-L11) shows 
```
@collection.intro_block
```

The [Find A Project page](https://fromthepage.com/findaproject), where the & are incorrectly escaped, [corresponds to](https://github.com/benwbrum/fromthepage/blob/main/config/routes.rb#L316) the [`dashboard#landing_page`](https://github.com/benwbrum/fromthepage/blob/main/app/controllers/dashboard_controller.rb#L141-L160) action. The view for the page has a [`snippet` variable](https://github.com/benwbrum/fromthepage/blob/main/app/views/dashboard/landing_page.html.slim#L12) representing the paragraph where the issue lies, and [it is set as](https://github.com/benwbrum/fromthepage/blob/main/app/views/dashboard/landing_page.html.slim#L6) 
```
-snippet = truncate(strip_tags(c.intro_block), length: 300, separator: ' ') || ''
```


Both of these reference a collection's `intro_block` property, but the one with the issue also calls `strip_tags` on it. As [this issue](https://github.com/rails/rails/issues/28060) and [this site](https://www.rubydoc.info/github/flavorjones/loofah/Loofah/TextBehavior) explain, `strip_tags` is escaping special characters such as &.

### Possible solutions:

1. Replace `strip_tags(c.intro_block)` with `Loofah.fragment(c.intro_block).text(encode_special_chars: false)`
[...]

PR #2994, opened 2022-02-09
Change how tags are stripped for #2939

sylvieed:
_Resolves #2393_

### The problem
Previously, HTML entities like ampersands in collection descriptions in the Find a Project page were incorrectly escaped (i.e `&amp;` is shown instead of `&`). 

escaped html

However, the ampersands show up as expected (`&`) on the Collection overview page and owner dashboard. The difference is that on the Find a Project page, `strip_tags` is called on the description, which [is escaping special characters](https://github.com/rails/rails/issues/28060).

### The solution
To change this behavior, we use `Loofah.fragment(desc).text(encode_special_chars: false)` instead of `strip_tags(desc)`, which does the same thing (removes HTML tags) without escaping special characters ([here's the documentation](https://www.rubydoc.info/github/flavorjones/loofah/Loofah/TextBehavior)). 

`strip_tags` is in three places in the Find a Project page, so they are all changed to use Loofah instead.

### Further Problems

`strip_tags` is still used in [other places in the code](https://github.com/benwbrum/fromthepage/search?q=strip_tags), and if what it's displaying has any HTML entities, they will not be shown correctly. We should change these to use `Loofah` as well (if it applies).

### DIFF
--- a/app/views/dashboard/landing_page.html.slim
+++ b/app/views/dashboard/landing_page.html.slim
@@ -3,7 +3,7 @@
     -@collections.each do |c|
       -cache c do
         -link_page = c.next_untranscribed_page
-        -snippet = truncate(strip_tags(c.intro_block), length: 300, separator: ' ') || ''
+        -snippet = truncate(Loofah.fragment(c.intro_block).text(encode_special_chars: false), length: 300, separator: ' ') || ''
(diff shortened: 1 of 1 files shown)


benwbrum (approved):
Looks good!

Issue #2872, opened 2021-11-09
Change “disable subject linking” to “enable”


saracarl:
Having a negative default that is checked is awkward. We should make it unchecked by default but call it "enable subject linking"

benwbrum:
What screen does this appear in?

benwbrum:
Sorry, this wording appears in the collection settings page.

benwbrum:
This appears in [the project configuration page for the sample collection](https://fromthepage.com/benwbrum/ben-s-test-collection/edit)

benwbrum:
Url is https://fromthepage.com/benwbrum/ben-s-test-collection/edit
This corresponds to a route at https://github.com/benwbrum/fromthepage/blob/development/config/routes.rb#L77
The controller action for this route is at https://github.com/benwbrum/fromthepage/blob/development/app/controllers/collection_controller.rb#L337-L374
The line of code in the view is https://github.com/benwbrum/fromthepage/blob/development/app/views/collection/edit.html.slim#L45
The message key is in this locale file:
https://github.com/benwbrum/fromthepage/blob/development/config/locales/collection/collection.en.yml#L83


We suspect that the easiest fix to this (at least for the display) would be to change [line 44](https://github.com/benwbrum/fromthepage/blob/development/app/views/collection/edit.html.slim#L44) to `value: !@collection.subjects_disabled`

Some additional work might be needed in the `update` action to reverse the boolean in params, but try this first.

PR #2995, opened 2022-02-10
changed disable subject linking to enable for #2872

sylvieed:
_Resolves #2872_

### The problem
Previously, the "subject linking" / "subject indexing" toggle in the Collection settings page had a double negative, which is counterintuitive and confusing.

disable subject indexing

### The solution
Changing the wording to "Enable subject linking" and flipping the boolean going to the database solves this.

enable subject indexing

### Changes made

Previously, the `check_box`es were using a `value` parameter to set their state. However, we found this parameter is not doing anything as `check_box`es [use the `method` passed to them to set their state](https://www.rubydoc.info/docs/rails/4.1.7/ActionView%2FHelpers%2FFormBuilder:check_box). So, all uses of the `value` parameter in the `collection/edit` view were removed.

### DIFF
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -357,6 +357,8 @@ class CollectionController < ApplicationController
   end
 
   def update
+    @collection.subjects_disabled = (params[:collection][:subjects_enabled] == "0") #:subjects_enabled is "0" or "1"
+    
     if params[:collection][:slug] == ""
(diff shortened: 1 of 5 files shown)


benwbrum:
The tests are failing because the `disable_subjects_spec` is looking for the old wording.  Changing that wording (and the click/unclick logic) around line 20 should fix them.

benwbrum (commented):
This looks good!

Issue #2328, opened 2021-02-03
work and collection slug validation -- require a string


saracarl:
Folks are using numbers for slugs (i.e. 1895) and this will or does cause problems.  Let's add a validation that requires slugs to include at least one letter.

sylvieed:
Is the input for the slug the "URL" input on the [collection settings page](https://fromthepage.com/benwbrum/ben-s-test-collection/edit)?

![slug input](https://user-images.githubusercontent.com/35716893/150711213-5d36f872-0f11-468c-a0a5-1f14046c7ab7.jpg)

benwbrum:
Yes, that's the slug input.  (We don't expect our users to know what a slug is ... or at least we didn't when we wrote this.)

sylvieed:
The slug/URL input's value is stored as `@collection.slug` in [the page's view](https://github.com/benwbrum/fromthepage/blob/main/app/views/collection/edit.html.slim#L12). 

To add validation for this value, we can add the following to [line 33 of the collection model](https://github.com/benwbrum/fromthepage/blob/main/app/models/collection.rb#L33).

```
validates :slug, format: { with: [/a-zA-Z/] , message: "#{t('.slug_include_letter')}" }
```

The `@collection.title` validation also needs to be updated with an error message. We'd change [line 32 of the collection model](https://github.com/benwbrum/fromthepage/blob/main/app/models/collection.rb#L32) to:

```
validates :title, presence: true, length: { minimum: 3, maximum: 255, too_short: "#{t('.title_too_short')}", too_long: "#{t('.title_too_long')}" }
```

Wording would be need to be added to the [collection translation files](https://github.com/benwbrum/fromthepage/tree/main/config/locales/collection); in the [English one](https://github.com/benwbrum/fromthepage/blob/main/config/locales/collection/collection.en.yml#L69) after line 69 we can add:

```
title_too_short: "The title must be at least 3 characters long."
title_too_long: "The title cannot be more than 255 characters long."
slug_include_letter: "Please include at least one letter in the URL."
```

To let the user know what's wrong with their input, an error message should be shown below the URL input. Something like the following lines can be added to the [view](https://github.com/benwbrum/fromthepage/blob/main/app/views/collection/edit.html.slim#L16) after line 16.

```
@collection.errors.full_messages.each do |msg|
  tr
    th.hidden URL
  td
    p.nomargin.error =msg
```

This accounts for the case where there is an error in both the Title and URL fields.

[...]

PR #3020, opened 2022-02-22
Added collection, work, and document set slug validation for #2328

sylvieed:
_Resolves #2328_

Previously, the input for slugs/URLs for collections and works had no validation and could be set as just numbers, which causes problems.

Now, when a user tries to save an invalid slug (that has no letters in it, via the regex expression `/[[:alpha:]]/`), an error message will show and the slug will not save.

On the collection page:
collection slug validation

On the work page:
work slug validation

There's also a couple of changes related to #2995 (subject linking enabled) that were necessary to get this working.

### DIFF
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -358,15 +358,14 @@ class CollectionController < ApplicationController
 
   def update
     @collection.subjects_disabled = (params[:collection][:subjects_enabled] == "0") #:subjects_enabled is "0" or "1"
-    
-    if params[:collection][:slug] == ""
-      @collection.update(collection_params.except(:slug))
-      title = @collection.title.parameterize
-      @collection.update(slug: title)
-    else
-      @collection.update(collection_params)
-    end
+    params[:collection].delete(:subjects_enabled)
 
+    @collection.attributes = collection_params
+
+    if collection_params[:slug].blank?
+      @collection.slug = @collection.title.parameterize
+    end
+   
     if @collection.save
       flash[:notice] = t('.notice')
       redirect_to action: 'edit', collection_id: @collection.id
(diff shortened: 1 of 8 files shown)


benwbrum:
We need to add slug validation to the `document_set` model as well.

sylvieed:
Slug validation is now added to `document_set`s

saracarl:
I'm not sure what's going on, but when I test a document set with a number I get:
```ActionView::MissingTemplate at /document_sets/public-tcdl-set
Missing template document_sets/edit, application/edit with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :slim]}. Searched in:
  * "/home/saracarl/dev/products/fromthepage/fromthepage/app/views"
  * "/home/saracarl/.rvm/gems/ruby-2.7.1/gems/devise-4.8.0/app/views"
  * "/home/saracarl/.rvm/gems/ruby-2.7.1/gems/actiontext-6.0.3.2/app/views"
  * "/home/saracarl/.rvm/gems/ruby-2.7.1/gems/actionmailbox-6.0.3.2/app/views"
  ```
  
 this is from line 69 of the document_set_controller.rb.

@sylvieed would you test this in your environment?  (Do you have any document sets?)  I'll go test in the dev branch to see if this was introduced or already there.

saracarl:
On the development branch it worked fine (other than taking a number!).  Here's where I'm testing; the document set settings view:  `http://localhost:3000/saracarl/public-tcdl-set/settings`

sylvieed:
The document slug validation works now:

document set slug validation

However, some slugs that pass validation are causing issues -- if the slug is `1a23` or `1test` (but `2test` is ok), when trying to access the document set (`/sylvied/1a23`) I get a `404` page.

Should I open a separate issue for this?

Issue #2956, opened 2022-01-28
Collection owners should be listed alphabetically


saracarl:
If you add multiple people as owners the owner list in the collection settings tab (i.e. https://fromthepage.com/woodruff/woodruffpapers/edit)  they are listed in the order they were added. They should be listed alphabetically because with long lists it's hard to find someone you want to remove.

This might also be true for collaborators; please check.

PR #3048, opened 2022-03-16
Sorted owners, collaborators, and reviewers alphabetically for #2956

sylvieed:
_Resolves #2956_

Right now, collaborator and owner lists on the collection settings page are listed in the order that they were added, which is unhelpful in finding specific users to remove them. 

These lists are now sorted alphabetically by the users' display name.


benwbrum (approved):
Looks good!

PR #3052, opened 2022-03-21
fix resolution of annotations for internet archive; update iiif refer…

saracarl:
…ences to proper iiif.io links


benwbrum:
It's so beautiful!

Issue #3041, opened 2022-03-11
Update collection statistics to include "pages needing review" and "incomplete pages"


saracarl:
This dashboard of statistics, visible to all transcribers, was developed before we had a lot of pages in "review" or "incomplete" status.  We should add, after "transcribed pages" on the 2nd line, "incomplete pages" and "pages needing review". 
![image](https://user-images.githubusercontent.com/708566/157901949-55ec1fdb-f9b1-4869-9a7d-9b5eed62b3a7.png)

For the "last 7 days" statistics, I'm not sure what to do.  "# pages marked needs review" "# of pages started"?  @benwbrum

PR #3079, opened 2022-04-05
Added incomplete pages and pages needing review to collection statisics for #3041

sylvieed:
_Resolves #3041_

This adds three new numbers to the collection statistics page:

- Incomplete pages
- Pages needing review
- Pages marked needing review (in last 7 days)

### Before

Previously, the statistics page looked like this:

statistics before

or this, if you have more stats to show:

big statistics before

### After

Now, it looks like this:

statistics after

or this, which is a little busy but still readable:

big statistics after

### Other changes made

1. "OCR corrections" and "Pages translated" were removed from the "Last 7 Days Statistics" when OCR correction was disabled or translation wasn't enabled on any works, respectively.
2. All words on the statistics page were moved to locales.


benwbrum:
This looks awesome!  Now that we see it, we should really add "Pages Reviewed" as a stat as well.

sylvieed:
OK, here it is with pages reviewed.

statistics with pages reviewed

PR #3080, opened 2022-04-05
Add "recipient" as new field for Work

sylvieed:
_Resolves #2974_

This adds a new "recipient" field to Works similar to the current "author" field that is present in Work metadata.


app/models/work.rb
```diff
@@ -11,6 +11,7 @@ class Work < ApplicationRecord
      "permission_description",
      "location_of_composition",
      "author",
+     "recipient",
```
sylvieed, line 14: This adds the `recipient` field to IIIF exports as the export uses the Work's `PUBLIC_ATTRIBUTES` to generate its metadata.

sylvieed:
The "recipient" field is now included in the following places:

### Work settings page

work settings

### Work about page

work about

### Example `metadata.yml` file

> Recipient -- the recipient of the document, i.e. for a letter or other targeted document

### TEI-XML exports

```
<recipient>
    <persName>Francisco Villarreal</persName>
</recipient>
```

### IIIF exports

```
{
    "label": "Recipient",
    "value": "Francisco Villarreal"
}
```

The work metadata CSV exports doesn't currently have any work-wide metadata (like author, recipient, place of creation) included, so I didn't add it there.

benwbrum (changes requested):
This all looks spectacular!  Over the course of reviewing, I learned that the TEI XML standard specifies a different way of encoding `recipient`. Rather than adding a tag under `author`, we need a new block in the `profileStmt` specifying a `correspStmt`.

[The `correspStmt` specification](https://cmohge1.github.io/lrbs-digital-editing-intro-2019/TEI-for-Correspondence.pdf) includes a lot of features to support forwarding letters, CCs, and other data modeling items which are beyond our use case.  Rather, if a work has a recipient, we want to assume that it's correspondence and include the sender and recipient in a new `correspStmt` stanza.

So, if the `work.recipient` is not blank, we should add something like this at the beginning of the `profileDesc` element:
```xml
<correspStmt>
  <correspAction type="sent">
    <persName>AUTHOR GOES HERE</persName>
  </correspAction>
  <correspAction type="received">
    <persName>RECIPIENT GOES HERE</persName>
  </correspAction>
</correspStmt>

**Review by benwbrum — APPROVED** [review]
This looks perfect!

**Comment by sylvieed** [comment]
Here's how the TEI export looks now:

```
<correspDesc>
    <correspAction type="sent">
        <persName>the Laredo Mayor</persName>
    </correspAction>
    <correspAction type="received">
        <persName>Francisco Villarreal</persName>
    </correspAction>
</correspDesc>
```

**Comment by benwbrum** [comment]
In testing revealed that the `correspDesc` element is nested directly under the `teiHeader` element.  It needs to be nested within the `profileDesc` element instead.

**Comment by sylvieed** [comment]
Ok, here's where it is now

```
<profileDesc>
    <correspDesc>
        <correspAction type="sent">
            <persName>the Laredo Mayor</persName>
        </correspAction>
        <correspAction type="received">
            <persName>Francisco Villarreal</persName>
        </correspAction>
    </correspDesc>
    <creation>
        <date when="1850-08-17">1850-08-17</date>
    </creation>
    <langUsage>
        <language ident="eng">English</language>
    </langUsage>
</profileDesc>
```

PR #3086, opened 2022-04-08
accept any work slug with alphabet characters or a dash or underscore

saracarl:
LVA has work titles with underscores or dashes, which were not allowed with the alpha regex matching.  Since they upload large zip files, we would get 80% of the way through, run into one of these, and then fail for the entire upload.


benwbrum:
This works as described.  However, when testing a non-Latin slug like ל"עולם, the slug I entered is replaced with a GUID.  Is that correct behavior?

PR #3103, opened 2022-04-13
update masquerade

saracarl:


benwbrum:
If we run into the `Undefined method user_masquerade_path`, something similar is mentioned [here]9https://github.com/oivoodoo/devise_masquerade/issues/66#issuecomment-606068076)

It suggests we change that to `masquerade_path`.

Issue #2976, opened 2022-02-03
Receiving Internal Server 500 Error when using '/' in Document Date field


anabeabas:
Molly of History Revealed emailed us about getting an Internal Server 500 Error. She said it happens when adding information to the Additional Metadata section, specifically in the Document Date field. It wouldn't accept 1786-1787 or 1786 to 1787, but it accepted 1786/1787.

Whenever she tried entering anything other than a '/' in the date field, this showed up:

1 error prohibited the form from being saved:
Document date must be in EDTF format

Fix:
- "change the "date" field on a work's additional metadata to a string

benwbrum:
Changing this date field to use an input mask (like we did in https://github.com/benwbrum/fromthepage/commit/baba15f2c580d7824f59daff8b47f5017bf37f03 ) is probably the best solution.

sylvieed:
Line 26 of `app/views/transcription_field/_metadata_field_layout.html.slim` from that commit

benwbrum:
Further usability enhancements are needed here:
We should display a hint/help text under the date entry field saying something like this:
`Document Date is in the EDTF format (e.g. 1843, 2001-02, 1643-06-30)`

We should also figure out why the code is January-padding: when the user enters 1862, the form reloads as "1862-01-01", which is incorrect.  This might be a problem in an EDTF library we are using, or in the MySQL date column type (if we are using that), or in some ActiveModel attribute accessor.

PR #3115, opened 2022-04-21
Make EDTF date input more intuitive for #2976

sylvieed:
_Resolves #2976_

### The problem

Currently, the "Document Date" field under "Additional Metadata" in work settings is required to be in ETDF format and a validation error pops up if it's not:

> 1 error prohibited the form from being saved:
> Document date must be in EDTF format

This works, but isn't very intuitive because some users (such as Molly of History Revealed, who emailed about this issue) don't know what EDTF format is or don't understand what the validation error means.

### The solution

A better way to restrict dates to EDTF format is to use an input mask (to `yyyy-mm-dd`), this pull adds that.

input mask

### An issue

Though this works perfectly with dates at one specific day, there are more possible date types. [EDTF allows](https://www.loc.gov/standards/datetime/):

- year, month, and day (e.g. 2016-05-10) _(what this allows)_
- year and month (e.g. 2016-10)
- year (e.g. 2016)
- date and time (e.g. 2001-02-03T09:30:01)
- interval (date range) (e.g. 1906-08/1910-12) _(this is what Molly of History Revealed was trying to input)_

The input mask doesn't allow any input outside of `yyyy-mm-dd`, so if we want to include other date types, this solution might not fit.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -213,7 +213,20 @@ class Work < ApplicationRecord
   end
 
   def document_date
-    Date.edtf(self[:document_date])
+    date = Date.edtf(self[:document_date])
+    
+    if self[:document_date].nil? # there is no document date
+      return nil
+    elsif date.nil? # the document date is invalid
+      return self[:document_date]
+    # assign date precision based on length of document_date string (edtf-ruby does not do this automatically)
+    elsif self[:document_date].length == 7 # YYYY-MM
+      date.month_precision!
+    elsif self[:document_date].length == 4 and not self[:document_date].include? "x" # YYYY
+      date.year_precision!
+    end
+    
+    return date.edtf
   end
 
   def document_date_is_edtf
--- a/app/views/work/edit.html.slim
+++ b/app/views/work/edit.html.slim
@@ -97,7 +97,7 @@
           td =f.text_field :location_of_composition, value: @work.location_of_composition
         tr
           th =f.label :document_date, 'Document Date'
-          td =f.text_field :document_date, value: @work.document_date
+          td =f.text_field :document_date, value: @work.document_date, class: 'field-input edtf', data: { inputmask: '"alias": "datetime", "inputFormat": "isoDate"'}
         tr
           th =f.label :genre
           td =f.text_field :genre, value: @work.genre
@@ -189,6 +189,7 @@
 
 -content_for :javascript
   javascript:
+    $('.edtf').inputmask();
     $(function() {
       $('.user-select-form select').select2({
         placeholder: 'Add new collaborator...',
--- a/app/views/work/edit.html.slim
+++ b/app/views/work/edit.html.slim
@@ -97,7 +97,7 @@
           td =f.text_field :location_of_composition, value: @work.location_of_composition
         tr
           th =f.label :document_date, 'Document Date'
-          td =f.text_field :document_date, value: @work.document_date, class: 'field-input edtf', data: { inputmask: '"alias": "datetime", "inputFormat": "isoDate"'}
+          td =f.text_field :document_date, value: @work.document_date, placeholder: 'yyyy-mm-dd'
         tr
           th =f.label :genre
           td =f.text_field :genre, value: @work.genre
@@ -189,7 +189,6 @@
 
 -content_for :javascript
   javascript:
-    $('.edtf').inputmask();
     $(function() {
       $('.user-select-form select').select2({
         placeholder: 'Add new collaborator...',


benwbrum:
If I understand correctly, the input mask performs validation instead of just suggesting text?  And that means we won't be able to accept "1854" by itself as an input value?

sylvieed:
It doesn't perform validation, but it forces the format, not suggests it. If you try to input "1854", it will show as "1854-mm-dd" and you can't remove the "-mm-dd"

input mask strict formatting
 
It's possible I implemented the mask wrong & it's not working how it should.

benwbrum:
Okay, that sounds like we've asked you to implement the wrong thing.

I see that the HTML specifications has some slightly different tools available instead of input mask:
[The `placeholder` attribute](https://www.w3schools.com/tags/att_input_placeholder.asp) works like a suggestion; I think it will disappear when the user types in the location.  Let's try switching to that here.

[The `pattern` attribute](https://www.w3schools.com/tags/att_input_pattern.asp) looks a little trickier, since we'd need to provide a regular expression that expresses essentially all valid EDTF formats.  (Such a regex might actually exist already somewhere.)  This looks like it may force the input to conform to the pattern the same way that a mask does.  If that's the case, we'd need to be very careful to get the regex correct.

Could you switch the mask to a `placeholder`, see how that works, and then explore `pattern` a bit?  We'd prefer accepting non-EDTF dates over not accepting valid EDTF dates.

sylvieed:
Here's how it looks with a placeholder:

document date placeholder

After you type anything in it, the placeholder goes away

document date placeholder after typing

I'll look into a `pattern` more next.

sylvieed:
`pattern` seems to just be a validation that shows up right next to the input box instead of the top of the page where we have it now. 

document date pattern

This message error pops up when you press "Save changes" and could be edited to give clues about EDTF format. It could also be used in conjunction with a `placeholder`.

Personally, I don't like this because having different validations in different places across the site seems confusing/inconsistent. However, it is a more immediate validation and could make more sense to users.

benwbrum:
Let's just go with the `placeholder` solution then.  Your intuition about different validations seems spot-on.

sylvieed:
OK, that's what's on this branch right now.

saracarl:
So the problem with this is that if I type "1943" into the field and save it, it pads out the whole data to "1943-01-01".  This is not desirable.  EDTF says "1943" is a valid date, so it should accept that.

sylvieed:
It looks like the problem with the EDTF dates being padded to day precision is the `edtf-ruby` gem we're using. From [their documentation](https://github.com/inukshuk/edtf-ruby#readme):

> Most, notably, EDTF dates come in day, month, or year precision. This is a subtle difference that determines how many other methods work. **As you can see, dates have day precision by default**:
> ```
> > Date.new(1966).year_precision! == Date.new(1966)
> => false
> ```
> You can set a date's precision directly, or else use the dedicated bang! methods:
> ```
> > d = Date.new(1993)
> > d.day_precision?   # -> true
> > d.edtf
> => "1993-01-01"
> > d.month_precision!
> > d.edtf
> => "1993-01"
> > d.year_precision!
> > d.edtf
> => "1993"
> > d.day_precision?    # -> false
> > d.year_precision?   # -> true
> ```

So, `Date.edtf(2022)` would return `2022-01-01`, which is why we're seeing the `document_date`s padded.

The good news is that the `document_date` is stored as a `string` in the database, and is only converted to EDTF (and padded) when it's accessed through the model. In the work model's `document_date` method:

https://github.com/benwbrum/fromthepage/blob/f679cdd74ed3ecd5612018a25ebb51cef1c9d4f4/app/models/work.rb#L214-L216

```
self[:document_date]            => "2022"
Date.edtf(self[:document_date]) => "Sat, 01 Jan 2022"
```
So we still have access to the precision of all of the `document_dates` previously entered.

saracarl:
Should we check the string and decide how to display it/which EDTF precision to use based on what it contains?

sylvieed:
We'll manually check the length of the `document_date` and assign its precision based on that. 

We'll include coverage for:
- decades and centuries
- `yyyy-mm-dd`
- `yyyy-mm`
- `yyyy`

We probably won't include coverage for:
- Years before 1000 (we can address this if it becomes an issue)
- Other EDTF specs like intervals, times, sets

sylvieed:
Day, month, and year precisions as well as decades & centuries now save!

month precision
year precision
century

benwbrum:
Tests fine.  Merging.

Issue #2952, opened 2022-01-24
restrict work checked with no collaborators


saracarl:
In a private collection, a project owner checked "restrict collaborators" at the work level but didn't add any collaborators.  I'm not sure what the right approach is, but this is a problem.  Definitely catch it when you save?

Maybe the right solution is not to have it for private collections?  

I think there's a UI issue with sending folks to the work settings tab too much; I'll open a different issue for that.

benwbrum:
Let's modify the work settings view to add a warning message if the work is restricted but the work collaborator count is 0.  The warning message should be displayed near the "restrict collaborators" button/option/thingy, and should read "Only project owners can edit this work.  Add collaborators to make the work editable by others."

PR #3132, opened 2022-04-27
Added warning message if work restricted with no collaborators for #2952

sylvieed:
_Resolves #2952_

### The problem

If a work is restricted, its collection is private, and there are no collaborators added, only project owners will be able to edit the work. There are some works that meet all these requirements but are probably unaware of this.

### The solution

This adds a warning message if a work is restricted, its collection is private, and there are no collaborators added. It looks like this:

### DIFF
--- a/app/assets/stylesheets/sections/work.scss
+++ b/app/assets/stylesheets/sections/work.scss
@@ -146,3 +146,10 @@
 .page_status {
   font-style: italic;
 }
+
+// Settings
+
+.warning {
+  font-style: italic;
+  color: $fgRed;
+}
--- a/app/views/work/edit.html.slim
+++ b/app/views/work/edit.html.slim
@@ -168,7 +168,7 @@
         =f.check_box :restrict_scribes
         |&nbsp;
         =f.label :restrict_scribes, 'Restrict collaborators'
-      -if @collection.restricted? && @work.restrict_scribes && @scribes.empty?
+      -if @work.restrict_scribes && @scribes.empty?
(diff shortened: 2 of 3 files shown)


saracarl:
Works right, but is failing tests.  Investigating.

benwbrum:
The test failure happens when we access a page that shows an exception:
```
NoMethodError at /margaret/test-document-set-1/mf1323-1197-reel-32-0046-1/edit ============================================================================== 
undefined method `restricted?' for # Did you mean? restricted 
> To access an interactive console with this error, point your browser to: /__better_errors app/views/work/edit.html.slim, line 171 --------------------------------------- 
166 =f.label :scribes_can_edit_titles, 'Allow collaborators to edit page titles' 
167 p 
168 =f.check_box :restrict_scribes 
169 |  
170 =f.label :restrict_scribes, 'Restrict collaborators' > 
171 -if @collection.restricted? && @work.restrict_scribes && @scribes.empty? 
172 p.warning =t('.restricted_work_no_collaborators_warning') 
173 small.legend =button_tag 'Apply', class: 'outline' 
174 
175 -if @work.restrict_scribes 
176 h3 Allowed Collaborators 

App backtrace ------------- - app/views/work/edit.html.slim:171:in `block in _app_views_work_edit_html_slim__3433860095451451688_616060'
```

This happens while executing this snippet of test code:  https://github.com/benwbrum/fromthepage/blob/main/spec/features/document_sets_spec.rb#L221-L236


So I'd try to edit a work in a document set and see if you can reproduce the error.

Issue #3129, opened 2022-04-27
Hide "review" tab if no pages to review


saracarl:
We're running into problems with the quality sampling code if there are not pages to review.  Let's hide the "review" tab if there are no pages to review.

Related to #3092

PR #3133, opened 2022-04-27
Hide review tab if no pages need review for #3129

sylvieed:
_Resolves #3129_

If there are no pages to review in a collection, the "review" tab will not show up at all.


saracarl:
Looks good.

Issue #2628, opened 2021-06-10
subject coocurrence by document export


benwbrum:
We should modify the CSV export built for #2613 as follows:

- Produce a row _per coocurrence_ rather than aggregating
- Produce a row per coocurrence _within documents_ rather than within _pages_
- Include the work metadata (same as the work metadata csv export)
- Move the coocurrence export to the bulk export screen
- Allow the user to choose to produce a file per subject or a file for the whole collection

benwbrum:
After discussing this with the CWRGM team today, we decided that a spreadsheet per subject (or possibly per category) is more usable and more performant than one giant bulk export.  The export should work as above, listing coocurrences for the subject being exported, and should be accessible from a single URL which can be deep-linked from external (Omeka) sites.

(Extra credit to do this export per category)

PR #3153, opened 2022-05-23
New subject distribution URL and export for #2628

benwbrum:
Subject distribution export is different from subject coocurrence export these ways:

- Limited to a single subject, listing subjects that occur in the same work as that subject
- Lists coocurrences even if they do not appear on the same page, but are on another page in the work
- Aggregated by subject _and work_

### DIFF
--- a/app/controllers/export_controller.rb
+++ b/app/controllers/export_controller.rb
@@ -79,6 +79,12 @@ class ExportController < ApplicationController
     cookies['download_finished'] = 'true'
   end
 
+  def subject_distribution_csv
+    send_data(@collection.export_subject_distribution_as_csv(@article),
+              :filename => "fromthepage_subject_distribution_export_#{@collection.id}_#{Time.now.utc.iso8601}.csv",
+              :type => "application/csv")
+    cookies['download_finished'] = 'true'
+  end
 
   def subject_index_csv
     send_data(@collection.export_subject_index_as_csv,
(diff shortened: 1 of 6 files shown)


saracarl:
Does this replace the existing subject co occurrence spreadsheet?  If so, I see zero difference between this branch and production for JWG.  If not, shouldn't there be an updated view that includes this?

saracarl (changes requested):
Does this replace the existing subject co occurrence spreadsheet? If so, I see zero difference between this branch and production for JWG. If not, shouldn't there be an updated view that includes this?

saracarl:
Test failures:
 
 1) document sets checks document set subject tabs
     Failure/Error: expect(page).to have_content("Description")
       expected to find text "Description" in "NoMethodError at /margaret/test-document-set-1/article/2 T
     # ./spec/features/document_sets_spec.rb:281:in `block (2 levels) in <top (required)>'
  2) subject linking deletes a subject
     Failure/Error: page.find('.tabs').click_link("Settings")
     
     Capybara::ElementNotFound:
       Unable to find css ".tabs"
     # ./spec/features/linking_spec.rb:72:in `block (2 levels) in <top (required)>'

benwbrum:
This now parses metadata values for works in the spreadsheet.

saracarl:
Two more issues:
1) After downloading the spreadsheet from the subject page, I get a spinny that never ends.
2) The data is suspect.  I downloaded Big Henry's spreadsheet from JWG, and "Lila" 66604 shows up twice in it with difference counts on each row (1 and then 16).

benwbrum:
The spinny is a problem.  The data shouldn't be -- each coocurring subject is _per document_, so Lila should have an entry per work she appears in (if Henry also appears in that work).

saracarl (approved):

PR #3155, opened 2022-05-23
2976 edtf document dates

sylvieed:
_Resolves #2976, related to #3115_

This adds a hint for the Document Date input in a work's additional metadata that the date should be in EDTF format.

document date hint

We'd previously discussed either adding this hint or adding (EDTF) after the "Document Date" label; I think the hint is more helpful.


sylvieed:
This PR is bad, look at https://github.com/benwbrum/fromthepage/pull/3156 instead

PR #3169, opened 2022-06-08
WIP: Translate the UI to French

sylvieed:
_Resolves #3165_

First, before translating the UI to French, we want to improve our locales so that we can translate efficiently (only what is necessary). This includes:

- [x] Normalizing keys
- [x] Removing unused keys
- [x] Making sure all keys are localized to at least English
- [x] Making sure all libraries are internationalized (specifically [Devise](https://github.com/benwbrum/fromthepage/issues/3174), [Active Record](https://github.com/benwbrum/fromthepage/issues/3176), and [will paginate](https://github.com/benwbrum/fromthepage/issues/3175))
- [x] Extracting string literals to keys

Then, we added French translations via `i18n-tasks` and the Google Translate API in #3225.

We now are reviewing all of these translations and updating them where they are wrong/bad or don't fit the UI.


config/i18n-tasks.yml
```diff
@@ -18,7 +18,8 @@ data:
     ## Default:
     # - config/locales/%{locale}.yml
     ## More files:
-    # - config/locales/**/*.%{locale}.yml
+    - config/locales/**/*-%{locale}.yml
```
sylvieed, line 21: Our locale files are: `config/locales/controller/controller-locale.yml`, so `i18n-tasks` previously wasn't picking them up by the default of only looking at `config/locales/%{locale}.yml`.

sylvieed:
The locales have been normalized using [`i18n-tasks normalize`](https://github.com/glebm/i18n-tasks#normalize-data). This:

1. Sorted keys in alphabetical order
2. Removed double quotes around translations (`key: "translation"` is now `key: translation`)
3. Added single quotes around translations when there is a `:` in them
4. Removed comments from locale files

config/i18n-tasks.yml
benwbrum: Nice catch!

sylvieed:
Unused keys have been removed using [`i18n-tasks remove-unused`](https://github.com/glebm/i18n-tasks#remove-unused-keys). This removed a lot of keys, most of which weren't being used. Some of the keys it removed were being used, but it didn't pick up, like:

- Where there was a typo in a key
- Keys under a `html` group? (notes, deeds)
- Where the key is not written explicitly in the view, like:
https://github.com/benwbrum/fromthepage/blob/f679cdd74ed3ecd5612018a25ebb51cef1c9d4f4/app/views/page/edit.html.slim#L14
- Some external library locales (Devise, Codemirror)
- ActiveRecord errors

There could be more keys that it incorrectly removed that I missed.

sylvieed:
[Keys that were missing messages](https://github.com/benwbrum/fromthepage/files/8913153/originally-missing-shortened.txt) were found using `i18n-tasks missing` and their messages were added in English.

There are a lot of keys still missing messages in Spanish and Portuguese, I'll open a separate issue for those.

sylvieed:
It looks like `i18n-tasks` can't look for string literals, but there are [other gems](https://github.com/glebm/i18n-tasks/wiki#extract-plain-strings-to-i18n-keys) that can. [`extract-i18n`](https://github.com/pludoni/extract_i18n) looks like the best one; it supports Slim, ERB, and Ruby.

benwbrum:
I'd be in favor of exploring `extract-i18n`.  We'd want to do it in something of a sandbox, though, if it actually modifies all of the views.

sylvieed:
For some odd reason, I couldn't get [`extract-i18n`](https://github.com/pludoni/extract_i18n) to work, so I used [Slimkeyfy](https://github.com/phrase/slimkeyfy) to extract string literals to keys. It's very similar to `extract-i18n` (`extract-i18n` actually uses Slimkeyfy), but doesn't check `ERB` files. We only have a few `ERB` files, so that was OK.

sylvieed:
String literals have been extracted into keys using [Slimkeyfy](https://github.com/phrase/slimkeyfy). This added around 520 keys. Slimkeyfy initially extracted a good chunk of the strings, but still missed a lot, which needed to be extracted manually. 

Places that were skipped:
- All admin-facing views
- Devise (will be covered in https://github.com/benwbrum/fromthepage/issues/3174)
- Static views
- Print views

Places that still need to be extracted:
- User mailers; this will be done after we get a way to test mailers

sylvieed:
User mailers' string literals have now been extracted into keys as well. This added around 50 keys.

sylvieed:
All the issues found in #3232 (except https://github.com/benwbrum/fromthepage/issues/3232#issuecomment-1201332142) are now solved in this PR!

Issue #2385, opened 2021-02-17
Organize subjects vertically instead of horizontally


saracarl:
> On the subjects pages, subject terms would be easier to read and identify if they were organized alphabetically vertically in the column, rather than horizontally.

IsaacArneson:
Here is where the problem lies

https://www.fromthepage.com/benwbrum/jeremiah-white-graves-diaries/subjects#category-260

IsaacArneson:
https://0ca9c3cc648c4c9da4d1bdfd31006feb.vfs.cloud9.us-west-2.amazonaws.com/benwbrum/jeremiah-white-graves-diaries/subjects#category-260

IsaacArneson:
look in app/views/article/list.html.slim:51

PR #3173, opened 2022-06-13
Made subject list organized vertically instead of horizontally for #2385

sylvieed:
_Resolves #2385_

### The problem

> On the subjects pages, subject terms would be easier to read and identify if they were organized alphabetically vertically in the column, rather than horizontally.

This is how the subject list was previously displayed, with subjects alphabetically going across and then down.

horizontal subjects

### The solution

The subjects are now listed alphabetically going down and then across.

vertical subjects

### Method

Because the subjects are showed in 3 columns using `float: left`, they are inherently sorted horizontally and cannot easily be organized vertically (there is no `float: down`). So, to organize them vertically, I could either: (1) change how the subject list is displayed, or (2) place the subjects into the list in an order in which they would be organized vertically. Because I didn't quickly find a way to implement the first, I decided the second option was the best solution.

For each category, there is a 2D array containing the subjects in the order they will be displayed. The array is 3 across and however long it needs to be. In the controller, each subject is placed into this array going down each column. Then, in the view, each subject in the array is placed, going across each row.

### DIFF
--- a/app/controllers/article_controller.rb
+++ b/app/controllers/article_controller.rb
@@ -33,10 +33,10 @@ class ArticleController < ApplicationController
     # Create a 2D array of articles/subjects for each category
     # where the articles are sorted alphabetically vertically
     # instead of horizontally -- so, going down each column
-    @vertical_articles = Array.new(@categories.length()) # Holds articles for each category
-    @categories.each_with_index do |category, i|
+    @vertical_articles = {} # Holds articles for each category
+    @categories.each do |category|
       articles = category.articles_list(@collection) # The articles for this category
-      @vertical_articles[i] = sort_vertically(articles)
+      @vertical_articles[category] = sort_vertically(articles)
     end
 
     # Do the same for uncategorized articles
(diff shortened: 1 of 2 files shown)


app/controllers/article_controller.rb
```diff
@@ -24,9 +25,23 @@ def list
...
+      @uncategorized_articles = @collection.articles.where.not(:id => @collection.articles.joins(:categories).pluck(:id)).reorder(:title)
+    end
+
+    # Create a 2D array of articles/subjects for each category
+    # where the articles are sorted alphabetically vertically
+    # instead of horizontally -- so, going down each column
+    @vertical_articles = Array.new(@categories.length()) # Holds articles for each category
+    @categories.each_with_index do |category, i|
+      articles = category.articles_list(@collection) # The articles for this category
+      @vertical_articles[i] = sort_vertically(articles)
```
sylvieed, line 39: TODO: Convert to hash keyed on category object

sylvieed:
Improvements made:
- Use a hash instead of an array to store the 2d array of articles for each category
- Use `ceiling` to find the number of rows
- Rename previously ambiguous variable `l` to `row`

benwbrum (approved):

PR #3177, opened 2022-06-14
Fix #3172 by reducing double square braces in IA ocr text

### LINKED ISSUE #3172: Internet Archive 422 error during import (USDA)
Tried it again and I am receiving the same error. Here is the work from Internet Archive: https://archive.org/details/CAT30275959004

 

Thanks for your help on this.

benwbrum:
The underlying problem was a page with three lines of text containing `[[`, an OCR artifact which we were interpreting as the beginning of a wiki-link.  This reduces any repeated `[` characters in imported OCR to a single `[`.


sylvieed:
This looks good to me, the [original archive](https://archive.org/details/CAT30275959004) imported with OCR without any problems.

PR #3186, opened 2022-06-24
Fixes #3185 by fixing typo

### LINKED ISSUE #3185: API explodes checking to see if collection is restricted
Hitting this URL causes a 500:
http://localhost:3000/iiif/collection/corbit-burton-family-collection
(Any other document set slug causes the same problem.)

benwbrum:


sylvieed:
This looks good to me; the API worked for both docsets & collections.

Issue #3174, opened 2022-06-13
Internationalize Devise


sylvieed:
_From #1820, helps #3165_

Devise is currently localized manually by us, and it seems like our locale files for Devise are necessary. When they are removed, Devise translations are only shown in English when the language is switched to Spanish or Portuguese. (Devise initially [only has an English localization](https://github.com/heartcombo/devise/tree/main/config/locales))

Since we've already localized Devise to Spanish and Portuguese, we can just keep those translations in place. Looking forwards to French, Devise has [resources to a lot of localizations](https://github.com/heartcombo/devise/wiki/I18n) and/or there is a [`devise-i18n` gem](https://github.com/tigrish/devise-i18n) that we could pull a French locale from.

Also, errors in Devise are not completely translated:

<img src="https://user-images.githubusercontent.com/35716893/172888024-973a36ea-7b53-4366-bb1b-056d5e24e459.jpg" width="58%">

### Next steps

1. add devise-i18n to the gemfile
2. bundle install
3. look and see what messages are translated.
4. identify which views we have overwritten (not entirely sure how; I would think just whatever is in the views/devise directory?) or maybe this is a no-op
5. copy the views/devise directory to a backup directory
6. test again (make sure the i18n stuff is getting picked up for the default views)
7. generate the views (see the devise i18n docs)
8. compare the views to see how much/what we changed.

sylvieed:
### Our changes to Devise views

I added `devise-i18n` to the Gemfile and generated its views. It looks like we've overridden most of the Devise views and made small changes, and have made larger changes to a couple of the views.

#### Small changes
- All overridden FTP views are in Slim, Devise views are in ERB
- Some FTP-specific messages or instructions are included in the FTP views, such as:
  - "Please enter the email address associated with your FromThePage account and we will resend the confirmation instructions at your email address in a few minutes."
  - "Create a new password for your account. Password must be at least 8 characters long and contain both letters and numbers. Please don't use the same password you use for your online bank or email account!"
- Slight changes in wording, such as:
  - "Resend Confirmation" (FTP) vs. "Resend confirmation instructions" (Devise)
  - "Change Password" (FTP) vs. "Change your password" (Devise)

#### Large changes
- FTP sign up, sign in, and profile edit pages are very different from the default Devise ones.
  - Devise only includes an email and password; FTP uses usernames as well as display names.
  - The Devise pages look very different
[...]

PR #3187, opened 2022-06-27
Extracted Devise string literals to keys for #3174

sylvieed:
_Resolves #3174, helps #3165_

This extracts all string literals in Devise views into keys with the help of [Slimkeyfy](https://github.com/phrase/slimkeyfy) so that our Devise views are internationalized.

I skipped the mailer views for now, I'll go back to them when I have a way to test mailers.


sylvieed:
Now Devise mailer string literals have been extracted to keys as well. 

Only the reset password instructions mailer has been changed, as the other two are [dead code](https://github.com/benwbrum/fromthepage/issues/3188).

PR #3216, opened 2022-07-14
Test translation to French on subject versions

sylvieed:
_For #3165_

Before translating all of our locales to French, we should test the translation on one small group. I used `article_version` for this as it only has 9 keys and around 600 characters.


sylvieed:
First, we have to set up for adding French. This includes:
* Adding French as a valid locale to `i18n`, by adding it to `config.i18n.available_locales`
* Adding French as a valid locale to `i18n-tasks`, by adding it to the `locales` list in `config/i18n-tasks.yml`
* Adding French to the language picker on the site. This is added automatically from `i18n`'s `available_locales`
  * Adding the translation of "French" for the language picker
* Adding the `easy_translate` gem, `i18n-tasks` uses this for Google Translate.
* Adding a Google Translate API key to `config/i18n-tasks.yml`

sylvieed:
See #3217 instead

PR #3217, opened 2022-07-15
Test translation to French

sylvieed:
_For #3165_ 

Before translating all of our locales to French, we should test the translation on one small group. I used `article_version` for this as it only has 9 keys and around 600 characters.


benwbrum (approved):
This looks great!

PR #3237, opened 2022-07-28
make spell checker optional instead of disabled

saracarl:


config/locales/collection/collection-en.yml
```diff
@@ -96,6 +96,7 @@ en:
       enable_field_based_transcription: Enable Field Based Transcription
       enable_metadata_description: Enable Metadata Description
       enable_ocr: Enable OCR
+      enable_spellcheck: "Enable spellcheck (not suitable for texts with irregular spelling)"
```
benwbrum, line 99: @saracarl Before we merge this PR, this might be a good opportunity to run through @sylvieed's documentation on i18n maintenance and actually try to add PT/ES/FR messages as well.

saracarl: great idea, and done.

Issue #3232, opened 2022-07-21
Issues with French translations in UI


sylvieed:
There are some issues with the new French translations in the UI, like the wording being too long to fit in buttons, etc. Let's note them here when we see them

sylvieed:
Toggle for "nightly email with activity in collections you've worked on" in "Edit profile" page

![profile- nightly activity toggle](https://user-images.githubusercontent.com/35716893/180304640-9da3d022-e3d8-4f19-813b-8061c7b9fa63.jpg)

https://github.com/benwbrum/fromthepage/blob/559b66f53c09ddd9992b3beac226424c098a7db8/config/locales/user/user-fr.yml#L56

sylvieed:
"Signed in as" on page header

![signed in as](https://user-images.githubusercontent.com/35716893/180304927-05d59e30-f741-4119-a452-17e295745a25.jpg)
https://github.com/benwbrum/fromthepage/blob/559b66f53c09ddd9992b3beac226424c098a7db8/config/locales/layouts/application-fr.yml#L24

sylvieed:
"Browse" when uploading a file on "Start a Project" page

![browse](https://user-images.githubusercontent.com/35716893/180305681-144d15c1-4369-4e6a-bb91-a652e7ad4689.jpg)

https://github.com/benwbrum/fromthepage/blob/559b66f53c09ddd9992b3beac226424c098a7db8/config/locales/dashboard/dashboard-fr.yml#L125

sylvieed:
"Import" when importing a work on "Start a Project" page

![import](https://user-images.githubusercontent.com/35716893/180306911-6e4bb729-b0cc-4fd2-b474-92cbcb22bb15.jpg)

https://github.com/benwbrum/fromthepage/blob/559b66f53c09ddd9992b3beac226424c098a7db8/config/locales/dashboard/dashboard-fr.yml#L80

benwbrum:
Site footer has lower-case "terms et conditions"
![Screenshot from 2022-07-27 09-34-59](https://user-images.githubusercontent.com/199961/181274881-67c97d73-3852-4aec-9c5a-531d2b5d0fad.png)

benwbrum:
Find A Project's Search button is too narrow
![Screenshot from 2022-07-27 09-36-18](https://user-images.githubusercontent.com/199961/181275194-bf833513-5168-400f-905d-ea363f6da1fe.png)

saracarl:
[...]

PR #3243, opened 2022-08-02
Changed deed type hash to be translatable

sylvieed:
_For #3232, deed type buttons not translated_

deed list

Previously, the deed types shown on the deed list (above) weren't translated to non-English locales because they were explicitly written in to the `DeedType` model:

https://github.com/benwbrum/fromthepage/blob/4fe18a3b14ceace5458e394db8151f3b3ef87b32/app/models/deed_type.rb#L30-L35

In this PR I changed the `TYPES` hash to map to calls to `I18n.t` instead of plaintext.

french deed list

This works to translate the deed list (above), but I'm not sure if it will break anything else?

### DIFF
--- a/app/models/deed_type.rb
+++ b/app/models/deed_type.rb
@@ -31,26 +31,26 @@ class DeedType
   # groups of deed types and also their human-readable names. Any new deed type
   # constant should also be added here.
   TYPES = {
-    PAGE_TRANSCRIPTION => 'Page Transcribed',
-    PAGE_EDIT => 'Page Edited',
-    PAGE_INDEXED => 'Page Indexed',
-    PAGE_MARKED_BLANK => 'Page Marked Blank',
-    ARTICLE_EDIT => 'Article Edited',
-    NOTE_ADDED => 'Note Added',
-    PAGE_TRANSLATED => 'Page Translated',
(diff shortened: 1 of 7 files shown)


benwbrum (changes requested):
I'm very suspicious of using `t` during the initialization of constant variables, since my understanding is that these variables will retain their value throughout the lifecycle of the Ruby process.  In development, the `rails s` process serves all requests, while in production `Passenger` connects each new HTTP request to one of a half-dozen Ruby processes which it keeps running at all times.  As a result, I suspect that the values of the deed type constants will be initialized to the locale of the very first user whose request initializes that code, and all follow users will see that same value, regardless of whether it matches their locale or not.

To test this hypothesis, view the deed list, then (without restarting the process) change your locale to a different language and view the list again.  

As an alternative, we could initialize the hash values to a (fully qualified) message key, then use `t` in the partial.

sylvieed:
That makes sense, you're right the locale switcher doesn't actually change the locale of the deed types.

sylvieed:
With the `TYPES` hash mapping to keys for `I18n.t` instead of calling `I18n.t` itself, the deed types are localized with the view instead of as a constant; they now change with the locale switcher.

Issue #3258, opened 2022-08-10
"Pages that need transcription" showing up everywhere


sylvieed:
The header "Pages that need transcription" is showing up even when you're not looking at a list of pages that need transcription.

For example, here is a list of a collection's pages that need review, but the "Pages that need transcription" header is showing up also.

![collection pages](https://user-images.githubusercontent.com/35716893/183964389-788e49a0-8da1-440c-9976-d125bdbb1bb9.jpg)

sylvieed:
When viewing the pages of a work, the header is always "Pages that need transcription" no matter what button you select on the right

![work pages](https://user-images.githubusercontent.com/35716893/183970219-10365311-7818-4988-bddb-9985a260116d.jpg)

PR #3259, opened 2022-08-10
Made pages_view heading changeable for #3258

sylvieed:
_Resolves #3258_

### The issue

Anywhere that pages were being listed (for needing indexing, review, transcription, all pages, etc.) had the heading "Pages that need transcription", but that's not actually what was being shown.

This was because the `_pages_view` partial that all of these use had the heading statically set to "Pages that need transcription".

### The solution

I changed the heading to display a `@heading` variable which each controller can set depending on what type of pages will be shown.

pages

review

transcription

indexing

### DIFF
--- a/config/locales/display/display-fr.yml
+++ b/config/locales/display/display-fr.yml
@@ -75,10 +75,10 @@ fr:
       translation: Traduction
       unable_to_find_pages: Nous n'avons trouvé aucune page correspondant à votre demande
     read_all_works:
-      pages: pages
+      pages: Pages
       pages_that_mention: Pages qui mentionnent %{article}
     read_work:
-      pages: pages
+      pages: Pages
       pages_need_indexing: Pages à indexer
       pages_need_review: Pages à réviser
       pages_need_transcription: Pages nécessitant une transcription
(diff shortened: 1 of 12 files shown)


config/locales/display/display-fr.yml
```diff
@@ -67,14 +67,24 @@ fr:
...
-      pages_that_need_transcription: Pages nécessitant une transcription
       review_metadata: Examiner les métadonnées
       this_page_is_blank: Cette page est vierge
       this_page_is_incomplete: Cette page est incomplète
       this_page_is_not: Cette page n'est pas %{status}, veuillez %{help} cette page
       transcribed: transcrite
       translation: Traduction
       unable_to_find_pages: Nous n'avons trouvé aucune page correspondant à votre demande
+    read_all_works:
+      pages: pages
```
benwbrum, line 78: This is capitalized in other cases.

config/locales/display/display-fr.yml
```diff
@@ -67,14 +67,24 @@ fr:
...
+      pages_that_mention: Pages qui mentionnent %{article}
+    read_work:
+      pages: pages
```
benwbrum, line 81: This is also capitalized in other cases.

benwbrum (changes requested):
A couple of capitalization inconsistencies and this is ready to test.

PR #3263, opened 2022-08-15
Localized prerendered deeds

sylvieed:
_Resolves #3262_

Deeds for short-form lists (i.e. on a collection's page, an owner dashboard, etc.) are pre-rendered and were previously stored as a string of HTML, in the locale of the user who created them. 
This changes the deeds' prerender to store the string of HTML in _each_ locale. When the deed is displayed, it's displayed in the _current_ user's locale.

### Storing

When the deed's prerender is stored, a hash is created that maps each locale (excluding regional locales, to save space as they aren't currently any different for deeds) to the string of HTML in that locale. The deed is rendered in different locales by putting the `_deed` partial in a `I18n.with_locale` block. The hash is then converted to JSON to be stored in the db as a string.

A migration was made to increase the size of `prerender`s in the db from 2047 to 8191; some can be pretty long depending on the length of their titles & links.

### Displaying

There are 3 possible cases for each deed when accessing/displaying it:
1. It's a string, not a hash, because it was created before this change
In this case, attempting to parse the prerender's JSON will fail, will be caught by a `rescue`, and this string is displayed.
2. It's a hash that includes the current user's locale
The JSON will parse and the prerender with the current user's locale is displayed.
3. It's a hash that doesn't include the current user's locale
The JSON will parse but will not have a key for the current user's locale. Look in the `I18n.fallbacks` for that locale, and the first fallback to be a key in the `prerenders` is the locale that the deed is displayed in.
Examples: if the current user's locale is `fr-CA`, the fallbacks would be `[:fr-CA, :fr, :en]`, and the deed is displayed in `:fr`. If the current user's locale is `:it`, the fallbacks would be `[:it, :en]` and the deed is displayed in `:en`.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -119,14 +119,12 @@ module ApplicationHelper
   def show_prerender(prerender, locale) 
     begin
       prerenders = JSON.parse(prerender)
-      if prerenders.key? locale.to_s
-        # show prerender in specified locale
-        prerenders[locale.to_s]
-      else
+      unless render = prerenders[locale.to_s] # show prerender in specified locale
         # prerender doesn't have specified locale, show first fallback that prerender has
-        fallback = I18n.fallbacks[locale].select { |locale| prerenders.key? locale.to_s }.first
-        prerenders[fallback.to_s]
+        fallback = (I18n.fallbacks[locale].map(&:to_s) & prerenders.keys).first
+        render = prerenders[fallback]
       end
+      render
     rescue JSON::ParserError => e
       # prerender is a string, not hash
       prerender
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -116,6 +116,22 @@ module ApplicationHelper
     render({ :partial => 'deed/deeds', :locals => { :limit => limit, :deeds => deeds, :options => options} })
   end
 
+  def show_prerender(prerender, locale) 
+    begin
+      prerenders = JSON.parse(prerender)
+      if prerenders.key? locale.to_s
+        # show prerender in specified locale
+        prerenders[locale.to_s]
+      else
+        # prerender doesn't have specified locale, show default
+        prerenders[I18n.default_locale.to_s]
+      end
+    rescue JSON::ParserError => e
+      # prerender is a string, not hash
+      prerender
+    end
+  end
+
   def validation_summary(errors)
     if errors.is_a?(Enumerable) && errors.any?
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -116,6 +116,23 @@ def deeds_for(options={})
...
 
+  def show_prerender(prerender, locale) 
+    begin
+      prerenders = JSON.parse(prerender)
+      if prerenders.key? locale.to_s
+        # show prerender in specified locale
+        prerenders[locale.to_s]
+      else
+        # prerender doesn't have specified locale, show first fallback that prerender has
+        fallback = I18n.fallbacks[locale].select { |locale| prerenders.key? locale.to_s }.first
```
benwbrum, line 127: I'm pretty sure this could be made a bit more comprehensible using something like `I18n.fallbacks[locale] & prerenders.keys` to grab the fallback.

app/helpers/application_helper.rb
```diff
@@ -116,6 +116,23 @@ def deeds_for(options={})
     render({ :partial => 'deed/deeds', :locals => { :limit => limit, :deeds => deeds, :options => options} })
   end
 
+  def show_prerender(prerender, locale) 
+    begin
+      prerenders = JSON.parse(prerender)
+      if prerenders.key? locale.to_s
```
benwbrum, line 122: this could also be streamlined using assignment within a conditional:
```ruby
unless rendered = prerenders[locale.to_s]
  rendered = #fallback code here
end
# at this point rendered has the value
```

app/helpers/application_helper.rb
sylvieed: Thanks that makes more sense

app/helpers/application_helper.rb
```diff
@@ -119,14 +119,12 @@ def deeds_for(options={})
...
-        prerenders[locale.to_s]
-      else
+      unless render = prerenders[locale.to_s] # show prerender in specified locale
         # prerender doesn't have specified locale, show first fallback that prerender has
-        fallback = I18n.fallbacks[locale].select { |locale| prerenders.key? locale.to_s }.first
-        prerenders[fallback.to_s]
+        fallback = (I18n.fallbacks[locale].map(&:to_s) & prerenders.keys).first
+        render = prerenders[fallback]
       end
+      render
```
benwbrum, line 127: This looks much better, but I'm a little concerned that the variable name `render` may be shadowing the method of the same name.  Can you rename `render` to something less suspicious?

Issue #3245, opened 2022-08-03
Improve Owner Dashboard Summary Stats


saracarl:
A number of folks have asked for improvements to the Owner Dashboard Summary Statistics.  

Specifically, both SSM and the WWPs ask for a way to pull stats over a specific time period (a month in particular).  I think a combo of the existing screen and the collection screen would accomplish that.  

Current Owner Summary Screen:

![image](https://user-images.githubusercontent.com/708566/182713278-860e5ac5-357c-4df7-93c3-a9661095a9a9.png)

Very rough Proposed Owner Summary Screen:

![image](https://user-images.githubusercontent.com/708566/182714539-79b33d0c-466a-48f9-a754-c6a095cfbafa.png)

PR #3274, opened 2022-08-23
First attempt at improving owner dashboard summary for #3245

sylvieed:
_Resolves #3245_

Here's a draft of improving the owner dashboard summary statistics.

summary

### Changes made
- I moved the date range input from the `collaborator_time` partial to the `statistics` one. 
- The second group of statistics now shows the statistics in the given date range instead of the last 7 days.
- Heading changed from "Last 7 days statistics" to "Statistics from date to date" 
- I updated the statistic methods in `owner_statistic.rb` to take date ranges (by changing `last_days_clause` to `date_range_clause`). An input of `7` for "the last 7 days" still works.

I didn't include the "Active collaborators" or "All collaborator emails" sections from the collection collaborators page, but I can add that if it's wanted.

### DIFF
--- a/app/models/owner_statistic.rb
+++ b/app/models/owner_statistic.rb
@@ -49,7 +49,9 @@ module OwnerStatistic
   end
 
   def contributor_count(days=nil)
-    User.joins(:deeds).where(deeds: {collection_id: collection_ids}).where(date_range_clause(days, 'deeds.created_at')).distinct.count
+    User.joins(:deeds).where(deeds: {collection_id: collection_ids})
+      .where.not(deeds: {deed_type: DeedType::COLLECTION_JOINED})
+      .where(date_range_clause(days, 'deeds.created_at')).distinct.count
   end
 
   def comment_count(days=nil)
--- a/app/models/owner_statistic.rb
+++ b/app/models/owner_statistic.rb
@@ -20,15 +20,15 @@ module OwnerStatistic
       .where(status: Page::NEEDS_WORK_STATUSES).count
   end
   
-  def needs_review_count(last_days=nil)
+  def needs_review_count(days=nil)
     Page.where(work_id: self.owner_works.ids)
-      .where("#{last_days_clause(last_days, "edit_started_at")}")
+      .where(date_range_clause(days, "edit_started_at"))
       .where(status: Page::STATUS_NEEDS_REVIEW).count
   end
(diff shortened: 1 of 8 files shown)


app/models/owner_statistic.rb
```diff
@@ -40,50 +40,50 @@ def collection_ids
...
-  def mention_count(last_days=nil)
-    PageArticleLink.where(article_id: owner_subjects.ids).where("#{last_days_clause(last_days, 'created_on')}").count
+  def mention_count(days=nil)
+    PageArticleLink.where(article_id: owner_subjects.ids).where(date_range_clause(days, 'created_on')).count
   end
 
-  def contributor_count(last_days=nil)
-    User.joins(:deeds).where(deeds: {collection_id: collection_ids}).where("#{last_days_clause(last_days, 'deeds.created_at')}").distinct.count
+  def contributor_count(days=nil)
+    User.joins(:deeds).where(deeds: {collection_id: collection_ids}).where(date_range_clause(days, 'deeds.created_at')).distinct.count
```
benwbrum, line 52: We've just heard from the client that they only want collaborators who actually worked on the project.  Since we have a "joined collection" deed that doesn't represent any actual activity, we need to filter by deed_type to make sure it's not included.

app/models/owner_statistic.rb
```diff
@@ -40,50 +40,50 @@ def collection_ids
...
-      timeframe = last_days.days.ago
-      clause = "#{column} >= '#{timeframe}'"
+  def date_range_clause(days, column = "created_at")
+    if days.nil?
+      return ""
+    elsif days.is_a? Integer
+      days = (days.days.ago)..(Time.now)
     end
-    return clause
+    return {column.to_sym => days}
```
benwbrum, line 86: That's fascinating -- I had no idea that active record where clauses could take a Ruby range to create a SQL `BETWEEN ... AND ...`!

benwbrum (changes requested):
One change requested based on discovered requirements.

sylvieed:
If a user has only a deed that's `collection_joined `(for the owner's collections), they won't show up in the contributor count.
If they have a deed that's `collection_joined` _and_ deeds with other deed types (for the owner's collections), they will show up in the contributor count.

benwbrum (approved):

Issue #3273, opened 2022-08-23
clarify transcriber activity time


saracarl:
This, from the collaborator's tab (and maybe somewhere else, too?):
![image](https://user-images.githubusercontent.com/708566/186181779-21887ed3-ab5f-47c4-b0bd-8ecaf0a97160.png)
Is really hard for project owners to understand.  

Could we:
1)  Turn it into a table -- a simple one is fine.
2)  "Time on this collection" and "Time on FromThePage"
3)  button should say "Export detailed activity as CSV"

PR #3276, opened 2022-08-24
Changed transcriber actvity time to a table for #3273

sylvieed:
_Resolves #3273_

This changes the "Active collaborators" section of the collection collaborators page to have a table with collaborator's time instead of the times in parentheses.
It also changes the "Export as CSV" button to say "Export detailed activity as CSV"

active collaborators

I'm pretty sure this is the only page that has this section.

### DIFF
--- a/app/views/collection/_contributors_body.html.slim
+++ b/app/views/collection/_contributors_body.html.slim
@@ -26,7 +26,7 @@
     h3 =t('.active_collaborators')
 
     -unless @active_transcribers.empty?
-      table.tabular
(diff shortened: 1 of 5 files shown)


benwbrum (changes requested):
This table looks unstyled.  Can you use the datagrid/striped classes as we do in https://github.com/benwbrum/fromthepage/blob/fb125e13ba0c816036ed2b931e7a653194ff7267/app/views/dashboard/editor.html.slim#L6

sylvieed:
Here's how it looks now


and with a longer user names

Issue #3286, opened 2022-08-26
Account type isn't translated


sylvieed:
The account type shown on the owner header isn't translated

![legacy account](https://user-images.githubusercontent.com/35716893/186960832-a278adc4-6be9-4a7c-9c1f-3b77fc89eb04.jpg)

PR #3287, opened 2022-08-26
Localized account types for #3286

sylvieed:
_Resolves #3286_

This localizes account types for the owner dashboard header.
Translations were added for 6 account types ("Small Organization", "Individual Researcher", "Trial", "Staff", "Large Institution", and "Legacy") in all 4 languages. To access these account types, a keyfied version of the type is used as a locale key (i.e. `large_institution`).

spanish small organization

### DIFF
--- a/app/views/dashboard/_owner_header.html.slim
+++ b/app/views/dashboard/_owner_header.html.slim
@@ -40,7 +40,8 @@
 
 -if !(current_user.account_type == nil)
   .owner-info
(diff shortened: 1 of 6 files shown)


benwbrum (changes requested):
All of the non-English messages for the `account_type` key need to have the order reversed:
`Cuenta %{type}` or `Compte %{type}`, etc.

PR #3315, opened 2022-09-19
add missing messages

saracarl:


benwbrum:
This is still missing label/description for `footnote` in English and French.

Issue #2811, opened 2021-10-12
Better use of `work.original_filename`


benwbrum:
Currently we store the original filename of an uploaded PDF on the `work` record, but it is not very visible in the UI, nor usable for exports.  Since some organizations embed catalog identifiers in filenames, losing this is problematic.

Enhancements:

- [ ] Display `original_filename` as a read-only field in the work settings tab
- [x] Add an option to use the original filename as the basename of PDF (and other?) formats in the bulk export
- [x] Check that `original_filename` and `slug` are both present in the metadata download spreadsheet for further reconciliation.

benwbrum:
The first task is certainly doable as is the third.  Two may require a bit of discussion.

benwbrum:
While the second task is finished, we need to address the first and third tasks.

To reproduce first task:

- Log in as a project owner
- Create a collection if there is not one already
- From the collection, click the "Add work" tab
- Upload a PDF file to the collection
- Wait for the work to process.
  - (This will create a `Work` object, which should have an `original_filename` attribute matching the PDF file you uploaded
- Visit the collection overview page
- Click on the work
- Click on the About tab.
- If there is a non-blank value in `@work.original_filename`, it should be displayed along with other data about the work.
  - (A location toward the bottom, near the IIIF manifest, is probably best)

For the third task
- After creating a collection and uploading a file--as in task 1--visit the Export screen
- Click the "Work Metadata CSV" button
- Look at the downloaded spreadsheet.  If there are no columns for `original_filename` and `slug`, add them.

symmetrically:
@benwbrum it states that we store the original filename of an uploaded PDF on the work record which has `original_filename` attribute. Although I can see the `uploaded_filename` attribute which I'm considering the attribute you're talking about. Apart from that, I need further clarification about the work data. Which data/attributes should be displayed there?

symmetrically:
Regarding the third task, `original_filename` and slug columns are already there at the downloaded spreadsheet.

benwbrum:
Ah.  `uploaded_filename` is the correct attribute -- I was mistaken.

Currently `app/views/works/show.html.slim` shows most of the attributes we need, excepting the filename, so we just want code similar to https://github.com/benwbrum/fromthepage/blob/development/app/views/work/show.html.slim#L62-L64 for the filename, displayed above the IIIF manifest code.

Issue #2136, opened 2020-10-29
Restructure Zip Export Hierarchy


benwbrum:
Currently, the zip file export produces files structured with one folder per work, each folder containing a folder of formats (HTML vs. TEI vs. plaintext).  Each of those folders contains a file for each level of content (full export vs. transcript-only, etc) and (if applicable) a folder of individual page files.

See this example of a three-work export from the East Civil War Letters:
```ascii
├── george-east-to-thomas-east
│   ├── html
│   │   ├── full.html
│   │   ├── full_pages
│   │   │   ├── Envelope Front.html
│   │   │   └── Letter Front.html
│   │   ├── text.html
│   │   └── transcript.html
│   ├── plaintext
│   │   ├── searchable_transcript.txt
│   │   ├── verbatim_transcript_pages
│   │   │   ├── Envelope Front.txt
│   │   │   └── Letter Front.txt
│   │   └── verbatim_transcript.txt
│   ├── README.txt
│   └── tei
│       └── tei.xml
├── john-c-east-envelope
│   ├── html
│   │   ├── full.html
│   │   ├── full_pages
│   │   │   └── Front.html
│   │   ├── text.html
│   │   └── transcript.html
│   ├── plaintext
│   │   ├── searchable_transcript.txt
│   │   ├── verbatim_transcript_pages
│   │   │   └── Front.txt
│   │   └── verbatim_transcript.txt
│   ├── README.txt
│   └── tei
│       └── tei.xml
└── shemi-watson-to-mary-farthing-shackleford-watson
    ├── html
    │   ├── full.html
    │   ├── full_pages
    │   │   ├── Letter Back.html
    │   │   └── Letter Front.html
    │   ├── text.html
    │   └── transcript.html
    ├── plaintext
    │   ├── searchable_transcript.txt
    │   ├── verbatim_transcript_pages
    │   │   ├── Letter Back.txt
    │   │   └── Letter Front.txt
    │   └── verbatim_transcript.txt
    ├── README.txt
    └── tei
        └── tei.xml
```

The problem with this structure is that libraries generally only care about one type of format, and would like files to be named after the document.

We propose flattening the structure and renaming the files, so that the East files would appear as follows
```ascii
.
├── html-full
│   ├── george-east-to-thomas-east.html
│   ├── john-c-east-envelope.html
│   └── shemi-watson-to-mary-farthing-shackleford-watson.html
├── html-full-pages
│   ├── george-east-to-thomas-east
[...]

PR #3328, opened 2022-09-22
Restructure zip export for #2136 use original filename for #2811

benwbrum:


saracarl:
This configuration:
image
Has the following issues:

1. We include folders for each work (the "old" way), and they appeared first in my zip file, which confused me, since they weren't what I asked for and they were empty.  We need to remove the work-folders if you request format level exports.
2.  We need a readme for type of exports.  I asked for work level html exports, organized by format.  I got 3!
image
3.  The html_full export looks correct -- it has versions with contributors & dates, linked subjects, and subject articles.  
4. The html_text export is transcripts, links subjects, and has subject articles.  Seems reasonable, but I'm not sure what was intended.
5. The html_transcript export does not include contributors/versions or subject articles, but *does* link the subjects (the links go to an anchor that doesn't exist in this file).  This is wrong.

saracarl:
We need to make sure this doesn't close issue #2811 because there are a number of separate changes there that aren't part of this PR.

saracarl:
passing tests.
the pdf export metadata has "in scope: true" listed, even though I didn't mess with the "additional metadata".  That seems like a bug, but outside of this.  Opening a new issue.

benwbrum:
Passing tests locally.  The directory restructuring has been fixed.

Issue #3338, opened 2022-10-03
Clean up uploaded files after successful upload


benwbrum:
Currently uploaded files are stored as file attachments to the `DocumentUpload` model, then extracted to `/tmp/fromthepage_uploads` and processed into pages and works.  Those original attachments are not deleted, so can take up substantial room.

We should clean up attachments as soon as a `DocumentUpload` is successfully processed, and also remove old file attachments in a migration.

PR #3339, opened 2022-10-03
Clean up old document uploads for #3338

benwbrum:
This should clean up document uploads with `created_at` dates of greater than 7 days ago, but needs to be tested as the underlying CarrierWave library really didn't function the way I expected it to.


benwbrum:
Repeat this without the `-T` parameter to `rake` and it should work.

PR #3345, opened 2022-10-06
2820 discussion forums

benwbrum:


app/overrides/controllers/thredded/moderation_controller_override.rb
```diff
@@ -0,0 +1,8 @@
+Thredded::ModerationController.class_eval do
+  private
+
+  def preload_posts_for_moderation(posts)
+    messageboard_ids = @collection.messageboard_group.messageboards.map{|mb| mb.id}
```
saracarl, line 5: messageboard_group is nil here in moderation mode.

Issue #3349, opened 2022-10-13
paginated IIIF collections


saracarl:

saracarl:
Closing this in favor of #3348 -- updates there.

PR #3350, opened 2022-10-13
Handle string links for paginated collection manifests in #3349

benwbrum:


saracarl:
So we need either more instructions on this import, or different functionality.  
I selected *1* item from the first page, and then paged to the second page and selected *1* item there.  Then I imported.  I got the second page's 1 item.  

If I had left "select all" checked, would I have gotten all the items from all the pages?  Let me go test that, although it's slightly hard to know how big the collection is.  (But I can figure it out.

Issue #3364, opened 2022-10-21
Outage during paged search and facet display


benwbrum:
We had a brief outage this morning.  The culprits seem to be two long-running transactions occuring simultaneously, one of them a paged search and the other a faceted search:

```
I, [2022-10-21T12:34:13.126470 #17630]  INFO -- : Started POST "/display_search?article_id=71951&collection_id=cwrgm-public&unlinked_only=true" for 73.101.108.151 at 2022-10-21 12:34:13 +0000
I, [2022-10-21T12:34:13.128755 #17630]  INFO -- : Processing by DisplayController#search as HTML
I, [2022-10-21T12:34:13.129104 #17630]  INFO -- :   Parameters: {"authenticity_token"=>"s5LlKZ6mGc4ripbmCo02A9avE8iyauXAAm/W5RHtWd1Rlvz/rB/287X2njAWd+MALcLnikl3Lzpxf6pZNbGWAQ==", "article_id"=>"71951", "collection_id"=>"cwrgm-public", "unlinked_only"=>"true"}
I, [2022-10-21T12:34:13.150048 #17630]  INFO -- : Redirected to https://fromthepage.com/paged_search?action=search&article_id=71951&authenticity_token=s5LlKZ6mGc4ripbmCo02A9avE8iyauXAAm%2FW5RHtWd1Rlvz%2FrB%2F287X2njAWd%2BMALcLnikl3Lzpxf6pZNbGWAQ%3D%3D&collection_id=cwrgm-public&controller=display&unlinked_only=true
I, [2022-10-21T12:34:13.160690 #17630]  INFO -- : Completed 302 Found in 31ms (ActiveRecord: 8.1ms | Allocations: 12181)
I, [2022-10-21T12:34:13.161144 #17630]  INFO -- : Oink Action: display#search
I, [2022-10-21T12:34:13.161418 #17630]  INFO -- : Memory usage: 1276692 | PID: 17630
I, [2022-10-21T12:34:13.161665 #17630]  INFO -- : Instantiation Breakdown: Total: 10 | PageBlock: 3 | DocumentSet: 2 | User: 1 | Article: 1 | Collection: 1 | Visit: 1 | Ahoy::Event: 1
I, [2022-10-21T12:34:13.161898 #17630]  INFO -- : Oink Log Entry Complete
[...]

PR #3366, opened 2022-10-24
Temporarily disable subject-seeded text search for #3364

benwbrum:
Temporarily disable subject-seeded text search for #3364

### DIFF
--- a/app/views/article/show.html.slim
+++ b/app/views/article/show.html.slim
@@ -41,11 +41,12 @@
     h3= t('.references')
     =render :partial => 'article_links'
 
-    h3= t('.text_search')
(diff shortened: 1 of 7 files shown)


saracarl (changes requested):
If you want to disable this feature, you also need to turn off the test for it:
```
 1) document sets checks document set subject tabs
     Failure/Error: click_button("Search All Pages")
     
     Capybara::ElementNotFound:
       Unable to find button "Search All Pages" that is not disabled
     # ./spec/features/document_sets_spec.rb:289:in `block (2 levels) in <top (required)>'
   ```
  and I'd like to make sure we have an issue that remains open to add it back at some point.

Issue #3399, opened 2022-11-15
alt-text transcription field type/metadata field type


benwbrum:
Some libraries are interested in using FromThePage to create alt-text and other WCAG material.  To enable this, we need to create a new type of `transcription_field`, "alt text", which looks like a regular textarea field, except that it has UI indicators to warn users when they are exceeding recommended length of alt-text contents.

So the UI should let the user type up to N (likely 255) characters, but should start warning them when they exceed 125, then more at 140, then again at 180.  (I'm thinking of shading the text yellow, orange, red).  We should also have a JS widget under the field showing the number of characters they have typed.

PR #3407, opened 2022-11-23
Added alt text transcription and metadata field type for #3399

sylvieed:
_Resolves #3399_

This adds a new transcription field and metadata field type, "alt text". It's the same as a textarea field, but it focuses on keeping the character count under recommended alt-text character counts by:
* Not allowing input longer than 255 characters (by using the `maxlength` attribute)
* Showing the total number of characters under the field
* Shades the input different colors when certain lengths are exceeded (125, 140, 180)

image

Here's how the different shadings currently look:
more than 120 characters - yellow
yellow
more than 140 characters - orange
orange
more than 180 characters - red
red

I picked these colors from the color scheme: `$bgYellow`, `$bgOrange`, and `$bgRed`, but there's a couple issues:
* First, I'm not sure how to use the `scss` colors in JavaScript (maybe it's better to just change the class in the JS and handle the colors in CSS?)
* Second, I don't think these colors are very good for this purpose -- the yellow and orange are too close, and the red doesn't really feel like a warning. Are there any colors FTP already uses for warnings?


sylvieed:
Also, the "n characters" is always in English because it's updated in JavaScript -- I don't think there's a built-in way to use i18n in JS

This might help: https://phrase.com/blog/posts/localizing-javascript-in-rails-apps/

Issue #3380, opened 2022-11-02
Customizing the Signature in Support Emails for Project Owners


anabeabas:
Patti Oshiro of Punahou asked if support emails (see screenshot below) that are signed with "FromThePage" could be replaced with the name of the institution.

![image](https://user-images.githubusercontent.com/98731543/199511499-cf12cdaa-1c5f-4e29-b824-b80bb03766f4.png)

PR #3419, opened 2022-12-13
Use institution name in email signature for #3380

sylvieed:
_Resolves #3380_

This adds a setting for collections to use the collection owner's institution name in user mailer email signatures instead of "FromThePage". It adds a new boolean field to the `collection` model, `institution_signature`. It's defaulted to `false`, which means user emails will be signed with FromThePage. If it's updates to `true`, user emails will be signed with the owner's display name.
I chose to keep this setting simple and just use a toggle instead of a fully customizable signature with a textbox because I assumed institutions wouldn't want to sign emails with anything other than their set display name.

Here's the setting:

setting
("Drexel Libraries" Is whatever the collection owner's display name is)

Here's how the emails look, with the customized signature:
html email

and default signature: (this is unchanged)
html email fromthepage

The emails this updates:
* Note added
* Collection collaborator
* Collection reviewer
* Work collaborator

The other user mailer emails are unchanged because they don't have a signature. The nightly user activity email is unchanged because it's not from a specific institution. Both text and html emails are updated.

### DIFF
--- a/db/migrate/20221213173527_add_institution_signature_to_collection.rb
+++ b/db/migrate/20221213173527_add_institution_signature_to_collection.rb
@@ -1,0 +1,5 @@
+class AddInstitutionSignatureToCollection < ActiveRecord::Migration[6.0]
+  def change
+    add_column :collections, :institution_signature, :boolean, default: false
+  end
+end
--- a/db/migrate/20221219210301_remove_institution_signature_from_collection.rb
+++ b/db/migrate/20221219210301_remove_institution_signature_from_collection.rb
@@ -1,0 +1,5 @@
+class RemoveInstitutionSignatureFromCollection < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :collections, :institution_signature, :boolean
+  end
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
   - export QMAKE=/usr/lib/x86_64-linux-gnu/qt5/bin/qmake
   - qtchooser -qt=qt5
   - mysql -e 'CREATE DATABASE test;'
-  - nvm install --lts
+  - nvm install 16
 
 before_script:
   - bundle exec rake db:create
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ gem 'iiif-presentation', git: 'https://github.com/benwbrum/osullivan', branch: '
 gem 'omniauth-saml'
 gem 'omniauth-google-oauth2'
 gem 'omniauth-multi-provider', '~> 0.2.1'
-gem 'omniauth', '~> 1.9.1'
+gem 'omniauth', '~> 1.9.2'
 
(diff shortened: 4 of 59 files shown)


saracarl:
Would we want the institution name in quotes?  

This implies that responses go to the institution; they don't.  Should we set a reply-to that's the institution's account email?  (Complicated issues)  Maybe just a disclaimer or a "Please respond on the page here rather than responding to this email"

sylvieed:
Added quotes around the institution names in the setting description:

quotes

sylvieed:
Changed the note added email to say "respond to the note" instead of "view the note" so it's more clear where the users should reply.

respond to

Added a reply-to to the emails from a specific collection to the collection owner's email. The reply-to is always there, not only when the owners check the "institution signature" checkbox for that collection.

replyto

db/migrate/20221213173527_add_institution_signature_to_collection.rb
```diff
@@ -0,0 +1,5 @@
+class AddInstitutionSignatureToCollection < ActiveRecord::Migration[6.0]
+  def change
+    add_column :collections, :institution_signature, :boolean, default: false
```
saracarl, line 3: Default to true.

saracarl (changes requested):
Hi Sylvie,

We talked about this some more, and realized that this option should be what happens all the time.  In other words, we shouldn't give institutions the option to choose.

So....  could you please back out the UI to choose sending with the institution name, and instead default to true.  You won't need the schema change because it will always be true.

Sometimes the institution's display name is blank -- we're not sure why (some sort of editing the profile causes this problem, if you want to look into that -- or search for related issues we might have one.).  Let's handle that case either by figuring it out, or more likely checking for a blank display name and using FromThePage in that case.

db/migrate/20221213173527_add_institution_signature_to_collection.rb
sylvieed: Could we just remove the `institution_signature` column completely because it won't be used anywhere?

sylvieed:
I removed the `institution_signature` column as it won't be used anymore.

Now, all of the user mailer emails that are from a specific project owner (i.e. about a specific collection, work) are signed by that project owner unless they have a blank display name, and then it's signed by FromThePage.

db/migrate/20221219210301_remove_institution_signature_from_collection.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveInstitutionSignatureFromCollection < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :collections, :institution_signature, :boolean
```
saracarl, line 3: So in a more perfectionist environment, you'd probably remove the add migration for this column instead of adding the column and then removing it.  But we're going to roll with it because we don't care that much.

sylvieed: Makes sense, thanks

Issue #1043, opened 2018-04-15
Restricted collaborator visibility


benwbrum:
Restricted collaborators have bugs, in which we may show the "help transcribe" button to people who are not allowed to transcribe.

benwbrum:
We need specific examples of this issue.

saracarl:
This caused a problem today because USDA restricted their works in a popular collection that didn't have good instructions (they wanted to work on the instructions).  The works still look like this one the collection page:

![image](https://user-images.githubusercontent.com/708566/206505709-302682c6-7fb8-4e10-b600-9cdb418afe3d.png)

When you click on "help out" you are redirected to your dashboard.

If the work is restricted we should say "restricted collaboration" instead of "help out"

PR #3421, opened 2022-12-13
Only show "help out" link to work collaborators for #1043

sylvieed:
_Resolves #1043_

Previously, the "help out!" link on the collection show page was shown to all users, even if the work was restricted and the user wasn't a collaborator. This PR makes the "help out!" link only show if the user `can_transcribe?` the work (if the work isn't restricted, if they're an owner of the work, or if they're a collaborator on the work). Otherwise, it'll say that "Collaboration is restricted" and won't have a link to transcribe.

 Here's how it looks, the work on the top is restricted and the work on the bottom is not.
restricted collaboration

### DIFF
--- a/app/views/collection/show.html.slim
+++ b/app/views/collection/show.html.slim
@@ -39,17 +39,17 @@
           p.collection-work_snippet = to_snippet(work.description)
           -if @collection.text_entry? && work.next_untranscribed_page
             p.collection-work_action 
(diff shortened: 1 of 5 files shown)


sylvieed:
I removed the "call to action" "Some pages still need work" from before "Collaboration is restricted". It's now only shown if you can transcribe, just like the "Help out!".

updated restricted collaboration

Issue #3430, opened 2022-12-14
i18n health upkeep


sylvieed:
Right now, `i18n-tasks` reports 17 "missing" translations (all `thredded.*`) and 128 "unused" translations (all `collection.edit_buttons.*`).

PR #3431, opened 2022-12-14
Ignore false missing & unsued translations for #3430

sylvieed:
_Resolves #3430_

This fixes the falsely reported missing and unused translations by adding them to `ignore_missing` and `ignore_unused` of `config/i18n-tasks.yml`.

The "unused" translations were reported because they're used dynamically:
https://github.com/benwbrum/fromthepage/blob/66c67e78f30ec7960428743743a2a635f18ac3f1/app/views/collection/edit_buttons.html.slim#L11-L12

The "missing" translations were reported because the translations aren't in our locale files, but they're not actually missing on the site.

### DIFF
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -107,12 +107,14 @@ ignore_missing:
     - '{!, !0, next, throw}'
     - 'deed.deed.*'
     - 'article.graph.title'
+    - 'thredded.*'
   es: &NOT_EN
     - '{!, !0, next, throw}'
     - 'deed.deed.*'
     - 'article.graph.title'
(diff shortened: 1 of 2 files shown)


saracarl:
I didn't know about this approach, but I like it!  Except it's not working.  :(

I was contemplating amending the tests to check for a number higher than 18 (? -- I think that was the number that was reported.

sylvieed:
Seems like it wasn't working because the tests were looking for 68 missing keys instead of 0.

saracarl:
Right!  That was my workaround; yours is better.  Passing now.  I'll merge.

Issue #3317, opened 2022-09-20
Owner summary reporting inaccurate


saracarl:
LVA Reports:

> I have one volunteer who is insisting that she completed time that is not captured in her active minutes export on Aug. 15. If you can please take a look at the attached and let me know if you see anything I have missed, that would be very appreciated. All her work is in LVA collections and should be visible to me/ recorded and exported. Also, is the FTP clock on GMT? She insists that all her volunteer time was done on the 15th. I'm not sure there's anything more to uncover, but promised to contact the FTP developers with this issue.
> 
> This same volunteer had previously completed several volunteer hours remotely and not had any issue with the amount of time that was recorded for her. I advised her to save more frequently to ensure that the window/browser tab remains active or join us at an event to ensure that she receives "full credit."
> 
> Let me know if have any insights into time from this volunteer or advice I can pass along to her.
> 

![image](https://user-images.githubusercontent.com/708566/191292326-c12b2538-1b9f-4c37-9ded-1953b450487f.png)

![image](https://user-images.githubusercontent.com/708566/191292579-fce68abd-4092-4c9d-8eea-dc82c014524f.png)

saracarl:
> I'd be frustrated, too, if I was the volunteer and felt like I was being shorted.  I've investigated this and here's what I found:
> 
> If I use the owner dashboard -> summary tab I get the same numbers as you show.
> BUT....  If I got into the Virginia Untold: Free Registers (Indexing) collection and pull the stats from the collaborators there, I show a lot more time -- 4 hours 42 minutes.  I dug into the collection level data some more:
> 
> 1)  Yes, we track time in GMT, which is 4 hours earlier than EST, so if you work over a date line the time logged will roll into the next day.  
> 2)  I pulled the "Export Activity as CSV" from the collaborators tab for Aug 15-17.  (To reproduce, click the button right above "Active Collaborators" [here](https://fromthepage.com/lva/virginia-untold-free-registers-indexing/collaborators?collection_id=%23%3CCollection%3A0x00007fcdfb0cc008%3E&start_date=Aug+15%2C+2022&end_date=Aug+17%2C+2022&commit=Update))
> and then I pulled just Lois's activity out of it and translated (manually!) the GMT times into EST times:
![image](https://user-images.githubusercontent.com/708566/191293294-bc077531-2da5-4111-8110-cafceb7d93f4.png)

> Based on that, she definitely worked over the date line on the night of 8/15.
[...]

PR #3449, opened 2022-12-21
Refactor owner summary contributors to use deeds for #3317

sylvieed:
_Resolves #3317_

Previously, the owner dashboard summary tab "Collaborator Time" section was somewhat inaccurate as it was pulling from the `AhoyActivitySummary`:
https://github.com/benwbrum/fromthepage/blob/fa6b500e4aae850bc7918df4aa2fa53696ba7a83/app/controllers/dashboard_controller.rb#L109-L113

instead of from deeds, like the "Collaborators" tab of a collection does, which is more accurate.

This PR refactors the owner summary collaborator section to work the same way that the collection collaborators tab does. The new code for this was put into the `owner_statistic.rb` model. The code is based on part of the [`new_contributors` method of `contributor_helper.rb`](https://github.com/benwbrum/fromthepage/blob/development/app/helpers/contributor_helper.rb#L29-L60), which gets the contributor data for the collection collaborators tab.

This changes the times shown on the summary tab as well as the times shown in the detailed activity CSV export.

Here's how it looks (should look the same as before), this is the info it's giving me for the past 6 days with my local data.
collaborator times

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -99,18 +99,8 @@ class DashboardController < ApplicationController
     @subjects_disabled = @statistics_object.collections.all?(&:subjects_disabled)
 
     # Stats
-    owner_collections = current_user.all_owner_collections.map{ |c| c.id }
-    contributor_ids_for_dates = AhoyActivitySummary
-        .where(collection_id: owner_collections)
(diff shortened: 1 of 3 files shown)


saracarl:
Yay!  This worked when I loaded the data I used to create the original issue with and ran an owner summary report for August 15-17.  I got the same time stamps as the individual collection gave us.  Whew!

PR #3451, opened 2023-01-02
Cache recent deeds in collaborator dashboard / watchlist for #3439

sylvieed:
_Resolves #3439_

This should help performance issues with the watchlist page by caching the recent deeds listed for each collection. 

### Changes made

* Moved the recent deeds shown for each collection on the watchlist to a partial, `collection/_recent_deeds` to make it simpler to cache.
* Added a `most_recent_deed_created_at` column to the `collections` table to invalidate a collection's cache when a deed is added.
  * The migration updates this field in all pre-existing collections to when their most recent deed was created.
* Added a `most_recent_deed_created_at` method for document sets, which references its collection. 
  * Although a docsest might have a different most recent deed than its collection, the `most_recent_deed_created_at` field is updated by the new deed and a `Deed` doesn't know what document set it's in; it only knows which collection (`deed.collection` is always a `Collection`, never a `DocumentSet`). 
* Cached the `_recent_deeds` partial based on (invalidated by an update in) each collection's `most_recent_deed_created_at` column.

### DIFF
--- a/app/models/deed.rb
+++ b/app/models/deed.rb
@@ -17,6 +17,7 @@ class Deed < ApplicationRecord
   visitable class_name: "Visit" # ahoy integration
 
   before_save :calculate_prerender, :calculate_prerender_mailer, :calculate_public
+  after_save :update_collections_most_recent_deed
 
   def deed_type_name
     DeedType.name(self.deed_type)
@@ -51,5 +52,9 @@ class Deed < ApplicationRecord
         renderer.render(:partial => 'deed/deed.html', :locals => { :deed => self, :long_view => true, :prerender => true, :mailer => true, locale: locale })
       ]
     }.to_json
+  end
+
+  def update_collections_most_recent_deed
+    self.collection.update(most_recent_deed_created_at: self.created_at)
   end
 end
--- a/app/models/document_set.rb
+++ b/app/models/document_set.rb
@@ -299,4 +299,8 @@ class DocumentSet < ApplicationRecord
   def api_access # API access is only controlled by public/private for document sets
     false
   end
+
+  def most_recent_deed_created_at
+    self.collection.most_recent_deed_created_at
+  end
 end
--- a/app/views/collection/_recent_deeds.html.slim
+++ b/app/views/collection/_recent_deeds.html.slim
@@ -1,0 +1,9 @@
+-collection.deeds.includes(:page, :user, :work).limit(5).each do |d|
+  -d.collection = nil
+  .user-bubble
+    =link_to(user_profile_path(d.user))
+      =profile_picture(d.user)
+    .user-bubble_content
+      =time_tag d.created_at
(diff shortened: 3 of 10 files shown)


sylvieed:
The tests failed because after an IA import, the deed had no collection because the work had no collection when the deed was created. I couldn't figure out why the work had no collection so I just added a simple check that the deed had a collection to remove the error. 
However, after the IA work is imported, the deed is never linked to the collection so the collection's `most_recent_deed_created_at` field isn't updated & this deed wouldn't be shown on the watchlist page.

benwbrum:
How did you test the caching behavior?  I don't see any evidence that a cache is actually being used on my local machine.

sylvieed:
At the top of the `_recent_deeds` partial, I added `="Cached #{Time.now}"` and saw that when I refreshed the page, the time was not updated. When a deed was added to one of the collections, that timestamp would update, but not the other collections'.

Maybe make sure caching is enabled in development with `rails dev:cache`?

benwbrum:
I see -- my problem was that I was looking in `tmp/cache` for cached template files, but development uses a memory cache store that doesn't write to the filesystem.

saracarl:
Testing with 2 collections in production:

MSA 1974 Marriage Certificates. 
Before caching:
42367.7 ms to load the deeds
43850.0 ms total page load response duration

Julia Brumfield Diaries:
Before Caching:
63.9 ms to load the deeds
899.0 ms total page load response duration

saracarl:
After caching:

MSA 1974:
55091.4 ms to load the deeds
213.0 ms total page load response time

Julia Brumfield Diaries:
12.2 ms to load deeds 
152.0 total load response time

saracarl:
For MSA, we're still hitting the database -- here's the 4th load of their collection page:



Rendering: deed/_deeds.html.slimT+1140.3 msReader50956.9 ms | app/views/deed/_deeds.html.slim:4:in `_app_views_deed__deeds_html_slim___2754572334976485685_149460' app/helpers/application_helper.rb:116:in `deeds_for' app/views/collection/show.html.slim:138:in `_app_views_collection_show_html_slim__1809131914721346880_144840' app/controllers/application_controller.rb:62:in `switch_locale' SELECT `deeds`.* FROM `deeds` INNER JOIN `users` ON `users`.`id` = `deeds`.`user_id` WHERE `deeds`.`collection_id` = 1130 AND `users`.`deleted` = FALSE AND (deed_type IN ('page_trans','page_edit','markd_blnk','art_edit','ocr_corr','review','pg_revd','xlat_rev','xlat_revd','coll_act','coll_inact','coll_join','pg_xlat','pg_xlat_ed','md_desc','md_edit')) ORDER BY deeds.created_at DESC LIMIT 20;
-- | --
19.60 ms | Rendering: deed/_deeds.html.slim — 11.80 ms






We're also running the same sql for Julia Brumfield (which looked faster, but....)  

I'm going to add this to the issue and we'll keep working it there.

Issue #3011, opened 2022-02-16
configurable welcome email by project or account


saracarl:
Spinning a bigger feature off of #3010, what I'd really like to do is let you configure a welcome email to be sent to any new signup that came through your landing page or collections. And/or new transcribers to your collections. That's a bigger deal, but would be nice.

Issue #3010, opened 2022-02-16
Send email upon signup


saracarl:
Thom at Mt. A says they do a lot of outreach on social media, and that tends to engage people on their phones.  They'll sign up for an account from their phones, but need to find it again when they are in front of their computers.

Basically, if you sign up with a specific collection or organization, send them an email that includes a link to sign back in from their computer and will take them there.

saracarl:
more input from Thom:
Found an example!

https://www.hubspot.com/email-signature-generator
ˆIf you view this article from a phone it give you a different pop up prompt saying "(Article Title) isn't available on mobile yet. We hope you'll try it out on a larger screen: Email Me the Link"

saracarl:
From our analysis of referrer success, the 11 referrers from m.facebook.com resulted in no transcribed pages (in Feb):
http://m.facebook.com/	11	0	0.00%

saracarl:
I just tried this (went to a collection page on my phone, wasn't logged in.  Tried safari, chrome, and chrome incognito), and couldn't get it to work.  Not sure what's going on, but I'd like @sylvieed to investigate (we can help) so we can get it working.

sylvieed:
Odd -- when I test this in production, it doesn't work, but testing the same way in development _does_ work. I wonder if it's something with the `mobile_device?` helper function:
https://github.com/benwbrum/fromthepage/blob/e84bbc4e92e9fc0e82e56dd798d6c112bca07a56/app/helpers/application_helper.rb#L262-L264
that doesn't work in production? Is there any way to test this? I can't see anywhere else that could be different in production.

PR #3467, opened 2023-01-24
Added popup to send an email to mobile users that join a collection for #3010

sylvieed:
_Resolves #3010_

### The problem

If a transcriber signs up to FromThePage on their phone and then wants to later come back to it on their computer, they have no easy way to find their account or the collection they joined again.

### The solution

This PR adds a popup to be shown after a user signs up through a collection that asks them if they'd like to be emailed a link to the collection. Here's how it works:

1. Someone (future user) gets referred to a collection's page on their mobile device
2. They sign up to FromThePage (and this generates a `collection_joined` deed)
3. Only if they are on a mobile device, a popup shows up (on the collection show page) saying that transcribing is easier on a larger screen, with a button to email them a link to the collection.
mobile device popup
4. If they click this button, they're sent the `new_mobile_user` email, which has a link to start transcribing that collection.
mobile device email
5. They're returned to the collection show page, with a green "Email sent." flash.

### Questions / Things to maybe change
* Right now, the popup only shows up if the user joined through a collection. Should it always come up?
* Would it be better to just always email users after they join a collection (#3011) instead of this mobile only & popup combo?
* I wasn't sure exactly what y'all wanted the popup & email to say. Let me know if you have any other wording in mind for these.

### Mobile layout issue

I tested this on my phone ([with ADB](https://stackoverflow.com/a/46795769/8326554)) and the popup actually looks really bad on mobile:
<p float="left" align="middle">
  Screenshot_20230122_214718_Chrome
  Screenshot_20230122_214725_Chrome
</p>

I'm not sure how to fix this because I don't really know how  `litebox`es work, it seems like all `litebox`es are doing this on mobile (ex: the API key popup looks very similar to this one above).


app/views/collection/show.html.slim
```diff
@@ -138,3 +138,13 @@
...
+
+=link_to(collection_new_mobile_user_path(@collection.id), :data => { litebox: { hash: 'new-mobile-user' }}, id: 'new_mobile_user_link', style: "display: none")
+  span =""
+
+-if @new_mobile_user
+  -content_for :javascript
+    javascript: 
+      $(document).ready(function(){
+        $('#new_mobile_user_link').trigger('click');
+      });
```
sylvieed, line 150: This is how I got the popup to show up only when `@new_mobile_user` is true, but this code seems weird & bad practice to me. Do y'all have any insight into a way to do this better?

app/views/collection/new_mobile_user.html.slim
```diff
@@ -0,0 +1,3 @@
+.litebox-embed(style="width:550px")
```
sylvieed, line 1: Maybe this set `width` attribute is why the `litebox` looks weird on mobile

sylvieed:
The tests are failing because of missing i18n translations, but I don't want to translate the new messages until we're sure about them, so we can ignore the failing tests for now.

app/views/collection/new_mobile_user.html.slim
saracarl: Can we set the width narrower in this instance?

saracarl:
I really like the pop up message, so let's keep that.
The mail message, though, could use more context.  Let's add the organization name (collection.owner) in the email "Milner Library's WWI Letters Collection" to remind them of what this is.  
Is there any way to sign it with the organization name like we're signing the notes emails (with a reply-to for the organization as well)?

While it would be lovely to do the custom welcome email.... hmmm....  I guess we could use the above as the default value and let customers change it in the collection settings tab.  It's just a text field there.  I'd be tempted to finish up this one, but then immediately go work on that one since it's in the same code and won't take as much re-familiarizing yourself.  

And if we read the mail content out of the DB we wouldn't have to translate it.  Project owners could if they wanted to.

saracarl (changes requested):
See comment on PR

sylvieed:
Changes made:
* The email is signed by the institution now
email

* The email subject was changed to "Owner's Collection name Collection" -- is this what you meant you wanted, or should the subject still have "Welcome to ..."? 
* I changed the width of the popup to `300px`, which looks much better in my mobile testing.

better popup
However, this isn't responsive, so it might look worse on other phones. It looks okay in landscape:

landscape

saracarl:
@sylvieed I just tried this -- in order to take a screenshot -- and couldn't get it to work.  Any ideas why?  Want to try it yourself when you have time?

PR #3471, opened 2023-01-26
remove deed list from owner profile

saracarl:


benwbrum:
I'm not sure I care for the way that the project listing has expanded to fill the space of the aside now that the deed list is removed.  Can we re-add the `aside.sidecol` element with nothing inside it (except maybe an empty paragraph)?
Screenshot from 2023-01-27 13-11-40

Issue #3415, opened 2022-12-05
login as transcriber


saracarl:
Add a guest login to the admin dashboard in the transcriber view.  Copy it from the owner view.

ajistrying:
I'll take a crack at this one.

saracarl:
@ajistrying if you send me your email address I can share a pruned database with a lot of data with you.  sara@fromthepage.com

ajistrying:
@saracarl That'd be great, my email is wjohnson@eleudev.com, thank you!

ajistrying:
Hoping I can still get that pruned db when you get a chance, looking to tackle another issue in the meantime 👍

ajistrying:
@saracarl I was able to get the dmp file and get it into the db, but I believe that data might not be as up to date as the shape of the database currently stands today. When I use that dmp file and query the users table for example, some fields seem to be missing (`undefined method` errors being thrown within the console). Like for instance when signing up a fresh account, the app gets snagged on an initial check for the `preferred_locale` field on the user model. The `preferred_locale` field wont be available on the model from the data of that dmp since its from 08-20-2020, but the addition of the `preferred_locale` field was added later that year on 11-28-2020.

Totally get that you'd need to give me an older more pruned down version of the db, but any help on addressing this would be awesome, thanks!

benwbrum:
Have you tried doing a `rake db:migrate` yet?

ajistrying:
@benwbrum I did, I'll caveat all of this by saying that I'm running the app within a docker container. 

The steps I took within the container were:
1. Run `rake db:migrate`
2. At this point all migrations are up and good to go
3. Load in the .dmp file via `docker exec -i [container_name] mysql -ufromthepage -pfromthepage diary_development < path/to/dmpfile`
4. When I run `bundle exec rails db:migrate:status` I see that some migrations have been rolled back, while some are still up, so I believe utilizing this particular dmp file may be rolling some migrations back, but I'm not 100% sure

Of course I'm down for any other way to do this, developing with docker might not be the best way to go here! Thanks

benwbrum:
Ah!  I see the problem.

The mysql dump file--if you look inside it--is a series of commands that drops all the tables from the original DB, then recreates them, then loads the data.  Running it after a db:migrate will leave you with inconsistent data -- you'll have old versions of all the tables we knew about when the dump was made, and new versions of any tables that we didn't know about. 

[...]

PR #3519, opened 2023-02-24
login as transcriber addition

ajistrying:
Definitely let me know if the order of the links needs to be different or if this needed to be approached another way, thanks!

For #3415


saracarl:
This works as expected and passes tests.  Merging.  (and thanks!)

PR #3526, opened 2023-03-07
add lastUpdated time to work status in iiif collection manifest

saracarl:


benwbrum:
If I add a note to a page, the last update time in the API is not updated.

app/models/deed.rb
```diff
@@ -59,4 +59,11 @@ def update_collections_most_recent_deed
       self.collection.update(most_recent_deed_created_at: self.created_at)
     end
   end
+
+  def update_works_most_recent_deed
```
benwbrum, line 63: This looks like the correct logic, but there is no callback that actually fires this.  (See line 20 for an example.)

benwbrum (changes requested):
Add callback to fix non-updating behavior.

PR #3536, opened 2023-03-10
german messages

saracarl:


saracarl:
@sylvieed -- look, another language!  I used your documentation and it was super fast.

sylvieed:
Great! 🎉

Issue #3329, opened 2022-09-27
"export everything" checkbox


saracarl:
> I try to export regularly so we have a hardcopy - is there a way to export everything more easily than checking every box?

Perhaps add a checkbox -- on the bottom or top -- of the export all screen that when you check it auto-checks all the options in some smart default configuration.

benwbrum:
Closed by #3539

benwbrum:
@saracarl This is the origin of the usability problem you noted with GRI

PR #3539, opened 2023-03-13
#3329-Button Added To Check All CheckBox

symmetrically:
Screenshot from 2023-03-13 21-28-07

@benwbrum I have added a button for the check boxes, if It looks good to you then we can proceed with this or If you still want checkbox instead of button then that can be replaceable too.


saracarl (changes requested):
@symmetrically the button is fine, but a couple of UI changes:

- Because it goes with "Export Formats" rather than the "Folder Organization" toggle, let's put it in line with "Export Formats"
- Let's style it the same as "Export All Works" on the Export tab:
image
Just without the icon
- We made changes to this same screen last week; please make sure that while "One file per work" for Text/DOCX gets checked, it's sub-option "Preserve linebreaks" doesn't. (I haven't tested this, just flagging it.)

saracarl (changes requested):
Functionality seems fine, but the UI still needs some changes.

1)  I think you copied the entire button, not just the styling, because the words changed to "Export all works" which isn't right.  "Check all" was better:
image

2)  I do like that you can click to check all and click again to uncheck all.  But the UI doesn't really indicate that.  I think the words should change to "uncheck all" once it's been clicked once.

3)  I don't like the green once the button has been clicked.  Just a word change would be more visually appealing.

Instead of doing all these changes would it make more sense to do a checkbox with the same sort of functionality?  I do think it would look better.  Let me find an example of where do this already and add it to this PR.

symmetrically:
@saracarl What I have done now is I have added a checkbox and when checked it will say "uncheck all" and when not checked it will say "check all", as we will not dive in UI changes here. Hope it make sense :)

saracarl:
@symmetrically if you import the following IIIF collection manifest into a FromThePage collection, you'll see the checkbox version of "check all" in the UI:  `https://fromthepage.com/iiif/collection/5` and here's where that code is:
https://github.com/benwbrum/fromthepage/blob/development/app/views/sc_collections/explore_collection.html.slim#L48

saracarl:
Looks perfect, merging!  Thanks for making the checkbox change.

PR #3554, opened 2023-03-16
Enhanced sorting for numeric and alphanumeric titles

### LINKED ISSUE #2809: Upload should order pages numerically instead of alphabetically
SLNSW would like the option to have pages ordered by the numeric value of filenames rather than the alphabetic ordering, so that they don't need to zero-pad their filenames.  This is particularly important because they are using the filename as an identifier in their catalogs, so renaming them pre-upload will present integration challenges.

symmetrically:
Closes #2809

### DIFF
--- a/app/controllers/display_controller.rb
+++ b/app/controllers/display_controller.rb
@@ -48,12 +48,7 @@ class DisplayController < ApplicationController
         @count = @pages.count
         @heading = t('.translations_need_indexing')
       else
-        @pages = @work.pages
-        numeric_pages, alpha_numeric_pages = @pages.partition { |page| page.title.to_i.positive? }
-        sorted_numeric_pages = numeric_pages.sort_by { |page| page.title.to_i }
-        @pages = sorted_numeric_pages.concat(alpha_numeric_pages)
-        @pages = Page.where(id: @pages.pluck(:id)).order("FIELD(id, #{@pages.pluck(:id).join(',')})")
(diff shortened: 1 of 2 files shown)


symmetrically:
@benwbrum I opted for the approach in which at the time of `read_work`(Display Controller) it will sort the pages based on the title of the page. The pattern would be numeric, then alphabets and then alphanumeric. I only did these changes for the else condition when `params[:needs_review]` is `nill` for the Read Tab.
Let me know what you think about this approach

[Screencast from 03-16-2023 09_18_38 PM.webm](https://user-images.githubusercontent.com/829262/225699226-a7f808b8-e91a-4da5-8c0d-8d99bc3b949c.webm)

benwbrum:
I don't think that ording on display is going to work, since part of the application allows project owners to manually change the order of pages within a work.  (Click on a work you own; click the Pages tab)

Screenshot from 2023-03-22 13-58-56

This is a feature which is not used often, but it is important for the few projects that use it.

So I think that we should continue to order based on the `acts_as_list` position when we display pages in Read Work or Contents tabs, but use numerical order (as appropriate) in the Ingestor's code which converts uploaded files to works.

benwbrum (changes requested):
See comment on PR.

symmetrically:
@benwbrum With the new changes, it will sort the pages numerically before creating them in the DB. I have tested it locally too, its working as expected now.

benwbrum (approved):
This is better logic than what I came up with.

Issue #3534, opened 2023-03-10
CodeMirror support for RTL languages


benwbrum:
FromThePage should support transcribing text in any script, but not all features work well for RTL projects (currently Arabic and Urdu).

When we implemented CodeMirror as our advanced editor, we ran into some configuration problems with RTL support.  (Codemirror documentation says that it supports RTL scripts, but we had difficulties when we tried.)  As a result, we took the [shortcut of disabling Codemirror for RTL scripts](https://github.com/benwbrum/fromthepage/blob/66c67e78f30ec7960428743743a2a635f18ac3f1/app/views/transcribe/display_page.html.slim#L106-L108).  In the last two months, one user has complained that the syntax highlighting and auto-complete functionality is not available for Arabic, so this decision is affecting users.

We should do the following:
* Remove the RTL test in `app/views/transcribe/display_page`
* Fix any bugs for Arabic or Urdu users -- change the "text language" in the Settings tab for a work to a RTL language, then test transcribing and editing a document in that script.  Watch for justification as well as odd behavior.
* Add editor buttons to the project (Collection->Settings->Configure Buttons) then try to add XML mark-up to text.  Watch for strange behavior on the cursor and other things that "feel wrong" to someone used to editing HTML+RTL text.

symmetrically:
@benwbrum From the description what i understand is that if i remove the given [check](https://github.com/benwbrum/fromthepage/blob/66c67e78f30ec7960428743743a2a635f18ac3f1/app/views/transcribe/display_page.html.slim#L106) then code-mirror functionality will work for rtl too. But the issue is that for the transcribe tab the cursor direction is on the left side instead of the right side of the editor.
Let me know if I understand the problem correctly?

![Screenshot from 2023-03-20 23-26-02](https://user-images.githubusercontent.com/829262/226594731-188e3790-0b4b-46c1-8783-7a47eb9fd5b1.png)

benwbrum:
I see two problems here -- the cursor starts on the left side instead of the right (as you point out), and also the text is justified against the left side instead of the right.

PR #3560, opened 2023-03-21
#3534-Configured CodeMirror for RTL

symmetrically:
[Screencast from 03-21-2023 10_33_00 PM.webm](https://user-images.githubusercontent.com/829262/226751147-3ca713c2-cf5e-4ff2-8488-e54ebb682d43.webm)

### DIFF
--- a/app/assets/stylesheets/codemirror/lib/codemirror.css
+++ b/app/assets/stylesheets/codemirror/lib/codemirror.css
@@ -5,7 +5,8 @@
   font-family: monospace;
   height: auto;
   color: black;
-  direction: ltr;
+  text-align: justify;
+  padding-right: 18px;
 }
 
 /* PADDING */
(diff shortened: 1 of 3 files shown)


benwbrum:
This looks spectacular!

We need to verify that this does not break field-based projects, however, since I see that a check for that was removed.  To test this, see [this project](http://localhost:3000/ushmmarchives/fielded-test/diaries-1943-1944/transcribe/554945), which uses field-based transcription instead of CodeMirror, as in the screenshot.

Screenshot from 2023-03-21 17-34-14

symmetrically:
@benwbrum Tested locally, not found any unusual behavior for the field-based transcription
[Screencast from 03-22-2023 04_31_58 PM.webm](https://user-images.githubusercontent.com/829262/226918795-7061b9c1-be03-424b-add3-dcc1595a193d.webm)

Issue #3544, opened 2023-03-13
Project owner cannot create a new custom messageboard


anabeabas:
David Judge of State Library of NSW emailed us to say that they cannot create a new custom messageboard despite being logged in as the project owner.

Issue #3543, opened 2023-03-13
Project owner cannot see the "delete a topic" button in the messageboard


anabeabas:
David Judge of State Library of NSW said that after creating a topic in a messageboard, they cannot see the delete a topic button.

Issue #3542, opened 2023-03-13
Owners can see messageboard settings for collections they do not own


anabeabas:
David Judge of State Library of NSW emailed us to say that they can see messageboard settings for many collections, including those which they do not own.

![image](https://user-images.githubusercontent.com/98731543/224783154-d1c0a854-8f3b-4277-9901-99f7e943cddc.png)

PR #3563, opened 2023-03-22
3541 multi tenant forum fixes

### LINKED ISSUE #3541: Multiple messageboards appearing on messageboard dropdown
David Judge of State Library of NSW emailed us to say that they can see multiple Messageboard options from the Messageboard dropdown. They are seeing (General, Help) from many collections.

image

benwbrum:
Closes #3541 #3542 #3543 #3544


saracarl (approved):
The only problem I spotted here was the confirmation message on deleting a topic shows up -- and has to be clicked -- twice.  I'll open a new issue for that.  (but I wouldn't be surprised if David notices the same thing.)

Issue #3570, opened 2023-03-23
Document Uploads with bad metadata.yaml fail silently


benwbrum:
Sabrina at ZIM uploaded two zip files with extensive `metadata.yml` files which violated some syntax rules.  Instead of emailing her a problem or explaining that there was an error in the upload, the `document_upload` got stuck in the status `Processing`.

We should handle bad YML parsing more gracefully than that, resolving to an error state and explaining the problem (if possible) rather than leaving the user in limbo.

[Problem upload file: MA-Rezepte.zip](https://github.com/benwbrum/fromthepage/files/11052161/MA-Rezepte.zip)

benwbrum:
To duplicate:

- Upload the attached file.
- Observe that the upload stays in the processing state

We need to catch the exception, print something useful in the logfile which is emailed to the user, and update the `document_upload.status` to Error

symmetrically:
@benwbrum 
I can see that when we catch the exception, it creates the work object. So I have the following questions:

1. When we catch the exception, do we need to create the work for the uploaded file?
2. Apart from printing to the logfile, do we need to write mailer for this or is this already done?

benwbrum:
1. We should not create a work for the uploaded file if there is an exception
2. I think that we only need to print an error to the logfile.  (I think we may already do this for some other exceptions, but I might be mistaken.)

PR #3587, opened 2023-03-29
#3570 Rescue document upload metadata fail

symmetrically:


lib/tasks/ingestor.rake
```diff
@@ -209,16 +209,21 @@ namespace :fromthepage do
...
+        yaml = YAML.load_file(File.join(path, 'metadata.yml'))
+      elsif File.exist? File.join(path, 'metadata.yaml')
+        yaml = YAML.load_file(File.join(path, 'metadata.yaml'))
+      else
+        print "\tconvert_to_work no metadata.yml file; using default settings\n"
+        yaml = nil
+      end
+    rescue StandardError => e
+      document_upload.update(status: DocumentUpload::Status::ERROR)
+      logger.info "Yml/Yaml Failed: Exception: #{e.message}"
```
benwbrum, line 223: It's a minor thing, but let's use `YML/YAML`, since YAML is an acronym for Yet Another Markup Language.

app/models/document_upload.rb
```diff
@@ -11,6 +11,7 @@ module Status
     QUEUED = 'queued'
     PROCESSING = 'processing'
     FINISHED = 'finished'
+    ERROR = 'error'
```
benwbrum, line 14: Is this used anywhere?  Can we set the value on `document_upload.status`?

benwbrum (changes requested):
Clarifying question, request for capitalization change.

app/models/document_upload.rb
symmetrically: Yes, you can see it on line #222 of the `lib/tasks/ingestor.rake` file

Issue #3468, opened 2023-01-24
Account/user wide footer setting


anabeabas:
Add a feature that lets project owners put the same footer across all collections.

benwbrum:
There are a handful of attributes which are stored on the `Collection` model that we have learned should be identical across all of a user's collections.  The `footer` is displayed at the bottom of each page on a user's projects, and is used almost exclusively for account branding and links to the institution.  Project owners would prefer to configure this once, in their profile, instead of needing to copy-and-paste the same HTML into the `footer` attribute for each collection.

We should move `footer` from `Collection` to `User`, make it editable (for project owners only) in the profile, migrate data to copy the `footer` contents from one of their collections (ideally the one with the longest `footer` contents) into this new field on `User.footer`, then display that in the application layout instead of displaying the collection footer.

symmetrically:
@benwbrum 
What I understand is that currently there is an attribute named footer_block in the Collection Modal which is used to display the footer for that collection(see Image below).  Now, I have to create a new attribute in the User Modal (let's say named footer_block) and copy the longest footer_block from the collections of that user and assign it to the newly created attribute in the User Modal. Then use that footer_block from the user Model, right?

![Screenshot from 2023-03-27 20-19-22](https://user-images.githubusercontent.com/829262/227991514-a9e08351-43a6-46f3-987c-2eadf20a4bb6.png)

symmetrically:
@benwbrum 
As per the requirement, I have added a new attribute in the user modal and moved the `footer_block` logic in the application layout. The question is should I remove the `footer_block` Column from the Collection Modal and the respective files?

benwbrum:
Yes, please remove `footer_block` from `Collection` -- for this attribute we want everything driven by the value in `User`.

PR #3589, opened 2023-03-29
#3468-Move footer_block to User Modal and Added Rake Task

symmetrically:


benwbrum:
Here's what I'm doing to diagnose the problem

* Check out the branch locally
* Run ` RAILS_ENV=test rake db:drop db:create db:migrate && RAILS_ENV=test rake db:fixtures:load FIXTURES_PATH=spec/fixtures&& rspec --fail-fast`
* Observe that RSpec reports the first failure at `rspec ./spec/features/collection_metadata_spec.rb:9 # collection metadata creates a collection as owner`
* Scroll up through the Capybara exception to find the failure at `     # ./spec/features/collection_metadata_spec.rb:13:in `block (2 levels) in <top (required)>'`
* Add a `binding.pry` after `spec/features/collection_metadata_spec:12`
* Run the test suite again
* When the test drops into the pry console, print the page text:
```
[3] pry(#<RSpec::ExampleGroups::CollectionMetadata>)> print page.text
NoMethodError at /collection/new ================================ undefined method `footer_block' for nil:NilClass 
```

So it looks like we need to take a look at the calls to `@collection.footer_block` throughout the application.  An alternative (and easier) strategy would be to create a new `footer_block` method on `Collection` which returns the collection owner's `footer_block` attribute.

symmetrically:
@benwbrum I have tried the same what you have specified above, Issue is this error does not link to the current fix and if you checkout a branch from development you will also get the same or you can run test in development you will get it there too

benwbrum (changes requested):
I have to say how much I love that this removes so much redundant code.

Do we need an extra migration to remove `collections.footer_block` as well?

symmetrically:
@benwbrum Yes, after the running the  `copy_footer_block.rake task`, we need a migration to remove the `collections.footer_block`

benwbrum:
I've fixed the merge conflicts and started another test run.  Let's see if this one passes.

benwbrum:
@symmetrically Please move the data copying logic from the rake task into the migration.

benwbrum:
I've fixed one test failure which should clear out a few things, but in `editor_actions_spec.rb:139`, there is a value for `@collection.footer_block` but no value for `@collection.owner.footer_block`.  I suspect that the fixtures might need to be migrated as well.

symmetrically:
@saracarl 

1. what I understood that i have to set default footer to each project owner profile that reads "For questions about this project, contact at ."  footer column in user model
2. In the collection previously I have removed the collection footer block  now I have to add it again but this time it should non-editable and there will have a link too which will take to the user profile setting if the `user == collection.owner`  else show flash that not the owner

Please let me know if I understood correctly or if you have any further clarifications.

benwbrum:
On 1, that's correct.
On 2, what we need to do is fix old test data which still has a fixture that matches the old schema.  

See https://github.com/benwbrum/fromthepage/blob/3468-account/user-wide-footer-settings/spec/fixtures/collections.yml#L6
which probably needs to move to
https://github.com/benwbrum/fromthepage/blob/3468-account/user-wide-footer-settings/spec/fixtures/users.yml#L61
since the collection 1 is owned by user 5 in the fixtures.

Issue #3575, opened 2023-03-27
Metadata description undefined method +


benwbrum:
When viewing http://localhost:3000/dartmouthlibrary/wrangelisland/stefansson-wrangel-09-07-089/describe (or in production) the server sends a 500 error with the log message
```

F, [2023-03-27T08:27:16.298116 #8419] FATAL -- :   
ActionView::Template::Error (undefined method `+' for nil:NilClass):
    36:             =svg_symbol '#icon-arrow-left', class: 'icon', title: 'Previous page'
    37: 
    38:           span.page-nav_info = t('.metadata')
    39:           -next_works = @collection.works.to_a[index+1..]
    40:           -next_work = next_works.detect {|candidate| candidate.pages.where(status: [nil, Page::STATUS_INCOMPLETE]).first || candidate.description_status == Work::DescriptionStatus::UNDESCRIBED }
    41:           -if next_work
    42:             -next_page = next_work.pages.where(status: [nil, Page::STATUS_INCOMPLETE]).first
  
app/views/shared/_description_tabs.html.slim:39
app/views/work/describe.html.slim:3
```

benwbrum:
A similar error:

```

F, [2023-03-27T09:19:05.970029 #8419] FATAL -- :   
ActionView::Template::Error (undefined method `can_review?' for nil:NilClass):
    65:           -else
    66:             =button_tag t('.done'), :name => 'save_to_transcribed', type: 'submit', id: 'finish_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.finish_to_transcribed_tooltip')}", class: "bggreen"
    67: 
    68:         -if @collection.review_workflow && @work.description_status == Work::DescriptionStatus::NEEDS_REVIEW && current_user.can_review?(@collection)
    69:           =button_tag t('.approve'), :name => 'approve_to_transcribed', type: 'submit', id: 'approve_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.approve_to_transcribed_tooltip')}", class: "bggreen"
    70: 
    71: 
  
app/views/work/describe.html.slim:68
```

benwbrum:
This works great!  Let's also hide the metadata tab if users are not logged in.

PR #3596, opened 2023-04-03
#3575 Metadata description undefined method fix

symmetrically:


benwbrum:
I did not describe this issue adequately.  If a user is not logged it, viewing the metadata description action should show them the metadata overview.  

This conditionalization fixes the original 500 error displaying the metadata description form, but if a site visitor hits the save button on the form, they will get a different 500 error.  We should not display the form to a user at all unless they are not signed in.

Result of hitting "Save" when not logged in

symmetrically:
@benwbrum Fixed!

benwbrum:
This could use one more change: we should hide the metadata description tab in the `work_tabs` view if the user is not logged in.  Currently you can see a tab, but clicking on it does nothing:

Screenshot from 2023-04-19 17-15-11

Issue #2433, opened 2021-03-08
change <s> tag to <strike> tag


saracarl:
It's more intuitive

PR #3601, opened 2023-04-05
#2433-Changed s to strike for button

symmetrically:


benwbrum:
This breaks existing projects that have an `s` button.

To reproduce:
* switch to the `development` branch
* create or edit a collection
* add an editor button for `s`
* edit a page
* verify that the button `s` appears
* switch to the 2433-change-s-tag... branch
* edit a page
* see an exception
Screenshot from 2023-04-06 10-48-23

I believe that a DB migration is needed to find all `EditorButton` records with the `key` of `s` and change those keys to `strike`.

benwbrum (changes requested):
This looks good but needs a migration for existing collections with `s`.

benwbrum:
Let's do this as a migration rather than a rake task so that the data will be updated during deployment.

Issue #3597, opened 2023-04-03
transcriber login sent to landing page


saracarl:
When we watched Dreanna log into FromThePage we were surprised that when she logged in -- as a transcriber -- she was taken to the landing page.  It should be the find a project page.

@anabeabas can you test this?  create a transcriber account, where are you taken?  log out, log back in, where are you taken?  (If it's a problem a quick video would be good to show it.)

Let's also test it when coming from an organization page (fromthepage.com/lva) or a specific work or collection.  You should be taken back to where you started the first time you login/create an account.  (If this is a problem, let's open a separate issue)

saracarl:
(assign to @symmetrically if it's a real problem rather than something I'm remembering incorrectly.)

anabeabas:
@saracarl @symmetrically 

I made a transcriber account and it took me to the landing page after I signed up. It did the same when I logged out and then logged back in. Video showing it here: https://www.loom.com/share/4f5b5c02b7114753be36ded57c9e8f60

I made another transcriber account from the LVA landing page (fromthepage.com/lva) and it took me back to the LVA landing page after I signed up, so no issues there.

anabeabas:
@symmetrically I rerecorded the video that explains the issue better: https://www.loom.com/share/e1c32e4e677940f0af774e771ce4336f

PR #3604, opened 2023-04-06
#3597 Transcriber login redirect

symmetrically:


saracarl (changes requested):
Ok, so this works as we asked, but we didn't do a good job defining what needs to happen.

If I'm a transcriber and I log in, I should go to my *dashboard*:
http://localhost:3000/dashboard/watchlist

(Sometimes that dashboard is empty, but we have text that sends them to the find a project page then.)

Apologies!

symmetrically:
@saracarl I want to confirm that transcriber users are those who have owner & admin flag set to false. Right?

symmetrically:
@saracarl I have done the required changes. Please check & let me know.

saracarl:
Looks great, thank you.

Issue #3565, opened 2023-03-22
Account-level "help" configuration" feature


anabeabas:
Clayton Ruminski of USDA suggested a feature for us to make: account level "help" configuration.

benwbrum:
Most--but not all--project owners use the same `Basic Help` text for each of their collections.  Allowing project owners to edit project help in their profile (similar to their footer in #3468) would save time and effort for them.  They should still be able to create collection-specific help text, however, which would override the account-wide help text.

(See an example of similar functionality with `Collection.transcription_conventions`, which is shown on each work, but can be overridden by `Work.transcription_conventions`.)

symmetrically:
Do we have to add Subject linking help text i.e. `link_help`  column in collection modal too in the User modal? @benwbrum

benwbrum:
I don't think we need to do this with subject linking.

PR #3626, opened 2023-04-28
#3565-Added help config in user

symmetrically:


app/models/collection.rb
```diff
@@ -443,4 +443,10 @@ def set_link_help
     end
   end
 
+  def user_help
```
benwbrum, line 446: Wouldn't it be a bit simpler to override the `help` accessor to return `self[:help] || self.owner.help`, so we wouldn't need to do checks when we display?

app/views/shared/_transcription_help.html.slim
```diff
@@ -1,6 +1,9 @@
 .columns
   article.maincol
-    =@collection.help.html_safe
```
benwbrum, line 3: See my comment on the Collection model

app/models/collection.rb
benwbrum: Note that if we do this, we will need to add a mutator override that makes sure that `collection[:help]` is set to `nil` if the user tries to set it to a value identical to `collection.owner.help`.  Otherwise we might end up with this scenario:
* Owner sets the help text on their profile to `AAAA`
* Owner edits a collection and modifies the description (or some value other than the help text)
* _`collection.help` is accidentally set to `AAAA`_
* Owner sets the help text on their profile to `BBBB`
* User views the collection help, expecting to see `BBBB` but instead sees `AAAA`

benwbrum (changes requested):
I've made some suggestions, but would like to hear your opinion on whether they are a good idea, @symmetrically.

app/models/collection.rb
symmetrically: @benwbrum In the view it was `@collection.help` previously which is pointing to help attribute of `@collection` instance variable. i.e we do not have any help accessor. If I create a new class method in collection model named as help to `return  self[:help] || self.owner.help` which i tried but it makes things unnecessarily cumbersome which I think is not a good approach.

benwbrum: Fair enough.

PR #3628, opened 2023-05-03
added line about read only data to pricing page

saracarl:


benwbrum:
I no longer see a link to the pricing page on the site footer in my dashboard or on the landing page.

PR #3632, opened 2023-05-05
Use timeago library for timestamps and cache deeds on collection page for #3439

sylvieed:
_For #3439_

Previously, when we cached deeds on the dashboard watchlist, their relative timestamp ("5 minutes ago") would be cached as well and so it would be inaccurate. This uses the [timeago jquery plugin](http://timeago.yarp.com/) to load these timestamps dynamically on the client side so that we can continue caching them.

image
(the "cached" line is for testing, it's not in this PR). The relative timestamp changes from "less than a minute ago" to "about a minute ago" while still being cached & not hitting the database.


sylvieed:
The "... ago" statements now support i18n in the five languages we currently support. Here's French:
image

I did this by adding a function in the timeago plugin that sets the locale for it to use. I manually added in the locales from [their GitHub](https://github.com/rmm5t/jquery-timeago/tree/master/locales), so if/when we add more languages, we'll need to add them here too. I can update that in the Wiki when this gets merged.

saracarl:
Tested with MSA (on my test server, so no load).  Previous results (on production) were:
MSA 1974 Marriage Certificates.
Before caching:
42367.7 ms to load the deeds
43850.0 ms total page load response duration

Current (on test) are:
34.7 to load the deeds
225 ms to load the page the first time
123 ms to load the page subsequent times

I just realized that's with @benwbrum's new db index, so not an apples to apples comparison, but still blazingly fast. Yay!

saracarl:
Also confirmed the translations load nicely.

saracarl (approved):

Issue #3461, opened 2023-01-13
FDP needs italics tag 


saracarl:
Add italics mark up tag to the advanced editor.

Inform Sue and Will K. about it when it is ready.

saracarl:
Here's the advanced editor overview:  https://content.fromthepage.com/project-owner-documentation/advanced-editor/

Reference #2263 for an example of how we added a different advanced editor tag.  

It should be processed as follows:
* Display: italicize the text (see strikethrough for an similar example)
* TEI: <emph rend="italics">XXX</emph> (@benwbrum to confirm)
* Verbatim plaintext: no style
* Emended plaintext: no style
* Searchable plaintext: no style

PR #3633, opened 2023-05-05
Added italics tag for #3461

sylvieed:
_Resolves #3461_

This adds an italics tag to the advanced editor.

Here is how it looks in the editor buttons settings:
settings

Here is how it looks in the editor:
editor
And the result:
result

### DIFF
--- a/app/helpers/abstract_xml_helper.rb
+++ b/app/helpers/abstract_xml_helper.rb
@@ -175,6 +175,14 @@ module AbstractXmlHelper
       end
     end
 
+    doc.elements.each("//emph") do |e|
+      rend = e.attributes["rend"]
+      span = e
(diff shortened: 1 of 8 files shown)


saracarl (changes requested):
1. Italics should show up in the advanced editor configuration screen in alphabetical order, not at the end.

2. The html export shows:
<i rend='italics'>asd</i>
when it should be just <i>asd</i>

3.  And the TEI tag should be
<hi rend='italics'>asd</hi>
(that was incorrect in the issue; just checked with @benwbrum)

sylvieed:
Made those changes,

Editor changed to `<hi rend="italics">`
editor

Export just has `<i>`
export

Settings is in alphabetical order
settings

Issue #3475, opened 2023-01-26
block contributor by displayname


saracarl:
Aaron at NBU has a contributor who is not responding to emails or comments or corrections.  He'd like to block him, and we decided it's the only thing we can do in this scenario.

benwbrum:
We should add a new "Block User" button on the collection settings screen.  This would bring up a modal (similar to the Add Owner or Add Reviewer modals) listing users who can currently see the project and letting the project owner choose ones to block.

If a user is blocked from a collection, the collection should be invisible to them when they are logged in -- as if the collection were private.  I believe that we have a `Collection#show_to?` method that will need to consult the block list, and we may also have a scope we use to populate the FindAProject or Collaborator Dashboard screens.

PR #3642, opened 2023-05-11
#3475-Added Block User Functionality

symmetrically:


symmetrically:
@benwbrum I have added the block user functionality and have added checks to restrict the blocked user. Kindly review the code and let me know if any changes are required!

app/controllers/collection_controller.rb
```diff
@@ -284,6 +295,11 @@ def add_reviewer
     redirect_to collection_edit_reviewers_path(@collection)
   end
 
+  def add_block_user
+    CollectionBlock.create(collection_id: params[:collection_id].to_i, user_id: params[:user_id].to_i)
```
benwbrum, line 299: This action should already have a `@user` and `@collection` variable.

app/helpers/projects_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ProjectsHelper
+    def show_project_snippet?(project, user)
+      CollectionBlock.find_by(collection_id: project.id, user_id: user.id).nil?
```
benwbrum, line 3: This should already have a `@user` and `@collection` variable.

app/models/user.rb
```diff
@@ -264,6 +267,8 @@ def visible_collections_and_document_sets(user)
     # document set show_to? logic:
     #     (!self.restricted && self.works.present?) || (user && user.like_owner?(self)) || (user && user.collaborator?(self))
     public_collections = self.unrestricted_collections
+    blocked_collection_ids = CollectionBlock.where(user_id: current_user.id).pluck(:collection_id)
```
benwbrum, line 270: `current_user.id` should probably be `user.id` since that's passed in as a parameter.

app/models/user.rb
```diff
@@ -264,6 +267,8 @@ def visible_collections_and_document_sets(user)
...
+    filtered_public_collections = public_collections.where.not(id: blocked_collection_ids)
```
benwbrum, line 271: This is very nice!  Much clearer than what I'd imagined.

app/views/collection/block_users.html.slim
```diff
@@ -0,0 +1,44 @@
+.litebox-embed(style="width:450px")
+  
+  h3 =t('collection.edit.block_users')
+  p.fglight =t('collection.edit.block_users_description')
```
benwbrum, line 4: We should use a different message here, which is more verbose.  Something like:

_These users have been blocked from access to the collection. The collection will not be visible to them, as if it had been marked private or deleted.  Adding users to this list will not notify them that they have been blocked._

benwbrum (changes requested):
Some small changes in wording and code style.

benwbrum:
In manual testing, I was able to add and remove blocked users as a project owner.  However, if I sign out of the site, then go to http://localhost:3000/findaproject I get a `nil` error, the same as the test failure.

In addition, we should hide the Block Users section on the collection edit view if the collection is not public, since this functionality doesn't really apply to private collections.

app/controllers/collection_controller.rb
```diff
@@ -262,6 +262,17 @@ def remove_owner
     redirect_to collection_edit_owners_path(@collection)
   end
 
+  def remove_block_user
+    collection_block = CollectionBlock.find_by(collection_id: params[:collection_id].to_i, user_id: params[:user_id].to_i)
```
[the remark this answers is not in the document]
symmetrically: The `remove` method can be used with a `has_and_belongs_to_many` association when you want to remove the join record. In your case, `CollectionBlock` is a model that represents a join table, not a join record.
@benwbrum

app/helpers/projects_helper.rb
symmetrically: Here the `@user` and `@collection` variable can't be use and is not feasible if you review the  view `fromthepage/app/views/dashboard/landing_page.html.slim `
from where this helper method is being used you will get the idea.

benwbrum:
We're getting a `nil` exception if we hit the http://localhost:3000/collections URL.

symmetrically:
[Screencast from 05-16-2023 03_53_48 PM.webm](https://github.com/benwbrum/fromthepage/assets/829262/2fd6a26f-6b08-4de8-8b3f-b930a293cea6)
@benwbrum Its working fine

benwbrum:
The tests are failing on this same path, with this error:
```
NoMethodError at /collections 
============================= 
undefined method `id' for nil:NilClass > 
To access an interactive console with this error, point your browser to: /__better_errors app/models/collection.rb, line 171
 ---------------------------------- 
166 
167 subjects.export 
168 end 
169 
170 def show_to?(user) > 
171 (!self.restricted && self.works.present? && self.collection_blocks.where(user_id: user.id).none? 
172 ) || (user && user.like_owner?(self)) || (user && user.collaborator?(self)) 
173 end 
174 
175 def create_categories 
176 #create two default categories

App backtrace ------------- 
- app/models/collection.rb:171:in `show_to?' 
- app/views/deed/_deed.html.slim:38:in `block in _app_views_deed__deed_html_slim___45116870111345761_3532440' 
- app/views/deed/_deed.html.slim:7:in `_app_views_deed__deed_html_slim___45116870111345761_3532440' 
- app/views/deed/_deeds.html.slim:12:in `block in _app_views_deed__deeds_html_slim___2056799528183471155_122600' - app/views/deed/_deeds.html.slim:4:in
```

If I run the test environment via `RAILS_ENV=test rails s` (after running the spec tests with `--fail-fast` to load data and fail at the appropriate place), I can reproduce the error


https://github.com/benwbrum/fromthepage/assets/199961/4426d4bd-99e6-4dff-b38d-8a33d1e5b090

benwbrum:
This works like a charm.  Only one thing needs changing: please change the text of the block user button from `Block` to `Block Users`.

Issue #1948, opened 2020-08-05
next page doesn't work on autolink [regression]


saracarl:
Another one is that if you click auto link and then click next page it breaks out of the work and goes to our collections home page

PR #3653, opened 2023-05-19
#1948 Navigation Warning Without Saving

symmetrically:


benwbrum:
The navigation buttons now warn the user after hitting Autolink.  However I also learned that the unsaved changes warning seems to appear even if the user has not changed anything.  This may be related to the autolink work, or it may be pre-existing, but I'd like to fix it before we ship this.

To reproduce:
* Navigate to a work
* Select a page (see Page Overview tab)
* Click the Transcribe tab
* Click the Next arrow
* Observe the warning about unsaved work, even though you have not changed anything.

https://github.com/benwbrum/fromthepage/assets/199961/18f10728-17da-4b58-85d6-76ff4dfbea7f

benwbrum (approved):
The code looks good.  However, see the comment on the PR about always getting the unsaved changes warning (which may be unrelated to this work).

benwbrum:
I've determined that when [the code](https://github.com/benwbrum/fromthepage/blob/1948-next-page-doesnot-work-on-autolink/app/assets/javascripts/transcribe.js.erb#L125) is checking for changes in CodeMirror, `localStorage.getItem("autoLinkAndPreviewFlag")` returns `true` for me.  Apparently this does not always get cleared.

Here is how I think you can reproduce this:
* Edit a page transcript and press Autolink
* See the refreshed page with post-Autolink text
* Click a navigation link
* View the alert
* Ignore the alert -- either close the tab or reload the window
At that point, nothing has cleared the Autolink flag, so any further edits to transcripts will show the error.

One caveat that occured to me when testing this -- Autolink does not always change the text by suggesting links, if it cannot find any.  (This is usually the case for new projects that have no subjects yet.)  In those cases, we should not display the warning when someone navigates away from the page after pressing Autolink.

PR #3654, opened 2023-05-22
Refactored Commit After Merge Conflicts

### LINKED ISSUE #3468: Account/user wide footer setting

symmetrically:
Closes #3468


app/views/collection/edit.html.slim
```diff
@@ -22,8 +22,15 @@
         tr
           td(colspan="2")
             =f.label :footer_block, t('.footer'), class: "above"
-            =f.text_area :footer_block, rows: 5, value: @collection.footer_block
-            p.nomargin.settings =link_to t('.learn_more_footer'), "https://content.fromthepage.com/project-owner-documentation/footer/", target: :_blank
+            - if current_user == @collection.owner
+              - if ["Legacy", "Small Organization", "Individual Researcher", "Trial", "Large Institution"].include?(current_user.account_type)
+                = f.text_area :footer_block, rows: 5, value: @collection.owner.footer_block, disabled: true
+                p.nomargin.settings = link_to t('.edit_profile'), user_profile_path(current_user)
+              - elsif current_user.account_type == "Staff"
```
benwbrum, line 29: I just did a test with a collection which was created by Owner X, who then made user Y an owner over the collection.  Ordinarily that would make user Y staff, but user Y was already an owner on a different collection.  (It's a rare use case, but sometimes happens.)  In that case, we do not see the message:
Screenshot from 2023-05-24 10-12-19

As a result, I don't think that testing for account types is useful.  If `@collection.owner==current_user`, show them the link to the field.  If not, show them the message.

Issue #3613, opened 2023-04-12
Can't click the Export tab to go back in Bulk Export


anabeabas:
Jon Fackrell emailed us about exporting all of the Subjects with the categories. Sara told him it was moved and that he has to go to "Export" then "Export All Works", where there's a "subject index export" there. It will kick off a task and send an email with a download link. He said that he couldn't click the Export tab to go back.

benwbrum:
We should add a breadcrumb back from Export All to the main export tab, unless we address this by #3612

benwbrum:
To be addressed by #3612 instead

benwbrum:
Opening after deciding this approach was better than #3613

PR #3668, opened 2023-05-31
#3613-Added Back Button to go Back to Main Export Tab

symmetrically:


saracarl:
image

So this change introduces a 3rd UI pattern for going "back" to a previous screen.  That breaks consistency rules -- we want users to see what they expect because the application behaves similarly from one screen to another.

I reviewed the other places we navigate "down" from a object's tab to see what our other approaches are.  

When you change the "transcription type" to "field based" in the collection settings tab (basically settings --> fields) we create/make visible and pop you into a "fields" tab:


image


The review tab sub tasks uses a pattern that I think would be best for the export --> export all case:
image

So what I'd like to see is:

"Export" --> "Export All Works" 

at the top of the export all works screen, styled like the "review dashboard --> recent contributors" example above.  "Export All Works" would not be a link but "Export" would be.  (and remove the button from the bottom)

saracarl (changes requested):
See UI changes above.

Issue #3619, opened 2023-04-24
Improve owner dashboard


saracarl:
Sonya at LVA requested that we simplify the owner dashboard

- can we drop the collection description
- can we remove the works list under the collection
- can we nest document sets under the collection
- can we show if the collection is public or private
- can we show the progress of the collection

Related to #3590

Issue #3590, opened 2023-03-29
Separate list for collections marked completed


anabeabas:
Christina Miller of the Missouri State Archives emailed us to request a separate list for collections marked completed or when we “restrict a completed work.” She also asked if we could have active projects at the top of the list on the owner dashboard.

benwbrum:
Let's address this in #3619

Issue #3205, opened 2022-07-12
Improvements to the "Your Collections" page


sylvieed:
The "Your Collections" page can become cluttered and hard to understand when:
* You have a lot of collections or works
* You have long collection descriptions
* You have long work names

> There is a lot of text that is visually hard to sort through. It’s hard to see the break between one project and the next. 

An example:

<img alt="long collections page" src="https://user-images.githubusercontent.com/35716893/178527121-c388e71f-411e-4901-84fb-4f00e60d6a6c.jpg" width="50%">

### Possible solutions

* Add clearer visual distinctions between collections, such as boxes around them or clearer lines between them
* Make the collections collapsible
* Add a search feature to quickly find a particular collection or work
* Add a sort feature for displaying the most recently created work or other filters

We could implement as many as these solutions as we see necessary or helpful.

saracarl:
Christina at Missouri recently asked that we move inactive collections to the bottom.

benwbrum:
To be covered in #3619

PR #3672, opened 2023-06-02
Simplified owner dashboard and added collection progress bars for #3619

sylvieed:
_Resolves #3619 (and #3590, #3205)_

This improves and simplifies the owner dashboard.

image

### Changes
* UI updated so that collections are more easily distinguished with a border, like how works are shown on the collection show page.
* Collection description removed.
* Works list removed.
* "You can add another work under Start A Project" was previously written under work list, now removed.
* Collection privacy added next to the collection title.
* Document sets nested under the collection they belong to instead of listed at the bottom of the dashboard.
* Inactive collections listed separate from active collections, at the bottom of the dashboard.
* Collection progress % completed statistics added (% completed, transcribed, reviewed, indexed).
* Collection progress bars added.

The logic for these last two statistics are based on the corresponding logic for the work progress bars on the collection show page.

### DIFF
--- a/app/assets/stylesheets/sections/dashboard.scss
+++ b/app/assets/stylesheets/sections/dashboard.scss
@@ -181,6 +181,32 @@
   font-style: italic;
 }
 
+.owner-collection {
+  clear: both;
+  padding: $gapSize;
+  border-radius: 4px;
+  margin-bottom: $gapSize / 2;
+  background-clip: padding-box;
+  border: 1px solid rgba(#000, 0.06);
+  background-color: mix(#FFF, $bodyBg, 35%);
+  overflow: hidden;
+  &.stats-visible { border-color: transparent; }
+  &.stats-visible &_stats-bubble { opacity: 1; }
+  &_title {
+    margin: 0;
+    a {
+      text-decoration: none;
+      color: darken($fgLink, 10%);
+      &:hover { color: $fgLink; }
+    }
+  }
+}
+
+.visibility {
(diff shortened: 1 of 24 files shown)


saracarl (changes requested):
I like this.  Could you also test the UI for when an owner has no collections?  This is the case where an owner is brand new;  I'd like to confirm we still send them to the "Start a Project" once they get the account created, but also that if they end up here there is a message of some sort with a link to Start a Project.

I also restarted the failing tests, since they were failing due to an InternetArchive problem.

sylvieed:
Here's how it looks, this wasn't changed in this PR.
image
It doesn't have a link to the Start a Project page, it links to create a collection. Do you want to change this @saracarl?

saracarl:
Nope, that's fine @sylvieed.  I just wanted to make sure we didn't have regressions that would fail in the common and very important new owner scenario.  

This is great -- I really like how the Active/Inactive headers disappear if you only one of those 2 types.

saracarl:
There's still failing tests, 3 of them.  They are all failing with this because of the UI changes:

>   1) owner actions moves a work to another collection
>      Failure/Error: page.find('.collection-works .collection-work_title').find('a', text: @title).click
>      
>      Capybara::Ambiguous:
>        Ambiguous match, found 4 elements matching visible css ".collection-works .collection-work_title"
>      # ./spec/features/owner_actions_spec.rb:170:in `block (2 levels) in <top (required)>'

Issue #2368, opened 2021-02-12
work level subject index


saracarl:
We provide subject index exports for a collection; it would be useful to provide them for a single work.

requested by UT

benwbrum:
Once this is built, we should provide access to it on the `seeAlso` block of the manifest, and add this documentation back to the Wiki API docs:

### Subject CSV
The existing CSV export of subjects mentioned within the work.
```
      {
         "@id": "https://fromthepage.com/export/csv?work_id=11",
         "format": "application/csv",
         "profile": "https://github.com/benwbrum/fromthepage/wiki/IIIF-Export-CSV-Subjects",
         "label": "CSV export of all subjects tagged within the work"
      },
```

benwbrum:
This should be accessed from the Download tab of a work.  I believe that it should be fairly straightforward to use the document set subject details export here, since that passes in an array of works.

benwbrum:
Currently it is possible for a project owner to export all of the mentions of subjects within a collection; 
* For a project with subject indexing enabled (like Dartmouth's Wrangel Island or benwbrum's Jeremiah White Graves), click Export
* Click Export All
* Choose Subject Index CSV.
This will download a file containing one row for each time a subject is mentioned within the entire collection.

We would like a new export format that contains one row for each time a subject is mentioned within a single work.  This should be launched from the Download screen for the work, and should launch a bulk_export for the single work.  The export file should only contain rows from a single work.

PR #3676, opened 2023-06-12
#2368-Added Work Level Subject Index

symmetrically:


benwbrum (approved):
Looks great; works great!

PR #3678, opened 2023-06-20
2429 allow transcribers to delete subjects

symmetrically:


app/controllers/article_controller.rb
```diff
@@ -47,8 +47,13 @@ def list
 
   def delete
     if @article.link_list.empty? && @article.target_article_links.empty?
-      @article.destroy
-      redirect_to collection_subjects_path(@collection.owner, @collection)
+      if @article.created_by == current_user
```
benwbrum, line 50: Lets add an OR clause here so that the collection owner can also delete articles, even if they did not create the project.  (I think the relevant method is `like_owner?`)

app/models/xml_source_processor.rb
```diff
@@ -369,6 +369,7 @@ def update_links_and_xml(xml_string, preview_mode=false, text_type)
         article = Article.new
         article.title = title
         article.collection = collection
+        article.created_by = collection.owner
```
benwbrum, line 372: I'm curious why this is setting the article creator to the collection owner instead of to `current_user`?

config/locales/article/article-en.yml
```diff
@@ -13,6 +13,7 @@ en:
       selected_subjects_combined: Selected subjects combined with %{title}
     delete:
       must_remove_referring_links: You must remove all referring links before you delete this subject.
+      only_subject_owner_can_delete: Only the Subject Owner Can Delete
```
benwbrum, line 16: Let's change this to "Only the subject creator or project owner can delete this."

benwbrum (changes requested):
There are a couple of changes in functionality relating to project owners versus subject creators

benwbrum:
We are consistently getting this exception in the test suite:
```
  1) XmlSourceProcessor#wiki_to_xml subject linking not disabled (default) builds the xml document
     Failure/Error: article.save! unless preview_mode
     
     ActiveRecord::InvalidForeignKey:
       Mysql2::Error: Cannot add or update a child row: a foreign key constraint fails (`diary_test`.`articles`, CONSTRAINT `fk_rails_35e2f292e3` FOREIGN KEY (`created_by_id`) REFERENCES `users` (`id`))
     # ./app/models/xml_source_processor.rb:373:in `block in update_links_and_xml'
     # ./app/models/xml_source_processor.rb:353:in `update_links_and_xml'
     # ./app/models/xml_source_processor.rb:114:in `wiki_to_xml'
     # ./spec/models/xml_source_processor_spec.rb:41:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Mysql2::Error:
     #   Cannot add or update a child row: a foreign key constraint fails (`diary_test`.`articles`, CONSTRAINT `fk_rails_35e2f292e3` FOREIGN KEY (`created_by_id`) REFERENCES `users` (`id`))
     #   ./app/models/xml_source_processor.rb:373:in `block in update_links_and_xml'
```
The failure happens on [line 39 of the XML Source processor spec](https://github.com/benwbrum/fromthepage/blob/2429-allow-transcribers-to-delete-subjects/spec/models/xml_source_processor_spec.rb#L39), which is before the actual test runs, indicating that the problem may be with the fixture or factory data.  Could it be that we need to include a user reference in the [article factory constructor](https://github.com/benwbrum/fromthepage/blob/2429-allow-transcribers-to-delete-subjects/spec/models/xml_source_processor_spec.rb#L36)?

benwbrum:
The test failure here seems to be due to code in `xml_source_processor.rb` which sets the article's user relation to the value of `User.current_user`.  When the test fails, the value of `User.current_user.id` is `4`, but there is no record in the `users` table with an ID of 4 -- only 1 and 2 exist.

DB foreign keys were introduced in later versions of Rails than those current when the most of the  `xml_source_processor.rb` functionality was written -- as a result, models like `Page` and `PageVersion` can have their users set to values which may not be present in the database and tests don't fail.

I think we have a couple of options here:
* Remove the foreign key from the Article table.  (I'm not a fan of this, but it would parallel the other tests and older models.
* Research foreign key constraints and the tools we're using for creating test data in this test.  (Mocks?  Factories?)  Surely we are not the first developers to run into this problem, and there might just be a configuration option or something we can change.

PR #3679, opened 2023-06-20
WIP: Added search attempt tracking for #3644

### LINKED ISSUE #3644: Track and report on search and navigation success
How findable are projects?  How useful is our search feature for volunteers looking for something to work on?

We'd like to track these things:
* What are non-owner users (registered or not logged in) searching for?
* When they perform a search, do they get any results?
* If they get results, do they click on any of them?  (This indicates that the results were potentially useful)
* If they click on results, do they make any contributions?

For the data gathering portion of this, we can start by modifying the search form as follows:

* Create a new SearchAttempt model, with attributes for search text, hit count, click through count, and contributions.  
* Rather than a field that posts to an action, have the search function hit the `#create` action of the new `search_attempt_controller`.  This would redirect to the `show` action of that controller, which would contain the current object displaying the search results.  It would also update the hit count on the search attempt record.  (If necessary in later stages, we might also need to load this search attempt object into the user's session.)
* If a user clicks on a project, we should increment the click through count on the search attempt record.  While displaying the collection, we should also display a breadcrumb letting the user get back to their search results (closing #3608).
* If a user logs a deed on a project, we should find the relevant search attempt object (if any) and increment the contributions.

For the display portion, we should (at minimum) add a `Searches` tab to the admin dashboard.  This should list all search_attempts, with relevant fields.  We should also add a roll-up statistic or two to the dashboard showing average successful searches recently.  (@saracarl, what else are we interested in here?)

sylvieed:
Resolves #3644 

This PR stores and tracks search attempts made from the Find a Project page. 

### Updated search results page

The search results page is now `search_attempt#show` instead of the Find a Project page. I also added a "results for..." heading. The url for this search would be `http://fromthepage.com/search_attempt/letter-76`

search results

Here's how it looks with a longer search query:
long search query

And with no results (we might want to add some help text here):
no results

### Search results URL

The search results slug has the search query followed by its ID (ie: `alabama-77`). This means that users can't change the URL and find a random search. It also makes the results shareable. 

### Search results breadcrumb

The search results are placed in the breadcrumbs so users can return to the search results easily.

search results breadcrumb

### Admin tracking tools

There are a few rollup statistics that I included at the top of the new admin "Searches" tab.

search statistics

These are all for the last seven days. For the success rates on the second row, I included two versions of each: one per search, and one aggregated by ahoy visit. The latter was included because a user could make multiple searches in one visit and if any of them are successful, that visit is successful. We can add/remove to these as we see what's helpful. We could also add one or two of these statistics to the admin summary.

Below this, there's a table of all of the most recent search attempts. This can be filtered to only include searches by non-owners.
admin search table

Finally, admins can export all search attempts in a given time period.
searches export UI

This gives a CSV that looks like this:
csv search export


sylvieed:
_Resolves #3730_

This tracks searches on collections by text, works by text, and collections by title/metadata. It does so using the same `SearchAttempt` model used for Find a Project searches. The search results pages are unchanged and remain on the `collection#paged_search` and `collection#show` pages. However, the backend now goes through the `search_attempt#create` action to record the search, and the results pages use this SearchAttempt object to find the search's results.

I also improved  search result click tracking to be much more straightforward and reliable. Clicking on a search result calls a JavaScript method (via `onclick`) which calls a `search_attempt#click` action via AJAX.

### Breadcrumbs

Collection/work search results also have breadcrumbs to return to search results once a result is clicked.

breadcrumbs

### Admin tracking

I split the searches/day and average hits statistics into Find a Project searches and collection/work searches.

admin stats

The searches table now has fields for search type, collection, and work. You can filter by search type!

admin table

The csv searches export now has fields for search type, collection id, collection title, work id, and work title.

export

app/controllers/concerns/owner_exporter.rb
```diff
@@ -86,7 +86,39 @@ def owner_mailing_list_csv(owner)
     csv_string
   end
 
+  def admin_searches_csv(start_date, end_date)
```
benwbrum, line 89: Can we move this to a different concern file for admin-initiated exports?  Something like `AdminExporter`?

app/controllers/application_controller.rb
```diff
@@ -317,6 +317,8 @@ def configure_permitted_parameters
     # Otherwise owners should go to their dashboards
     # And everyone else should go to user dashboard/watchlist
   def after_sign_in_path_for(resource)
+    update_search_attempt_user(current_user)
```
benwbrum, line 320: I'd expect `after_sign_in_path_for` to be a read-only method, so adding this call to update the search attempt user creates an unexpected side-effect.  Is there a way to move the call to whatever calls `after_sign_in_path` or to possibly use a Devise-specific call-back to execute `update_search_attempt_user`?

app/controllers/collection_controller.rb
```diff
@@ -186,8 +186,15 @@ def show
...
 
-        if params[:work_search]
-          @works = @collection.search_works(params[:work_search]).includes(:work_statistic).paginate(page: params[:page], per_page: 10)
+        # Coming from work title/metadata search
+        if params[:search_attempt_id]
+          @search_attempt = SearchAttempt.find_by(id: params[:search_attempt_id])
+          if session[:search_attempt_id] != @search_attempt.id
+            session[:search_attempt_id] = @search_attempt.id
+          end
+          @query = @search_attempt.query
```
benwbrum, line 195: Would it make sense to have the @search_attempt model perform the search itself, combining this line and the @colllection.search_works call into one method?  That might encapsulate (some) search logic into the `SearchAttempt` model.

benwbrum (changes requested):
Please take a look at the comments -- there are some stylistic changes, some datatype things to check, and a bit of reorganization.  Generally, it looks really good!

db/migrate/20230519170746_create_search_attempts.rb
```diff
@@ -0,0 +1,15 @@
+class CreateSearchAttempts < ActiveRecord::Migration[6.0]
```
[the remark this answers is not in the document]
sylvieed: The new keys are `bigint`s, and the old ones are just `int`s

app/models/search_attempt.rb
```diff
@@ -0,0 +1,16 @@
+class SearchAttempt < ApplicationRecord
+    belongs_to :user, optional: true
+    belongs_to :collection, optional: true
+    belongs_to :work, optional: true
+    visitable class_name: "Visit" # ahoy integration
+
+    after_create :update_slug
```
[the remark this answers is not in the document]
sylvieed: I initially tried this, but you can't use the model's `id` for a slug using FriendlyId because the slug is set before the model is saved (before the id exists).
Using FriendlyId and only the query for the slug, duplicate searches get slugs like `letter-bbe44046-21d6-43a0-8e8d-c5fd3a80fa2e` 😞

app/controllers/concerns/owner_exporter.rb
sylvieed: Yep, done 👍

db/migrate/20230519170746_create_search_attempts.rb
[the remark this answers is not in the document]
benwbrum: I'm surprised that the references worked, then.  If you change the migration header to [5.0] instead of [6.0], for the table creation/column creation migration files, that should fix things.

app/controllers/application_controller.rb
sylvieed: Good idea, just moved it to a Devise/Warden [`after_set_user` callback](https://github.com/wardencommunity/warden/wiki/callbacks#after_set_user).

db/migrate/20230519170746_create_search_attempts.rb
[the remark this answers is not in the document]
sylvieed: Thanks, updated that.

app/controllers/collection_controller.rb
sylvieed: Yes, I feel like this makes it much cleaner! I moved all of the result finding logic to a `results` method in the `SearchAttempt` model, for all search types.

app/models/search_attempt.rb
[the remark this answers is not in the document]
benwbrum: Great!

PR #3690, opened 2023-07-06
2139-Version comparison diff view is made clearer for reverse chronological order comparison

symmetrically:


benwbrum:
There seem to be two problems with this fix; although they might have resulted from poorly thought-through problem definitions.

My Setup:
* Visit http://localhost:3000/benwbrum/citsci-2021-workshop/fieldnotesmexic00pete/display/25076988
* Add Title at the top of the transcript and save
* Add a row of equals signs at the bottom and save

Observed behavior
* Select a version and compare it with an earlier version: all looks fine (fig. 1)
* Now compare it with a later version: the text which was added in the later version looks like it was deleted (fig. 2)

We need to do another pass at this UI, following this rule:
* Green text should always be an addition from an older version to a newer version
* Red text should always be a deletion from an older version to a newer version

### Fig. 1
Screenshot from 2023-07-18 10-50-05

### Fig. 2
Screenshot from 2023-07-18 10-50-34

PR #3693, opened 2023-07-06
Fixes #3692 by respecting print configuration in facing edition view

### LINKED ISSUE #3692: Configurable PDF does not respect excluding metadata/contributions
If a user tries to run a configurable export from the work download screen, they still see metadata and contributors even if they un-check those two items.

When testing this, we should also test the bulk PDF export and bulk text-only PDF exports to verify no regressions were added.

benwbrum:
Closes #3692 

(See issue for testing notes)

### DIFF
--- a/app/controllers/concerns/export_service.rb
+++ b/app/controllers/concerns/export_service.rb
@@ -27,7 +27,7 @@ module ExportService
     out.write file.read
   end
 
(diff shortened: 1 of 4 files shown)


saracarl (changes requested):
Error trying to run this, on line 48 of export_service.rb
```
ArgumentError at /export/wwi-va-questionnaires/aaron-harry-bertram/printable
wrong number of arguments (given 4, expected 6)
```

Issue #3702, opened 2023-07-13
"Mark Blank" checkbox can be clicked accidentally, blanking transcripts


benwbrum:
Edits that created the transcripts:

### First edit in production.log.1
```
I, [2023-07-12T18:17:21.946461 #22582]  INFO -- : Started PATCH "/nehgsadmin007/divided-hearts-ma-loyalists-1765-1790/review/one_off/33116430" for 67.165.108.46 at 2023-07-12 18:17:21 +0000
I, [2023-07-12T18:17:21.948236 #22582]  INFO -- : Processing by TranscribeController#save_transcription as HTML
[...]

PR #3714, opened 2023-07-19
#3702-Added Check For Mark As Blank On Save

symmetrically:


benwbrum (changes requested):
Please internationalize the confirmation message.

benwbrum:
This works well for text transcription, but does not catch spreadsheet transcription.

I tested with http://localhost:3000/benwbrum/cosa-ledger-demo/tax-lists-pasquotank-county-1769 and did not get the warning.  We should also test with field-based transcription without spreadsheets.

Issue #3566, opened 2023-03-22
Show contributors in work metadata download spreadsheet.


anabeabas:
Jason Godfrey of the Wilford Woodruff Papers Foundation asked for a feature that would let him see which specific documents collaborators are transcribing. He said being able to differentiate the documents that staff and public volunteers transcribe would be ideal. He said he wants to monitor the progress of collaborators so he can train them on specific components of documentary editing related to their project and recruit them to do more transcribing.

He said he wants to see this information for all of the collaborators, but especially new ones.

**NOTE: We should include contributors in the collection metadata spreadsheet**

benwbrum:
This has been fixed.

PR #3741, opened 2023-08-11
#3566 Add column contributors in metadata csv

symmetrically:


app/controllers/concerns/export_service.rb
```diff
@@ -407,7 +407,8 @@ def export_work_metadata_as_csv(collection)
         'Pages Translated',
         'Pages Needing Review',
         'Pages Marked Blank',
-        'work_id'
+        'work_id',
+        'Contributors Users'
```
benwbrum, line 411: Please relabel this to "Contributors" and move it before `work_id` in the column list.

benwbrum (changes requested):

benwbrum:
This changes the order of the column headers but not the column data itself, leading to results like this:
Screenshot from 2023-08-16 14-33-33

Issue #3738, opened 2023-08-10
Export notes for a collection


saracarl:
It would be useful, for mass-review, to have a notes export for a collection.  It would need:

The note
The contributor
The page (link)
The work (name)
The page (name)

This would be a CSV.  Backgrounded, in the export tab (or the export all tab?)

Project owners would use it for 2 things:
preservation, outside of the document itself.
and quick review for words like "duplicate" "dupe" "copy" "missing" etc.

Requested by LVA

PR #3742, opened 2023-08-14
Added collection notes bulk export for #3738

sylvieed:
_Resolves #3738_ 

This adds a bulk export for all notes in a collection, to allow project owners to preserve & review notes.

The export is located on the collection export page:

export page

And here's how the exported CSV file looks, with the Utah Birth Certificates collection as example:

export csv


benwbrum:
Sylvie, do you think you could move this to the bulk export screen?  We're trying to move one-off exports off of this page, since these launch exports which run within the request lifecycle, causing performance bottlenecks down the road.

PR #3750, opened 2023-08-17
Improved collection notes CSV export for #3738

sylvieed:
_For #3738, updates #3742_

The notes export is now in the Export All Works page:
export UI

And the export's columns have been updated to the same order as the table/field CSV export: 
Work Title, Work Identifier, FromThePage Identifier, Page Title, Page Position, Page URL, Page Contributors, Page Status, Note, Contributor, Date

export CSV


db/migrate/20230814171021_add_collection_notes_to_bulk_exports.rb
```diff
@@ -1,5 +1,5 @@
 class AddCollectionNotesToBulkExports < ActiveRecord::Migration[6.0]
   def change
-    add_column :bulk_exports, :collection_notes, :boolean
```
benwbrum, line 3: Because we've already applied the earlier version of this code, we need a new migration that renames `collection_notes` to `notes_csv`.  When I try to do a bulk export on this branch, I get "undefined method `notes_csv' for #<BulkExport:0x00007f09953aeaa8>"

benwbrum (changes requested):
Let's create a new migration here instead of re-using the old one.

db/migrate/20230814171021_add_collection_notes_to_bulk_exports.rb
sylvieed: Fixed 👍

benwbrum (approved):
This looks great!

Issue #3680, opened 2023-06-20
Display page status on page version screen


benwbrum:
Currently there is no way to see when a page moves from `not started` to `incomplete` to `approved`, or other status changes.  We should store the status-at-the-time-of-creation on the `page_versions` table and display it in the page version screen.

(See also #2772 and #2139; related #2771 and #2282; related #2744)

benwbrum:
Ideally, the page status would be displayed alongside each version date/user so that anyone looking at the version screen could skim over the page's history easily.

PR #3753, opened 2023-08-22
#3680-Added Status on Version page

symmetrically:


app/views/page_version/show.html.slim
```diff
@@ -22,6 +22,7 @@ table.diff-versions(data-fullheight='{ "bottom": 30, "cssrule": "min-height" }')
           =link_to(page_version_show_path(:page_version_id => version.id), class: ('selected' if version == @selected_version))
             =time_tag(version.created_on)
               =l(version.created_on.to_date)
+              = version.status && render_status_icon(version.status)
```
benwbrum, line 25: Let's pull the status from the messages using something like `t("page.edit.page_#{version.status}")` here so that we can see a more attractive (and translated) status on hover.

benwbrum (changes requested):
We need to internationalize the status string pop-up.

benwbrum (changes requested):
This internationalizes the status, but does not use the human-friendly status values that already exist here: https://github.com/benwbrum/fromthepage/blob/development/config/locales/page/page-en.yml#L18-L25

Also, let's make the icon for a `nil` status blank.  The red question mark is likely to disturb project owners on older systems.

Issue #3717, opened 2023-07-20
show works list for a collection to transcribers


saracarl:
We have a "works" tab for collections if you're an owner.  It's useful.  We should show it to transcribers.

PR #3755, opened 2023-08-23
Implemented DataTables for works list for #3736

sylvieed:
_Resolves #3736_

This uses [DataTables](https://datatables.net/) to show the "Works List" for collections and document sets.
This enables:
* Showing a variable amount of works on the page (10, 25, 50, 100)
* Filtering works by name (or other columns)
* Sorting ascending or descending, and more intuitively
* Everything is handled on the client side!

image


saracarl (changes requested):
This is amazing and folks are really going to love it.  2 changes:
1)  Let's offer an "all" in the number of works.  And once we add that, I think having 5 options is too many.  Maybe 10/50/all?  (open to discussion there).
2)  This is so useful I think we should show it to transcribers as well.  Can we add a "works" tab to the transcriber view of the collection?

saracarl:
Related:  #3717

sylvieed:
The number of works options is now 10, 50, and All:
image

And the Works List tab is visible to everyone:
image

saracarl:
I'm a bit worried that the status bars seem different in this branch vs production.  My data isn't a perfect match/isn't 100% up to date but look at the differences here:

Production:
image

- yellow for needs review on 1923
- light green for transcribed
- dark green for indexed

This branch:
image
- 1921 doesn't show the light green for the transcribed but not indexed pages
- no yellow for needs review on 1923

Any idea why this changed?

saracarl (changes requested):
see comment

sylvieed:
From my testing, the progress bars are accurate now -- in your case where some of the sections were gone, and in mine where 100% completed works weren't showing as 100%.

PR #3756, opened 2023-08-23
Remove search attempt from session when searcher leaves results for #3644

sylvieed:
This adds a `:before_action` function that removes the search attempt from the session depending on the action's controller. If they access their dashboard, landing page, find a project page, or profile page, their search activity is done and so it's removed from the session. It's also removed if they log out. 
This is important for reliable search attempt data.


benwbrum (approved):
This looks good and works well.

Issue #3758, opened 2023-08-24
forums broken with 500 error due to new landing_path


saracarl:
```

I, [2023-08-24T13:36:00.609342 #3485]  INFO -- :   Rendered thredded/messageboard_groups/show.html.erb within layou
ts/application (Duration: 114.8ms | Allocations: 36359)

F, [2023-08-24T13:36:01.441046 #3485] FATAL -- :   
ActionView::Template::Error (undefined method `landing_path' for #<#<Class:0x00007f2def4e3d28>:0x00007f2def4e1aa0>)
:
    35:     header.header
    36:       .toolbar
    37:         .toolbar_group
    38:           =link_to landing_path(logo: true), class: 'logo', title: t('.home'), 'aria-label' => t('.home')
    39:             -if USE_PNG_LOGO
    40:               =image_tag 'logo.png', width: 134, height: 50, alt: t('.home')
    41:             -else
  
app/views/layouts/application.html.slim:38
```

To reproduce you'll need to enable forums for a collection and then navigate to the forums tab on the collection

benwbrum:
This can probably be resolved by prefixing `landing_path` with `main_app` so that our application renders within the context of the Thredded engine.

PR #3760, opened 2023-08-24
#3758 fixed broken forum on collection show page

symmetrically:


benwbrum (changes requested):
Please add a feature test that enables a forum for a collection, then visits the forum.

Issue #3739, opened 2023-08-10
Filter/search comments on a collection


saracarl:
We have a "notes" section for a collection, i.e. https://fromthepage.com/deed/notes/jeremiah-white-graves-diaries

LVA would like a way to search or filter all the notes for the collection (rather than page through them) in order to find things like "mismatch" "duplicate" "missing", etc.

Maybe datatables?

benwbrum:
How is this different from #3738?  Just on-screen?

benwbrum:
This was fixed with the #4387

PR #3761, opened 2023-08-24
Implemented DataTables to deed list for #3739

sylvieed:
_Resolves #3739_

This is meant to solve filtering/searching collection notes by implementing DataTables which comes with a search. Because the notes list is made by filtering the deed list to only notes, I just edited the deed list to use DataTables.

Here's the notes list, it can quickly be filtered and sorted:
image

And this also changes the all deeds list:
image

If we don't want to change the deeds list, I can separate it from the notes list.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -337,7 +337,7 @@ Object.assign(DataTable.defaults, {
 
   "drawCallback": function(oSettings) {
     // don't show pagination if only one page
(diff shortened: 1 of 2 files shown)


saracarl:
I *love* this.  I think our super transcribers and project owners will, too.  You can filter by just your name to see what you've done.  You can search deeds, which gives you a list by person to look through.  You can filter or sort by "needs review" or "edited" etc.  I can't wait to see how folks use this.  Once we get it deployed I'm going to let a handful of both project owners and super transcribers know and ask for their feedback.

PR #3783, opened 2023-09-12
added precautionary check for checking class

symmetrically:


benwbrum:
This fixes the 500 error, but is not quite the correct behavior.

If the `@collection` object is actually a document set, we should return the parent collection's help text.  Several methods in the `DocumentSet` class return an attribute of the collection, so there should be good examples.

Issue #3790, opened 2023-09-18
deed#notes takes two hours to render


benwbrum:
This happened at 12:40 PM on September 17.
https://onenr.io/0LRExLJ3nwa

It also happened at 5:26 on September 16.
https://onenr.io/0BQ1vKb6Vwx

Both appear to be launched from the profile partial

benwbrum:
The production log seems to show this as a full retrieval of all notes for the WWI questionnaires at LVA, taking a couple of hours to render everything.
```
I, [2023-09-17T13:09:23.311481 #2109]  INFO -- : Started GET "/deed/notes/wwi-va-questionnaires" for 157.90.182.28 at 2023-09-17 13:09:23 +0000
I, [2023-09-17T13:09:23.313317 #2109]  INFO -- : Processing by DeedController#notes as HTML
I, [2023-09-17T13:09:23.313654 #2109]  INFO -- :   Parameters: {"collection_id"=>"wwi-va-questionnaires"}
I, [2023-09-17T13:09:23.314542 #2109]  INFO -- : [ahoy] Visit excluded
I, [2023-09-17T13:09:23.319860 #2109]  INFO -- :   Rendering deed/list.html.slim within layouts/application
I, [2023-09-17T13:09:24.855163 #2109]  INFO -- :   Rendered shared/_profile_picture.html.slim (Duration: 1.9ms | Allocations: 243)
I, [2023-09-17T13:09:25.558540 #2109]  INFO -- :   Rendered deed/_deed.html.slim (Duration: 702.8ms | Allocations: 422471)

[Hundreds of lines skipped]

I, [2023-09-17T17:39:00.894826 #2109]  INFO -- :   Rendered deed/_deed.html.slim (Duration: 1076.3ms | Allocations: 152539)
I, [2023-09-17T17:39:00.897947 #2109]  INFO -- :   Rendered shared/_profile_picture.html.slim (Duration: 0.9ms | Allocations: 215)
I, [2023-09-17T17:39:01.389105 #2109]  INFO -- :   Rendered deed/_deed.html.slim (Duration: 490.4ms | Allocations: 150738)
I, [2023-09-17T17:39:01.391453 #2109]  INFO -- :   Rendered shared/_profile_picture.html.slim (Duration: 1.1ms | Allocations: 249)
I, [2023-09-17T17:39:01.928249 #2109]  INFO -- :   Rendered deed/_deed.html.slim (Duration: 536.2ms | Allocations: 150738)
[...]

PR #3792, opened 2023-09-19
Implemented server-side datatables for notes list for #3790

sylvieed:
_Resolves #3790_

After implementing DataTables on the notes & deeds list, these pages have performance issues on long lists because all of the rows must be loaded by the server at once.
To fix this, this PR changes the notes DataTable to use [server-side processing](https://datatables.net/manual/server-side). This means that the server only loads the currently displayed rows at once, eliminating the issue for large tables. This also means that all sorting and searching is handled on the server-side instead of the client-side, resulting in a slower UI (but it's consistent over all table lengths!)

I used the [ajax-datatables-rails](https://github.com/jbox-web/ajax-datatables-rails) gem, which handles a majority of the server-side processing. 

I split the deeds list and notes list, so the deeds list is always paginated and the notes list is always a DataTable. Now, the notes list has individual sortable columns rather than one large "deed" column. 

The notes list for a collection:
image

The all notes list:
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -125,3 +125,5 @@ gem 'intercom', '~> 3.9.0'
 gem 'newrelic_rpm'
 
 gem 'open3'
+
+gem 'ajax-datatables-rails', '~> 1.0.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,6 +100,8 @@ GEM
       safely_block (>= 0.2.1)
     airbrussh (1.4.1)
       sshkit (>= 1.6.1, != 1.7.0)
+    ajax-datatables-rails (1.0.0)
+      railties (>= 4.2)
     akami (1.3.1)
       gyoku (>= 0.4.0)
       nokogiri
@@ -597,6 +599,7 @@ DEPENDENCIES
   acts_as_list
   acts_as_tree
   ahoy_matey
+  ajax-datatables-rails (~> 1.0.0)
   autoprefixer-rails
   better_errors
   binding_of_caller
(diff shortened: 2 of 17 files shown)


saracarl:
Yay!  This is great, and I think the fields for the Notes view is going to be useful.  

Does this: _so the deeds list is always paginated_ mean you can't see all deeds at once?  (And if so is that what you and Ben decided on?)

saracarl:
Test failure is this:
```
  1) Deed.order_by_recent_activity lists deeds in order by most recently created
     Failure/Error: expect(Deed.order_by_recent_activity.first).to eq(second_deed)
     
       expected: #<Deed id: 110, deed_type: "page_trans", page_id: nil, work_id: nil, collection_id: nil, article_id: ...ed_at: "2023-09-19 20:21:28", visit_id: nil, prerender: nil, prerender_mailer: nil, is_public: true>
            got: #<Deed id: 109, deed_type: "page_trans", page_id: nil, work_id: nil, collection_id: nil, article_id: ...ed_at: "2023-09-19 20:21:28", visit_id: nil, prerender: nil, prerender_mailer: nil, is_public: true>
     
       (compared using ==)
     
       Diff:
       @@ -1,5 +1,5 @@
       -#<Deed:0x0000000013204190
       - id: 110,
       +#<Deed:0x0000000013094c88
       + id: 109,
         deed_type: "page_trans",
         page_id: nil,
         work_id: nil,
       
     # ./spec/models/deed_spec.rb:27:in `block (3 levels) in <top (required)>'
```

sylvieed:
Yes, you can't see all deeds at once. This just returns the deeds list to how it originally was -- it was never the plan to edit this list, I just did because it was the same view as the notes list.

deed list

Issue #1983, opened 2020-08-17
Navigate from Document Set to Collection


benwbrum:
If a user is working on a document within a document set, and if they have access to the document set's parent collection, they still have no way to navigate to that collection without starting at FindAProject page.

Implementation:
Analyze possible locations for a link to the parent collection.  (I'd favor the line above the current Document Set - Work - Page header which serves as a breadcrumb, but in a smaller font.)  Add the parent collection link, provided that the current user can see the collection.

Testing:
Test on projects that have a private collections and public document sets (like LVA's WWI questionnaires), as well as projects that have public collections and sets, and projects that have no sets.

ThomBurns:
This would be wonderful! Love this idea.

PR #3802, opened 2023-09-27
#1983 Added functionality to redirect from document set to its parent collection

symmetrically:


benwbrum (changes requested):
1) We need to handle the situation where a collection is private and a document set is public.  Currently this code displays a link to the collection in the header, even though the user cannot see the collection.
2) We should also display the collection breadcrumb in the `collection#show` action header.

benwbrum:
https://github.com/benwbrum/fromthepage/assets/199961/a17cdee3-25bc-401f-837e-2ebc45df80b7

benwbrum:
I just tried this out and found a scenario in which the collection breadcrumb does not appear when it should.

* Create a collection which I own
* Make the collection restricted
* Create a public document set
* Place a work in the document set
* View the work.

Even though I own the collection and can see it, the collection is hidden.

I think that the logic added in 43e18db2b8101f69af49cd1c87e6c976a560cf9f should be changed to call `@collection.show_to?(current_user)` , which will handle the restricted/blocked logic for us.

Issue #3798, opened 2023-09-22
add contributor's real name to work metadata export


saracarl:
When you export the work metadata, we'd like to add a field of contributors that includes their real_name if we have one, or their displayname if we don't.  

Currently we just do user.display_name:
https://github.com/benwbrum/fromthepage/blob/c614fdea263852fd44bac539c63d055f5e7afc52/app/controllers/concerns/export_service.rb#L437

UT Austin asked for this.

benwbrum:
This should be in a separate column, following the current contributor column.

PR #3822, opened 2023-10-17
#3798 Added new column for contributors real names

symmetrically:


app/controllers/concerns/export_service.rb
```diff
@@ -435,6 +436,7 @@ def export_work_metadata_as_csv(collection)
       collection.works.includes(:document_sets, :work_statistic, :sc_manifest).reorder(:id).each do |work|
     
         work_users = work.deeds.map{ |d| "#{d.user.display_name}<#{d.user.email}>".gsub('|', '//') }.uniq.join('|')
+        contributors_real_names = work.deeds.map{ |d| "#{d.user.real_name}" }.uniq
```
benwbrum, line 439: This looks like it'll produce an array, rather than the string we need in the CSV output.  Please join the contributor names using a pipe delimiter as in the previous line.

benwbrum (changes requested):

PR #3826, opened 2023-10-20
Fixed Codespaces environment

sylvieed:
This fixes a couple of errors that were preventing the Codespaces environment from working correctly.

Initial setup was in #3627.


sylvieed:
The problems fixed are:
* Codespaces host URL was blocked by Rails because the domain name changed slightly
* Bundler was not installing because the Ruby version, 2.7, is unsupported by the newest Bundler
* Newest changes were not automatically fetched
* There was an authentication token error when logging in or signing up (still not sure what exactly caused this -- I fixed it just by removing the check for this token when in the codespaces environment)

Issue #2858, opened 2021-10-28
volunteer hours downloadable letter


saracarl:
Similar to #1240, we should provide a downloadable letter that documents user contributions.  

We could start with a simple letter that says who the owning organization is (what if they work on multiple projects?), but we could also let project owners upload a logo and a signature and format a PDF for their specific project.

anabeabas:
A transcriber needed it today.

saracarl:
Here's a sample letter we built by hand today for someone needing documentation for court appointed service hours.  
[MichaelLaleskaServiceHours.pdf](https://github.com/benwbrum/fromthepage/files/13033255/MichaelLaleskaServiceHours.pdf)

(I have a doc file as well, if we want to use that while building this.)

Key things we decided:
-- document total hours for a time frame; that will include clicking around time and commenting/editing time.  
-- for the table, we counted "pages they had worked on" -- transcriptions, notes, edits, etc.  
-- We dropped the one collection that had a page count of 0 (not sure why it showed up, but it did)

benwbrum:
Code we used to do this as a one-off in the console today:
### Get the user's time

```ruby
user_id = 32017019
user = User.find user_id

summaries_for_user= AhoyActivitySummary.where(user_id: user_id)
```

### Record time per collection
*Note that some time does not have a `collection_id`, so this could be undercount "overhead" time navigating to projects, etc.
```ruby
collection_minutes_map= summaries_for_user.group(:collection_id).sum(:minutes)
collection_minutes_map.each_pair do |collection_id, minutes|
  collection=Collection.where(id:collection_id).first
  if collection.nil?
    output_array << [nil, nil, minutes]
  else
    output_array << [collection.owner.display_name, collection.title, minutes]
  end
end
f = File.open("/tmp/time_per_collection.csv", "w+")
f.print output_array.map{|row| row.join(",")}.join("\n")
f.close
```
### Record time per day
```ruby
date_minutes_map= summaries_for_user.group(:date).sum(:minutes)
output_array= []
date_minutes_map.keys.sort.each do |date|
  output_array << [date, date_minutes_map[date]]
end
f = File.open("/tmp/time_per_day.csv", "w+")
f.print output_array.map{|row| row.join(",")}.join("\n")
f.close
```

### Record pages per collection
```ruby
pages_per_collection = user.deeds.where(deed_type: DeedType.collection_edits).group(:collection_id).count(:page_id)
output_array=[]
[...]

PR #3837, opened 2023-10-30
#2858 Added feature for showing contributions hours view and PDF letter download

symmetrically:


benwbrum (commented):
The code looks good to me!

benwbrum:
I'm not entirely sure what to do about the line breaks with `collec-` -- there are a [handful of options](https://pandoc.org/MANUAL.html) you can pass to Pandoc, so could we explore a bit?

Let's right-justify the `Pages Count` column and rename the header to `Pages`.  That may give us a bit more room for the collection.  (Note that we had to use a special markdown tables plugin when we run Pandoc in the work PDF export, which is passed in as a command-line argument.)

saracarl (changes requested):
Ok.  2 more small changes:
1)  Add a carriage return between the logo and the date.
2)  Make sure it's FromThePage, not FromthePage.  (Branding!)  It's wrong 2 places and correct 1 place.

Otherwise, looks good.

saracarl (changes requested):
Messages are missing:
[letter (2).pdf](https://github.com/benwbrum/fromthepage/files/13427508/letter.2.pdf)

I would probably have added the messages to the dashboard messages file, to match the controller you're using.  With your current setup, you can probably make this work by fully qualifying the message references, i.e. `hours_letter.pdf.letter.certification_text`

We use i18n-tasks to do a lot of this work; it makes it *much* easier.  I'll send an email with some details about that library and our google translate key.  (I probably wouldn't use it with this PR, because it requires starting over, but I would with all the ones in the future.)

Issue #2282, opened 2021-01-20
Show source in page version history


benwbrum:
Currently the version history tab for each page shows a selected version of the rendered page, and a diff version of the rendered page.  If we followed each of these with a source version, it would make changes to mark-up visible.

PeterMarteau:
Here's my ideas to that:

**Current situation**
Right now, you get an introduction, transcribe your pages, chat while doing asking questions and doing the work. then someone reviews it and corrects your transcription. 

**Why you probably don't really learn from your errors**
now you could look up the corrections to learn from them. the problem is, that to fully comprehend what you did wrong and why you did it wrong, it's not enough to see the corrected transcription. you need to see the actual ms to reframe the scribes hand with the correct translation in your head. you could go back to the ms of course, but to be honest, no one probably does that. 

**How to solve that**
So my idea was to add another tab called 'review', copy the content of the version history tab and then modify that. Mainly by replacing the left src frame with the facsimile. So that you end up with sth like this:
![image](https://user-images.githubusercontent.com/13455634/105251022-09e46480-5b7b-11eb-9e2b-8433ad47b213.png)
This way you can leave everything as it is and for the new pages you can reuse stuff you already have. 

**How that could pave the way for a different, "agile" workflow in transcribing**
also when you can learn from the review, what one could then propose to teachers as a pedagogical concept is something like "agile transcription": instead of doing 10 pages of transcription in a row and then doing a review phase, one could translate one page, doing a review, learn from the review, doing the next page. i think that would result in a more productive transcription and better learning cause you incorporate your new knowlege and use it directly when transcribing the next pages.

Issue #2744, opened 2021-08-27
Revert in versions tab


saracarl:
Molly K asked if it was possible to revert to a particular version.  With text you can just copy and paste (although that doesn't work with advanced mark-up or subject linking, see #2282 ), but with spreadsheets and densely marked up material you need an actual revert button.

benwbrum:
After discussions with NEHGS, it seems like this is the best UI.  (Note that NEHGS is using spreadsheet-based transcription, so we will not be able to revert their pages until #1615 is completed and we have saved spreadsheets on page versions.

##### Me
We've never had a "revert" functionality before, except when dealing with spam or other abuse.  Can you tell me a bit more about how you see this working?  We'd probably want to preserve the work that the lower-quality user did, as it was honest effort and we don't want to dissuade anyone. 

My initial thought is to add a two-step process, in which you as project owner would hit "restore" next to an old version, then would be dropped directly into the transcription screen to make any additional edits you see as necessary.

##### NEHGS
It would be nice to have the option, if not to "restore" a previous version (though I really like the idea of a two-step process, as long as it also displays the page facsimile to work off of!), to at least be able to download a previous version as a csv, without all the edits/changes showing. I think having an "export version" button appear would be very helpful in situations like this.

PR #3838, opened 2023-10-31
#2282 added review tab for page versions

symmetrically:


saracarl:
This PR implements more or less what an academic transcriber sketched out in #2282.

There are some implementation problems, namely no line breaks and no mark-up, which means this isn't very useful:

image

But there's a much bigger problem, which is "what's the purpose of this tab?"

Was the intention of this to be able to load the page source at a given revision and compare it against the facsimile?  *To what point*?  I feel if you're are going to review, you must be able to edit.  But do we really want any transcriber to be able to set a page back to an earlier version and save it?  I'm not sure we do.  (This becomes effectively "revert", #2744)

@benwbrum to review.

benwbrum:
Replaced by #4289

PR #3850, opened 2023-11-14
#2095 Include page counts on subject browse

### LINKED ISSUE #2095: Include page counts on subject browse
> been talking about product features we would be interested in. We are working with our subjects and it would be really helpful to be able to see the number of pages that each term appears on when browsing the list subjects, perhaps as a small gray number next to the term. This would help the causal researcher figure out who among our 800+ people might be interesting and/or important.

symmetrically:
Closes #2095

### DIFF
--- a/app/assets/stylesheets/sections/collection.scss
+++ b/app/assets/stylesheets/sections/collection.scss
@@ -298,6 +298,31 @@
   }
 }
 
+.category-article dd {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.category-article dd p {
+  width: 33%;
+  display: flex;    
+}
+
+.category-article dd p a {
+  width: 100%;
+}
+
+.category-article dd p span {
+  width: 40%;
+}
+
+.grey-number {
+  color: grey;
+  margin-right: 5px;
+  font-size: medium;
+  padding-left: 10px;
+}
 
 // Collection stats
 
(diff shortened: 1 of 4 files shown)


benwbrum:
The UI for this looks very good, but the performance is too slow for large projects.

If I log in as owner `petelin` then visit http://localhost:3000/cwrgm/cwrgm-rev2/subjects the page takes five minutes to load; most of which is spent executing SQL queries to find the page count for each subject.

`Completed 200 OK in 318310ms (Views: 267481.2ms | ActiveRecord: 48510.1ms | Allocations: 104765563)`

benwbrum:
Functionality and performance are great.  Could we modify the UI so that the number of pages is displayed next to the subject, in parentheses, but still in the grey color?

benwbrum:
Let's abandon the column spacing and add the parenthetical page counts directly after the subject titles.

symmetrically:
@benwbrum If we abandon the column spacing, Then, it will look weird in context of UI because the Subject titles can be varied in length. Look at below screenshot.

As we are using that span in `Link_to` of subject title
image (98)

saracarl:
Dropping in 2 screenshots of the earlier layout for comparison.
Screenshot from 2023-11-16 10-33-51
Screenshot from 2023-11-16 10-42-20

saracarl:
@symmetrically could we do the link_to *inside* the span, with the page count immediately afterwards with only a space separating it?  
Like this, more or less:

[Newspapers](url) (80)

saracarl (changes requested):
See earlier UI comments.

Issue #3247, opened 2022-08-04
extreme top and bottom of image unreadable after rotating in over/under view


saracarl:
> My “frustrating” issue is as follows:
> 
> When transcribing, I prefer to use “Layout: Image at the top” because I can zoom in on the image (facsimile) and still have an entire line in the viewport.
> 
> The problem occurs when a document was scanned in portrait orientation, but written in landscape orientation.
> 
> This requires me to rotate the image 90 degrees to read it.
> 
> Then, when I drag the image to see the extreme top (or bottom) of the image, the image “rebounds” back to a more centered position when I let go of the mouse, and now I cannot see the extreme top or bottom lines (which are the left or right sides in portrait orientation).
> 
> This does not happen in “Layout: Image at the left” or “right”, but those layouts are too narrow to see the whole line when zoomed in enough.
>
> I’m not sure why the viewport acts this way, but it’s preventing me from performing efficient transcription.
> 
>  
> 
> You can test out what I mean on this page: https://fromthepage.com/harvardlibrary/colonial-north-america-countway-library-of-medicine/waterhouse-benjamin-1754-1846-diary-of-diseases-then-at-st-thomas-s-hospital-london-1775-b-ms-b10-3-countway-library-of-medicine/transcribe/32087532

PR #3919, opened 2024-01-16
"Rebound" effect in OpenSeadragon viewer

rsimon:
This small PR addresses issue #3247: the bounce-back effect that users experience when dragging the image outside of the viewer area.

The effect is controlled by OpenSeadragons's `visibilityRatio` init argument. In your current setting (visibilityRatio = 1), OpenSeadragon will attempt to bounce the image back to a state where __the viewer shows no empty background.__ Consequently, if users have a narrow viewport set, then attempts to drag the image to an "extreme edge" position  (as reported in issue #3247) can easily lead to this effect.

- user drags the image slightly out of the image
- the inertial bounce-back effect pushes the image back (possibly overshooting the original starting position)

The easiest way to address this is to set `visibilityRatio` to a more relaxed setting, like 0.5. This instructs OSD to allow 50% of empty background before bouncing the image back, thus preventing the effect in most cases. (0.5 is actually the default. Therefore you could also remove the init argument in this particular case!)

An alternative approach (not implemented in this PR!) would be to combine:

```
visibilityRatio: 1,
constrainDuringPan: true
```

This will prohibit empty background in OSD altogether, and apply the constraint immediately during panning. I.e. there is no way to ever push the image out of the viewer, and there is no inertial rebound effect that bounces the image back. Instead, the image just stops as soon as the user reaches the edge.

An example of this combination is shown at the [official OSD docs here](https://openseadragon.github.io/examples/ui-zoom-and-pan/).

### DIFF
--- a/app/views/shared/_osd_div.html.slim
+++ b/app/views/shared/_osd_div.html.slim
@@ -74,7 +74,7 @@
         previousButton: "osd_prev",
         nextButton: "osd_next",
         maxZoomPixelRatio: 10,
-        visibilityRatio: 1,
+        visibilityRatio: 0.5,
         fitBounds: true,
         preserveViewport: true,
         showReferenceStrip: #{@page.nil?},


rsimon:
P.S.: I noticed you are using a `fitBounds` init argument. As far as I'm aware, this argument doesn't have any effect. (I checked specifically for OSD v3.0.) The initial image position always defaults to a "fitBounds" setting. This arg should be safe to remove.

saracarl:
This solution works nicely, and I can't think of any scenarios where it would cause problems.

p.s.  @rsimon feel free to remove `fitBounds` the next time you're in the area.  (This PR would have been fine.)

PR #3925, opened 2024-01-19
#3812 added filter for works require only transcription

### LINKED ISSUE #3812: work tab improvements
Feedback from Christina at MO State. Archives on the sortable and filterable work tab:

> -Is there anyway to split the progress section into two different lists so we can sort by transcription progress and by review progress separately.

I like this a lot.

> -a more defined filter might be useful.  The search bar works to filter works title, completed works, and needs review works well, but there is not really a way to pull up just the works that need transcription but don’t have anything to review.

I don't know if there's a way to do this -- I'd like to try the first change, and then see if we can accomplish it by sorting by reviewed (most to least) and then by transcription.

symmetrically:
Closes #3812


saracarl (changes requested):
see above comment.

PR #3930, opened 2024-01-23
#3921 don't log a version or time if a page hasn't changed

### LINKED ISSUE #3921: Volunteer Hours Abuse: Don't Log a Version or Time if a Page Hasn't Changed
Sonya of LVA emailed us about a transcriber who wants his volunteer hours certified but wasn't doing any actual work. Here’s a particularly egregious example:

https://fromthepage.com/lva/virginia-untold-deeds-of-emancipation/7800740-0007-0006/versions/32227816

The solution we came up with: If a page hasn't changed, we shouldn't log a version or time on it.

symmetrically:
Closes #3921

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -244,31 +244,39 @@ class Page < ApplicationRecord
   end
 
   def create_version
-    version = PageVersion.new
-    version.page = self
-    version.title = self.title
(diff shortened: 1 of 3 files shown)


saracarl:
Doesn't log a version and doesn't log time.  :+1:

Issue #3955, opened 2024-02-07
public document sets link to the private main collection


saracarl:
A document set has a link to the parent collection ("Ben's Test Collection").
  https://fromthepage.com/benwbrum/public-test-set

![image](https://github.com/benwbrum/fromthepage/assets/708566/62c80198-2e21-485d-aea3-b78ec7d7ef60)

But the collection is private and I can't see it:
![image](https://github.com/benwbrum/fromthepage/assets/708566/015065ed-c8c0-4b0c-915e-8ffd4c5877b5)

reported by Dreanna

PR #3964, opened 2024-02-20
#3955 Added checks for not showing private collection link on public document set

symmetrically:


app/views/shared/_breadcrumbs.html.slim
```diff
@@ -10,7 +10,7 @@
   -if @collection.is_a?(DocumentSet) && @collection.show_to?(current_user)
     -path.push(link_to @collection.collection.title, collection_path(@collection.collection.owner, @collection.collection))  
   -path.push(link_to @collection.title, collection_path(@collection.owner, @collection))
--elsif @collection && @collection.title.present? && @collection.is_a?(DocumentSet)
+-elsif @collection && @collection.title.present? && @collection.is_a?(DocumentSet) && @collection.is_public? && !@collection.collection.restricted?
```
benwbrum, line 13: Can we try using `@collection.is_public? || @collection.show_to?(current_user)` here?  There is a scenario in which a user is looking at a document set in a private collection, but they have been added as a collaborator on that private collection so they should be able to see it.

benwbrum (changes requested):

app/views/shared/_breadcrumbs.html.slim
symmetrically: @benwbrum The check that you have provided in feedback will not work for the required scenario. I have updated the check and added a check for checking the collaborator of collection while showing the link in breadcrumbs. I have pushed the changes.

PR #4101, opened 2024-05-20
Fixed bad field alignment on long labels

### LINKED ISSUE #4097: Bad field alignment on long labels
For field based projects, long-labels may cause fields to move out of algnment with other fields on the same line.

For example, see this page (with side-by-side image view): https://fromthepage.com/missouristatearchives/missouri-state-penitentiary-register-of-inmates/msp-vol-80/transcribe/34038831

Screenshot from 2024-05-17 13-40-11

A related problem to explore is the movement of the last field to a new line [in this project](https://fromthepage.com/stanfordlibraries/monterey-bay-historical-phytoplankton-project/phytoplankton-data-sheets-1954-603803d3-346e-4b5e-b442-4212adc6c21c/transcribe/1403594), which may be unrelated.

atbah:
Closes #4097 

Fixed result:
image

### DIFF
--- a/app/assets/stylesheets/sections/page.scss
+++ b/app/assets/stylesheets/sections/page.scss
@@ -292,7 +292,8 @@
 
 .fields-row {
   display: flex;
-  flex-wrap: wrap;
+  align-items: center;
+  // flex-wrap: wrap;
 
   &:not(:last-child) {
     margin-bottom: 10px;
@@ -314,6 +315,9 @@
     label {
       display: block;
       text-align: left;
+      display: flex;
+      min-height: 42px;
+      align-items: flex-end;
     }
   }
 
--- a/app/assets/stylesheets/sections/page.scss
+++ b/app/assets/stylesheets/sections/page.scss
@@ -292,7 +292,7 @@
 
 .fields-row {
   display: flex;
-  align-items: center;
+  align-items: stretch;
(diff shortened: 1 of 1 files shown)


benwbrum:
This works perfectly for the ten field-based projects I tested.  However, I did spot something strange with one spreadsheet/field combo project:

[This transcription screen](https://fromthepage.com/bgscollections/gsm-registers/reg01/transcribe/32792459) shows a field, followed by a spreadsheet, followed by a field in production (running the development branch).  However, my local copy of this branch puts the fields and the spreadsheet on the same line:
Screenshot from 2024-05-20 22-07-03

I think there are two possibilities to research, based on my vague memory that the production code always renders a spreadsheet on its own line:
* If the project is configured correctly, with the "Register page number" field on line 1, the spreadsheet on line 2, and the "any technical issues" field on line 3, the code in this PR needs to be changed to render these fields on separate lines.
* If the project is configured incorrectly, with both fields and the spreadsheet on the same line, then it has been taking advantage of a previous bug which this code fixes.  We should not change this code, but rather modify the production project configuration--or ask the customer to do so--to correct the configuration.

Either @saracarl or I will investigate the configuration of this project.

benwbrum:
This project is configured incorrectly, so the new code is great.

Merging the PR and updating the configuration for the customer.

PR #4102, opened 2024-05-20
'Done' button auto-advances to next page

WillNigel23:


benwbrum (approved):
This looks great.  Ready to test and merge.

benwbrum:
If I hit "Done" from a transcription screen on Page 3, I am advanced to transcribing Page 4.  However, if I hit "Save" in the middle of transcribing Page 3, I am also advanced to Page 4.

The Save button should never redirect to a new page when clicked.  Both "Done" and "Approve" should advance to the next page.

See https://github.com/benwbrum/fromthepage/blob/development/app/views/transcribe/_save_buttons.html.slim for most of the permutations of the buttons and their state.

benwbrum:
Save and Approve buttons work perfectly for both Optional Review and Review Required flows. 

In testing, I realized that there is another terminal state for a page which should auto-advance the page: marking a page Blank.  Let's add that as well.

WillNigel23:
Changes made @benwbrum ready for test

PR #4108, opened 2024-05-21
4106 - Fix search in new_landing_page

WillNigel23:
Included a few refactors to improve readability and appease rubocop.


app/controllers/dashboard_controller.rb
```diff
@@ -181,7 +181,7 @@ def guest
   def landing_page
     if params[:search]
       # Get matching Collections and Docsets
-      @search_results = Collection.search(params[:search]).unrestricted + DocumentSet.search(params[:search]).unrestricted
+      @search_results = search_results
```
WillNigel23, line 184: This line is repeated in new_landing_page. Moved it to a private function

benwbrum: Would it be a bit cleaner/more functional if we passed `params[:search_results]` to the private function as an argument, so that it isn't as tightly coupled to the UI?

app/views/dashboard/new_landing_page.html.slim
```diff
@@ -1,86 +1,83 @@
--unless params[:search]
+- unless params[:search]
   .carousel
```
benwbrum, line 2: Is it preferred style to add a space between `-` or `=` in slim templates?  We've generally connected those operators to the code that follows them in the past.

Unless there's a standard that we've messed up on--which is entirely possible--we'd prefer to keep code consistent with the rest of the project.

app/views/dashboard/new_landing_page.html.slim
```diff
@@ -1,86 +1,83 @@
...
+                  - unless snippet.empty?
+                    .projects_collection_snippet = snippet
+            .projects_link = link_to t('dashboard.landing_page.more'), user_profile_path(owner)
 
     .project-list_link
-      =link_to t('.all_collections'), collections_list_path
+      = link_to t('.all_collections'), collections_list_path
 
   aside.sidecol
-    =form_tag({:controller => 'dashboard', :action => 'landing_page'}, method: :get, enforce_utf8: false, class: 'collection-search') do
```
benwbrum, line 75: Looks like you apprehended the culprit!

benwbrum (changes requested):
I've got a handful of questions/comments and a request to move a reference to `params` from the body of a method to its arguments.

benwbrum:
This fixes the original problem.  Ready to merge once review feedback is incorporated and tests pass.

app/controllers/dashboard_controller.rb
WillNigel23: Agree, will do this instead.

benwbrum (approved):
This looks great!

PR #4109, opened 2024-05-21
2416 - Improve account creation

WillNigel23:
Error messages improvements in account creation.


app/helpers/devise_helper.rb
```diff
@@ -21,8 +21,13 @@ def devise_error_messages!
...
 
-    return "" if resource.errors.empty? && flash_alerts.empty?
-    errors = resource.errors.empty? ? flash_alerts : resource.errors.full_messages
+    return '' if resource.errors.empty? && flash_alerts.empty?
+
+    errors = if resource.errors.empty?
+               flash_alerts
+             else
+               resource.errors.messages.map { |key, msg| "#{I18n.t("devise.errors.keys.#{key}")} #{msg.first}" }
+             end
```
WillNigel23, line 30: Previously, we are using full_messages. This works well for some errors, but it automatically joins the error key (ex: :login) to the message.

As a user, I found it confusing that:
1. The field label is User Name, and yet the error message says 'Login has already been taken'. 
2. The error messages will not take into account translation of the fields/keys

This line intends to decouple the key and msg and run them through translations to give more control on the error message shown which hopefully provides better account creation experience

app/models/user.rb
```diff
@@ -63,19 +63,22 @@ class User < ApplicationRecord
...
   scope :trial_owners,     -> { owners.where(account_type: 'Trial') }
   scope :findaproject_owners, -> { owners.where.not(account_type: [nil, 'Trial', 'Staff']) }
   scope :paid_owners,      -> { non_trial_owners.where('paid_date > ?', Time.now) }
   scope :expired_owners,   -> { non_trial_owners.where('paid_date <= ?', Time.now) }
   scope :active_mailers,   -> { where(activity_email: true)}
 
-  validates :login, presence: true, uniqueness: { case_sensitive: false }, format: { with: /\A[^<>]*\z/, message: "Invalid characters in username"}, exclusion: { in: %w(transcribe translate work collection deed), message: "Username is invalid"}
+  validates :login, presence: true,
+                    uniqueness: { case_sensitive: false },
+                    format: { with: /\A[^<> ]*\z/, message: I18n.t('devise.errors.messages.login.format') },
```
WillNigel23, line 75: Based on comments in Issue, it seems that adding " " spaces in username is not allowed (I assume the validation is in devise gem itself). But adding spaces on username does not trigger the format error on validation, thus I included in regex to improve on the error handling and message.

config/locales/devise/devise-en.yml
```diff
@@ -101,7 +111,7 @@ en:
         sign_up_new_account: Sign up for a new account
     sign_in: Sign In
     sign_up: Sign Up
-    user_name: User Name
+    user_name: Username
```
WillNigel23, line 114: I noticed inconsistencies in the labels and error messages for username. Let me know if we should follow User Name or Username. Still we have to make this uniform I believe.

saracarl (changes requested):
image
The format message isn't working.

WillNigel23:
Will look at it

app/models/user.rb
saracarl: I don't think the i18n reference here is working (locally for me; it does seem to be for Will).  Here's pry in my environment:
```
[12] pry(#<ActiveModel::Errors>):2> messages
=> {
       :email => ["can't be blank"],
    :password => ["can't be blank"],
       :login => ["Translation missing: en.devise.errors.messages.login.format"]
}
```

saracarl (changes requested):

benwbrum:
Is there a chance that the `validates with` section in `user.rb` is getting evaluated differently by each of you?  Since it's not a lambda, I assume that the `I18n.translate` call is happening when `user.rb` is loaded into the development environment -- as a result, one of you might have a current locale while the other wouldn't.

WillNigel23:
It might be related to this: https://github.com/ruby-i18n/i18n/issues/652

I am not encountering this in newer versions of rails so it might be fixed already. But as a workaround the issue suggested a lambda function for passing in the message.

saracarl (approved):
image

PR #4112, opened 2024-05-22
Keep tab's URL when moving from page to page

### LINKED ISSUE #3828: keep page context (tabs) when moving from page to page
Dr. Lee at Princeton suggested this:

_One suggestion is to allow us to go to the next page from the tabs we’re using. So, if I’m using the translate tab, I’d like to be able to move back and forth on the tab and not be taken back to the transcription tab and then take the extra step to find the page._

and I think it's a great idea.

atbah:
Closes #3828


atbah:
sorry, typo in issue number, closing this PR.

PR #4120, opened 2024-05-23
only return collections with works

saracarl:


app/controllers/dashboard_controller.rb
```diff
@@ -216,7 +216,7 @@ def new_landing_page
 
   def browse_tag
     @tag = Tag.where(ai_text: params[:ai_text]).first
```
WillNigel23, line 218: I think we can use `Tag.find_by(ai_text: params[:ai_text])` in here.

saracarl: done!

PR #4123, opened 2024-05-23
3768 - Fix error saving field-based page

WillNigel23:


app/models/transcription_field.rb
```diff
@@ -1,20 +1,24 @@
...
   has_many :table_cells
-  has_many :spreadsheet_columns, -> { order 'position' }, :dependent => :destroy
-  validates :options, presence: true, if: Proc.new {|field| field.input_type == 'select'}, on: [:create, :update]
+  has_many :spreadsheet_columns, -> { order 'position' }, dependent: :destroy
 
+  validates :options, presence: true, if: Proc.new { |field| field.input_type == 'select' }, on: [:create, :update]
   validates :percentage, numericality: { allow_nil: true, greater_than: 0, less_than_or_equal_to: 100 }
   validates :page_number, numericality: { allow_nil: true, greater_than: 0, less_than_or_equal_to: 1000 }
-
+  validates :label, format: { without: /[\[\]\{\}]/, message: "cannot contain '{', '}', '[' or ']'" }
```
WillNigel23, line 14: Validation for checking characters that might be interpreted as hash like '{}' or '[]'

app/views/transcription_field/_field_layout.html.slim
```diff
@@ -14,32 +14,17 @@
...
-                -options = field.options.split(";") unless field.options.nil?
-                =select_tag("fields[#{field.id}][#{field.label}]", options_for_select(options, content), class: 'field-input')
-              -elsif field.input_type == "description"
-                =hidden_field_tag "fields[#{field.id}][#{field.label}]", content, class: 'field-input'
-              -elsif field.input_type == "alt text"
-                =text_area_tag "fields[#{field.id}][#{field.label}]", content, class: 'field-input alt-text', maxlength: '255'
-                div.character-count
-              -else
-                =text_area_tag "fields[#{field.id}][#{field.label}]", content, class: 'field-input'
+              =generate_field_input(field, cell)
```
WillNigel23, line 21: I moved some of the logic here to the helper function.

app/helpers/transcription_field_helper.rb
```diff
@@ -22,4 +24,35 @@ def field_layout(array)
     @field_array = array.zip(@values)
   end
 
+  def generate_field_input(field, cell)
+    input_name = formatted_field_name(field)
+
+    label = label_tag(field.label.parameterize, field.label.titleize)
```
WillNigel23, line 30: I titleize the label for the field inputs. Let me know if it should not be like this after all

WillNigel23:
Moving back to in progress. seems there are a lot of specs failing due to changed name.

app/models/transcription_field.rb
benwbrum: Let's remove the `{}` validation -- only the hash element accessor `[]` is used by Rails to create structured parameters, `{}` isn't a problem.  (I just tested that on `development` and `{}` works.

app/views/transcription_field/_field_layout.html.slim
benwbrum: Great!

app/helpers/transcription_field_helper.rb
benwbrum: Let's not titleize the label -- many of our customers have strong opinions about this.

benwbrum (changes requested):
Remove titleize and permit curly-braces.

PR #4135, opened 2024-05-28
2532 - Assign document set as you move to a collection

WillNigel23:


app/views/work/edit.html.slim
```diff
@@ -21,6 +20,10 @@
         tr
           th =f.label :collection_id, t('.collection')
           td =f.collection_select(:collection_id, @collections, :id, :title)
+        tr
+          th =f.label :document_set_ids, t('.document_sets')
+          td#document_sets
```
WillNigel23, line 25: We add identifier to the html element whose contents we want to replace. In this case, id: 'document_sets'

app/views/work/edit.html.slim
```diff
@@ -239,3 +242,30 @@
...
+      const collectionSelect = $('#work_collection_id');
+      const documentSetsDiv = $('#document_sets');
+
+      function updateDocumentSetOptions(selectedId) {
+        $.ajax({
+          url: "#{work_document_sets_select_path}",
+          method: 'GET',
+          data: { collection_id: selectedId },
+          success: function(response) {
+            documentSetsDiv.html(response);
```
WillNigel23, line 258: We fetch to work_document_sets_select_path, it's response is an html partial. In which case we will be using it's response to replace the contents of our html target

app/views/work/edit.html.slim
```diff
@@ -239,3 +242,30 @@
...
+            $('#work_document_set_ids').multiSelect();
```
WillNigel23, line 259: Re-initialize multiSelect() as it is reloaded to html, and we only initialize on page load at application.js

app/views/work/edit.html.slim
```diff
@@ -239,3 +242,30 @@
...
+          },
+          error: function(xhr, status, error) {
+            console.error('Error fetching document sets:', error);
+          }
+        });
+      }
+
+      collectionSelect.on('change', function() {
+        const selectedId = $(this).val();
+        updateDocumentSetOptions(selectedId);
```
WillNigel23, line 269: On collection select change, we pass in the selected id which we will be using as params[:collection_id] for the controller

app/controllers/work_controller.rb
```diff
@@ -246,6 +246,12 @@ def update_featured_page
     redirect_back fallback_location: @work
   end
 
+  def document_sets_select
+    document_sets = current_user.document_sets.where(collection_id: params[:collection_id])
+
+    render partial: 'document_sets_select', locals: { document_sets: document_sets }
+  end
```
WillNigel23, line 253: The secret sauce. We have a route that, instead of rendering a full response, will render a partial instead.

WillNigel23:
@benwbrum @saracarl  Added comments in my implementation. 

PS:
This, in general, is how turbo_streams work. Instead of writing the js for replacing the dom elements with the response, it handles that automatically. Great tool with lots of application.

config/locales/work/work-de.yml
```diff
@@ -105,6 +105,7 @@ de:
       document_date: Datum des Dokuments
       document_date_hint: Datum des Dokuments ist im EDTF-Format anzugeben (z. B. 1843, 2001-02, 1643-06-30)
       document_history: Dokumentenverlauf
+      document_sets: Dokumentensätze
```
benwbrum, line 108: The German word we are using for "Documents Sets" is "Dokumentensets"

benwbrum (changes requested):
Single change for consistency in the German interface.

benwbrum:
This is tested, and works great.  We only have a single word to change in the German version.

PR #4136, opened 2024-05-29
4020 - "collaboration is restricted" for unrestricted works if not logged in

WillNigel23:


app/models/work.rb
```diff
@@ -483,4 +483,8 @@ def save_metadata
       FacetConfig.update_facets(self)
     end
   end
+
+  def user_can_transcribe?(user)
```
WillNigel23, line 487: These are just minor optimizations/logic improvements I can think of. I doubt this will make any performance improvements at all but I believe it makes more sense to add this function as well in work.rb.

My reasoning for that are:
1. When we want to check if a work can be transcribed by a user, `work` must always be present in this scenario. For our usecase, there are instances like this where current_user is not present, and thus would require extra checks like current_user && current_user.can_transcribe?(work) instead of reliably doing work.user_can_transcribe?(current_user) where current_user being nil is handled.
2. The precedence for checking if a work can be transcribed is as follows:
   `!work.restrict_scribes`
   `user.like_owner?(work)`
   `work.scribes.include?(user)`
   no need to check if user is even present if the first condition is met. So checking `current_user && current_user.can_transcribe?(work)` can be improved this way.

benwbrum (approved):
This looks like exactly what we need.

app/models/work.rb
benwbrum: That's really good thinking.  I like it!

PR #4137, opened 2024-05-29
3053 - wrap column for long text

### LINKED ISSUE #3053: Long text columns make spreadsheet view unusable
Since the spreadsheet columns resize to fit their text, long values can create columns so long that the cell is off the screen.

See examples in which the cursor is in the cell for row 3:
Usable (over-and-under is long enough):
Screenshot from 2022-03-21 14-48-33

Unusable (side-by-side is too narrow):
Screenshot from 2022-03-21 14-48-52

If there is a way to automatically resize such long columns into word-wrapped multi-line rows, that would be one way to address this.

atbah:
Closes #3053

### DIFF
--- a/app/views/shared/_handsontable.html.erb
+++ b/app/views/shared/_handsontable.html.erb
@@ -146,6 +146,8 @@
     contextMenu: ['row_above', 'row_below', 'remove_row', 'undo', 'redo', 'cut', 'copy'],
     dropdownMenu: true,
     columns: columnConfig,
+    autoWrapCol: true,
+    colWidths: 100,
     manualColumnResize: true,
     manualRowResize: true,
     startRows: <%= transcription_field.starting_rows %>,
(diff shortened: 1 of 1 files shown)


atbah:
@benwbrum   This fix involves in Handsontable's configuration.   I investigated some time and added new 2 configs to wrap up the long text.  But please have a test on your end and make sure if it works for both short/long columns.

benwbrum:
That makes sense -- we'll do some more careful testing on this, then.

benwbrum:
Atef's particular concern:
> In order to wrap up the long text, I have set column's width 100px and allowed autoWrapCol.
> I just wonder if 100px works for all cases.

benwbrum:
@atbah You were right to be cautious about the 100px.  I tried this out on two collections with long (and sometimes short) column names, and found that some column headers were truncated:

Screenshot from 2024-06-01 16-11-32
Screenshot from 2024-06-01 16-11-14

These spreadsheets were configured like this:
Screenshot from 2024-06-01 16-25-25
Screenshot from 2024-06-01 16-25-17

atbah:
okay, I will investigate more and try to find the better solution.

atbah:
@benwbrum In order to wrap up, long text in the column, we should set max width for column.  
I tested several alternative approaches but the last commit works the best I think. 
image

PR #4151, opened 2024-06-05
3358 transcribe full screen

### LINKED ISSUE #3358: Making the Transcription side of the Transcribe full screen and draggable divider
John of WWP has a feature request: a way to make the transcription side of the Transcribe tab go full screen. According to him, they have 2 use cases:
1. Some people would like to use 2 monitors and have the document take up one screen and the transcription box take up the other screen.
2. When doing a second level verification, generally they don't really need the document image.

He also requested for a draggable divider between the transcription and the document. One idea is that this would allow one to use the full screen button with the Image at the bottom layout and just make the transcription portion take up the full height.

atbah:
Closes #3358


app/views/shared/_osd_div.html.slim
```diff
@@ -9,6 +9,8 @@
   a#osd_zoom_out =svg_symbol "#icon-zoom-out", class: 'icon'
   a#osd_home =svg_symbol "#icon-fit-view", class: 'icon'
   a#osd_fit_width(title="#{t('.fit_width')}") =svg_symbol "#icon-fit-width", class: 'icon'
+  -if !@is_monitor_view
+    a#osd_monitor(title="#{t('.monitor')}") =svg_symbol "#icon-window", class: 'icon'
```
benwbrum, line 13: Should we replace this with a "close window" icon when the monitor view is being displayed?

benwbrum:
It looks like something strange may be going on with sprockets.  I first got an error message saying that I needed to add `//= link 'transcribe.js'` to `app/assets/config/manifest.js`.  

After I did that, I see this syntax error in SCSS:
```
SassC::SyntaxError at /benwbrum/ai-assist-sandbox/emily-caroline-creaghe-diary-22-dec-1882-5-sept-1883-2946fa12-d861-43f8-95be-6a412b2d1286/transcribe/33267572
Error: Undefined variable: "$inputDisabledColor".
        on line 27:29 of app/assets/stylesheets/splitter.scss
>>     border-left: 2px dashed $inputDisabledColor;
   ----------------------------^
   
```

app/views/shared/_osd_div.html.slim
atbah: yeah, that would be great and what I had been thinking of. 
I will try to change the icon using Javascript.

benwbrum:
This seems to work great now!

Let's add the window closing icon to the monitor view (which I didn't see locally when I tested) and attempt to persist the window-drag dimensions between one page and another, the same way we do with zoom.

On the second change, the use case here is that someone generally uses their second monitor to look at the image, so they have reduced the size of the image viewer portion of the transcription screen to something minimal, so that they have more room to type.  If they save a page and navigate to the next one, we don't want the to have to re-resize their transcription window again.

atbah:
okay, let me figure them out.

atbah:
@benwbrum I pushed a new commit for showing close icon.  Please have a look

benwbrum:
The close icon works really well.  We still don't save the slider position, however: 
* Navigate to a transcripton page
* Move the slider to minimize the image viewer and maximize the transcription area
* Click on the error to the next page
* Observe that your slider changes are lost

atbah:
Would you attach a screen record for the slider position issue?  
I couldn't generate that issue on my end.

benwbrum:
Strange, it works fine for me now.  Merging.

PR #4153, opened 2024-06-06
2527 - Activity emails shouldn't link to "transcribe" page

WillNigel23:
Followed notes from @sylvieed


WillNigel23:
Specs fixed @benwbrum @saracarl

PR #4154, opened 2024-06-06
2682 - Error in subject versions tab when no display name

WillNigel23:


WillNigel23:
Specs fixed @benwbrum @saracarl !

benwbrum:
Apparently there is more than one place where we look at `version.user.display_name`, and the second one still breaks.

I tried testing this as follows:
* Find a collection with subjects
* View a subject
* View the Versions tab
* Verify that the current user is displayed
* `rails c`
* `article = Article.find 9422` (or whatever the ID of the subject is)
* `version = article.article_versions.last`
* `version.user_id = -4`
* `version.save!`
* Restart the server
* Reload the version screen


```
NoMethodError - undefined method `display_name' for nil:NilClass:
  app/views/article_version/show.html.slim:29
  app/views/article_version/show.html.slim:26
  app/views/article_version/show.html.slim:25
  app/controllers/article_version_controller.rb:11:in `list'
  app/controllers/application_controller.rb:64:in `switch_locale'
```

Issue #4070, opened 2024-05-01
"done" button should auto-advance


saracarl:
If you click "Done" you should be taken to the next page to transcribe.

@anabeabas worth thinking about on the save/done usability.

PR #4155, opened 2024-06-06
2651 - Page state after checking needs review

WillNigel23:
Most of the ground work here has been done in #4070 so I based off my branch there, will rebase once done.


app/views/transcribe/_save_buttons.html.slim
```diff
@@ -2,27 +2,24 @@
...
-    -else
-      =button_tag t('.save_changes'), :name => 'save_to_transcribed', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_transcribed_tooltip')}", class: "bggreen", data: { disable_with: t('.please_wait') }
+    =button_tag t('.save_changes'), name: 'save_to_needs_review', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_needs_review_tooltip')}", data: { disable_with: t('.please_wait') }
 
-  -if [nil, "", Page::STATUS_INCOMPLETE].include?(@page.status)
-    -if @collection.review_workflow
-      =button_tag t('.done'), :name => 'save_to_needs_review', type: 'submit', id: 'finish_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.finish_to_needs_review_tooltip')}", class: "bggreen", data: { disable_with: t('.please_wait') }
+  -if [nil, "", Page::STATUS_NEEDS_REVIEW, Page::STATUS_INCOMPLETE].include?(@page.status)
+    -if @collection.review_workflow && @page.status != Page::STATUS_NEEDS_REVIEW
```
WillNigel23, line 19: Based on descriptions, I don't think we need a separate 'review_incomplete' status. We just have to double check that if review_workflow == true && status != review for the condition here

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,124 @@ def needs_review
...
+
+    # check to see if the page needs to be marked as needing review
     needs_review
-    if params['save'] || params['save_to_incomplete'] || params['save_to_needs_review'] || params['save_to_transcribed'] || params['approve_to_transcribed']
+
+    save_to_incomplete = params[:save_to_incomplete] || params[:done_to_incomplete]
+    save_to_needs_review = params[:save_to_needs_review] || params[:done_to_needs_review]
+    save_to_transcribed = params[:save_to_transcribed] || params[:done_to_transcribed]
+    approve_to_transcribed = params[:approve_to_transcribed]
+
```
WillNigel23, line 145: in #4070 I have introduced separating the `save_to` and `done_to` buttons so we know when to progress to the next page when clicked save or done.

Same applies in here, where having clear distinction between saving and clicking done can simplify things rather than introduce a new status `review_incomplete`

benwbrum: I think that I like that differentiation.  Nice job on the renaming -- this is much clearer.

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,126 @@ def needs_review
...
+    if params['save'] || save_to_incomplete || save_to_needs_review || save_to_transcribed || approve_to_transcribed
       message = log_transcript_attempt
-      #leave the status alone if it's needs review, but otherwise set it to transcribed
 
-      if params['save_to_incomplete'] && params[:page]['needs_review'] != '1' 
+      # leave the status alone if it's needs review, but otherwise set it to transcribed
+      if save_to_incomplete && params[:page]['needs_review'] != '1'
         @page.status = Page::STATUS_INCOMPLETE
-      elsif params['save_to_needs_review']
+      elsif params[:done_to_needs_review] && @page.work.collection.review_workflow
```
benwbrum, line 152: Reconsidering the renaming; if we are in the optional workflow, can there ever be a `done_to_needs_review` scenario?

benwbrum: Does it make sense to be checking `params[:done_to_needs_review]` here when have variables we're using elsewhere?

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,126 @@ def needs_review
...
-          #don't reset subjects if they're disabled
-          unless @page.collection.subjects_disabled || (@page.source_text.include?("[[") == false)
+
+          # don't reset subjects if they're disabled
+          unless @page.collection.subjects_disabled || (@page.source_text.include?('[[') == false)
             # use the new links to blank the graphs
             @page.clear_article_graphs
 
             new_link_count = @page.page_article_links.where(text_type: 'transcription').count
             logger.debug("DEBUG old_link_count=#{old_link_count}, new_link_count=#{new_link_count}")
```
benwbrum, line 182: Let's remove this debug statement; we haven't used it for a decade.  (If that makes the `_link_count` variables redundant, we can remove them too.)

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,126 @@ def needs_review
...
-            if old_link_count == 0 && new_link_count > 0
-              record_deed(DeedType::PAGE_INDEXED)
-            end
-            if new_link_count > 0 && @page.status != Page::STATUS_NEEDS_REVIEW && @page.status != Page::STATUS_INCOMPLETE
+
+            record_deed(DeedType::PAGE_INDEXED) if old_link_count.zero? && new_link_count.positive?
```
benwbrum, line 184: Ah -- that's why we might need those variables...

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,126 @@ def needs_review
...
+          # if this is a guest user, force them to sign up after three saves
           if current_user.guest?
             deeds = Deed.where(user_id: current_user.id).where(deed_type: DeedType.edited_and_transcribed_pages).count
             if deeds < GUEST_DEED_COUNT
               flash[:notice] = t('.you_may_save_notice', guest_deed_count: GUEST_DEED_COUNT)
             else
-              session[:user_return_to]=collection_transcribe_page_path(@collection.owner, @collection, @work, @page.id)
-              redirect_to new_user_registration_path, :resource => current_user
+              session[:user_return_to] = collection_transcribe_page_path(@collection.owner,
+                                                                         @collection, @work, @page.id)
```
benwbrum, line 204: If parameters in a method call don't all fit on one line, let's put each of them on their own line.

app/controllers/transcribe_controller.rb
```diff
@@ -123,113 +123,126 @@ def needs_review
...
-              redirect_to collection_sampling_review_page_path(@collection.owner, @collection, @quality_sampling, next_page.id, flow: "quality-sampling")            
+              redirect_to collection_sampling_review_page_path(@collection.owner,
+                                                               @collection, @quality_sampling,
+                                                               next_page.id, flow: 'quality-sampling')
             else
               redirect_to collection_quality_sampling_path(@collection.owner, @collection, @quality_sampling)
             end
           else
-            redirect_to :action => 'assign_categories', page_id: @page.id, collection_id: @collection
+            save_button_clicked = params[:save_to_incomplete] || params[:save_to_needs_review] ||
```
benwbrum, line 225: Would it make sense to move this variable definition earlier?  Do we need it for anything else?

app/views/transcribe/_save_buttons.html.slim
```diff
@@ -2,27 +2,24 @@
...
+    =button_tag t('.save_changes'), name: 'save_to_incomplete', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_incomplete_tooltip')}", data: { disable_with: t('.please_wait') }, onclick: 'unsavedNotes(event);'
 
   -if [Page::STATUS_NEEDS_REVIEW, Page::STATUS_INDEXED, Page::STATUS_TRANSCRIBED].include?(@page.status)
-    -if @collection.review_workflow
-      =button_tag t('.save_changes'), :name => 'save_to_needs_review', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_needs_review_tooltip')}", data: { disable_with: t('.please_wait') }
-    -else
-      =button_tag t('.save_changes'), :name => 'save_to_transcribed', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_transcribed_tooltip')}", class: "bggreen", data: { disable_with: t('.please_wait') }
+    =button_tag t('.save_changes'), name: 'save_to_needs_review', type: 'submit', id: 'save_button_top', onclick: 'unsavedNotes(event);', title: "#{t('.save_to_needs_review_tooltip')}", data: { disable_with: t('.please_wait') }
```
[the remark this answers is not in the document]
benwbrum: One thing that we need to discuss--possibly not now--is what the proper behavior is when someone edits a page that has been approved in the Review workflows.  Currently that makes the page need review again, but our users hate that and want post-approval edits to not need further approval.

config/locales/display/display-es.yml
```diff
@@ -45,6 +45,7 @@ es:
     multi_page:
       categories: Categorías
       describe: Crear metadatos
+      pages_need_completion: Páginas que necesitan completarse
```
benwbrum, line 48: I suspect that this should be something like "Necessitan Completión" to make it parallel with the other Spanish.  (We should check this, however -- my Spanish is pretty bad.)

config/locales/display/display-pt.yml
```diff
@@ -45,6 +45,7 @@ pt:
     multi_page:
       categories: Categorias
       describe: Criar metadados
+      pages_need_completion: Páginas que precisam ser completadas
```
benwbrum, line 48: I dont' speak Portugese, but the other examples are "para ???ar" or "para ???er".  We should look at that here.

config/locales/transcribe/transcribe-es.yml
```diff
@@ -59,7 +59,6 @@ es:
       save_changes: Guardar Cambios
       save_to_incomplete_tooltip: Guardar una página incompleta
       save_to_needs_review_tooltip: SGuardar tus cambios
```
benwbrum, line 61: It looks like there is a typo in this line.  Let's replace `SGuardar` with `Guardar` to match the others.

benwbrum (changes requested):
This looks great!  I had a couple of questions, and see some issues in the translation files (including one that already existed).

app/views/transcribe/_save_buttons.html.slim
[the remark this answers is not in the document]
WillNigel23: That may be annoying indeed, but at the same time we can argue that that's the purpose of review_workflows afterall. Any changes made should be under review once more.

Might be worth adding another checkbox for review workflows that were complete already, a skip-review feature. But I don't think that's relevant in this particular ticket

app/controllers/transcribe_controller.rb
WillNigel23: I don't think we need it for anything else right now.

app/views/transcribe/_save_buttons.html.slim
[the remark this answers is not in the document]
benwbrum: Agreed.  I'll open a new issue to track that.

app/controllers/transcribe_controller.rb
benwbrum: You're right -- `done_to_needs_review` will not be possible in the optional workflow.

PR #4169, opened 2024-06-19
4158 - Fix work pages tab link to page throws 500 error

WillNigel23:
Upon further investigation, it seems there is an enable messageboards in settings. Problem found is, when this is disabled we still render the Thredded links that needs a message board thus the error.

We can prevent the error by wrapping the breadcrumbs in a conditional to check first if @collection.messageboards_enabled? But this is just me assuming what's most logical in this scenario, I may be wrong in my understanding of messageboards so please do correct me if so @saracarl @benwbrum


saracarl:
Where does it link if forums are enabled?  Does this actually involve linking to forums, or is the error a side affect?  This link shouldn't go to the forum, so I'm confused why thredded is involved.

WillNigel23:
After clicking on said link we get here

image


As you can see in the image we have the threaded there, but we render it without checking if the messageboards is enabled.

benwbrum:
This no longer throws the 500 error, however if we enable forums, navigate to the collection pages tab, then click on a page title, we see the page contents correctly but the tabs and breadcrumbs seem to think that we are in a forum.

It is likely that this has something to do with the special code we added to the messageboard lists that let a user navigate to the page notes from the Forum tab, so we'll need to test leaving notes on a page, navigating to the forum, then clicking through to the page note and looking at breadcrumbs/tabs.

Screenshot from 2024-06-20 16-46-45

WillNigel23:
Can you please point me to the relevant PR and provide more details? This is likely a bug as you mentioned, as it seems the thredded links should not appear here am I correct to understand that?

benwbrum:
Here's the place where we link to the number of notes for a collection within Thredded forums:
https://github.com/benwbrum/fromthepage/pull/3345/files#diff-376219ac35a2792eb8e9e7f2b20713cb0e4839046171d14179e4a9742d5bf07eR28-R36

Here is the place where we display the list of notes as if they were part of a Thredded forum: https://github.com/benwbrum/fromthepage/pull/3345/files#diff-90699ecd7c314cd351e35990759014bd9617dbc32794ae8abb6b0391f8259cefR28

PR #4170, opened 2024-06-20
Added model annotations

WillNigel23:


WillNigel23:
@benwbrum purely annotations, should have no effect on specs/functionalities

benwbrum:
I really like this.  Will it be updated automatically whenever migrations are run?

WillNigel23:
Yes, it will run as we do db/migrate

PR #4173, opened 2024-06-24
4152 - Make export work pagination searchable

WillNigel23:


WillNigel23:
@benwbrum you are correct, it does not work with paginate.

I have 2 possible solutions for this:
1. Turn off pagination like you mentioned
2. Custom JS. We fetch from db when search is made, then use the response to change contents without page reload. (Again this is touching how turbo_stream works in general). But I don't think this is advisable unless we actually commit to using that as without the gem and library the implementation will be very hacky (although doable)

benwbrum:
Hmm.  Take a look at https://github.com/benwbrum/fromthepage/pull/3792 for a third option.

WillNigel23:
I see. It follows the same concept as turbo_streams as the one I suggested earlier, except it uses ajax. That should work

Issue #4005, opened 2024-03-27
Page count for subject out of sync after deduplication merge


benwbrum:
Jon at WWP reports that subject page counts are off after a deduplication merge.  This appears to be a result of the Rails counter cache call-back not being called when we update the `page_article_links` in the article text rename/page save phase of article combining.

```ruby
{"id"=>89811,
 "title"=>"Alexander Badlam, b. 1808",
 "source_text"=>"",
 "created_on"=>Tue, 02 Feb 2021 00:59:38.000000000 UTC +00:00,
 "lock_version"=>1,
 "xml_text"=>
  "<?xml version='1.0' encoding='UTF-8'?>    \n" +
  "      <page>\n" +
  "        <p/>\n" +
  "      </page>\n",
 "graph_image"=>
  "/home/fromthepage/deployment/releases/20240325204529/public/images/working/dot/89811.png",
 "collection_id"=>970,
 "latitude"=>nil,
 "longitude"=>nil,
 "uri"=>"",
 "provenance"=>nil,
 "created_by_id"=>221669,
 "pages_count"=>126}
 => 0 
2.7.3 :028 > to_article.pages.count
 => 172 
2.7.3 :029 > to_article.pages.uniq.count
 => 143 
2.7.3 :030 > to_article.pages_count
 => 126 
```

saracarl:
This also is a problem when someone is updating a link on page:
subject A is linked withing a page transcription
user updates that to instead link to subject B
then both A and B will have incorrect page counts.

We suspect if we figure out what's going on there, the dedupe problem outlined above will also be fixed.  Both seem to be in the page save callback.  (We thought that should be automated, but we're not sure how that works with had_and_belongs_to_many; we haven't used counter caches a lot.)

WillNigel23:
Seems like the counter_cache is bugged with has_many :through relationships. I would say the approach should be a callback to the article model itself to update the counts.

benwbrum:
Do you think that this bug has been fixed in newer versions of Rails?  If so, I'd prefer to put our effort into upgrading our framework rather than doing work-arounds here.

WillNigel23:
I tried and confirmed this behaviour is working correctly in ruby 3.3.3, and it works with update callbacks as well.

benwbrum:
Hey, does this work now that we've upgraded to ruby 3.3.5?

WillNigel23:
Let me refresh my memory about this and get back to your comment. As per our previous discussion it should be.

Issue #3993, opened 2024-03-20
Orphan article-article links cause 500 error


saracarl:
When going to this subject page:
https://fromthepage.com/gri/stendahl-letters-qc/article/32081279
The customer gets the following 500 error:
```
I, [2024-03-20T17:55:01.594240 #11487]  INFO -- : Started GET "/gri/stendahl-letters-qc/article/32081279" for 136.62.254.224 at 2024-03-20 17:55:01 +0000
I, [2024-03-20T17:55:01.596616 #11487]  INFO -- : Processing by ArticleController#show as HTML
I, [2024-03-20T17:55:01.596898 #11487]  INFO -- :   Parameters: {"user_slug"=>"gri", "collection_id"=>"stendahl-letters-qc", "article_id"=>"32081279"}
I, [2024-03-20T17:55:01.746109 #11487]  INFO -- :   Rendered article/show.html.slim within layouts/application (Duration: 76.8ms | Allocations: 39797)
I, [2024-03-20T17:55:01.746930 #11487]  INFO -- :   Rendered layout layouts/application.html.slim (Duration: 77.7ms | Allocations: 40499)
I, [2024-03-20T17:55:01.748097 #11487]  INFO -- : Completed 500  in 151ms (ActiveRecord: 29.1ms | Allocations: 52682)
F, [2024-03-20T17:55:01.759191 #11487] FATAL -- : 
ActionView::Template::Error (undefined method `title' for nil:NilClass):
    19:     -@article.target_article_links.each do |link|
    20:       li
    21:         =svg_symbol '#icon-page', class: 'icon'
    22:         =link_to link.source_article.title, collection_article_show_path(@collection.owner, @collection, link.source_article.id)
    23:         =="&nbsp;&mdash;&nbsp;#{link.display_text}"
  
app/views/article/_article_links.html.slim:22
app/views/article/_article_links.html.slim:19
app/views/article/show.html.slim:42
app/controllers/application_controller.rb:64:in `switch_locale'
```

Reported by Peyton at GRI.

saracarl:
I think this is caused by `source_article` reference this line:
https://github.com/benwbrum/fromthepage/blob/development/app/views/article/_article_links.html.slim#L22
[...]

Issue #3576, opened 2023-03-27
Orphan article-article or page-article links?


benwbrum:
Viewing this article yields the following error: http://localhost:3000/mountauburncemetery/business-correspondence/article/71294

```
I, [2023-03-27T09:10:16.328835 #8419]  INFO -- : Completed 500  in 99ms (ActiveRecord: 18.1ms | Allocations: 41078)
F, [2023-03-27T09:10:16.337426 #8419] FATAL -- :   
ActionView::Template::Error (undefined method `title' for nil:NilClass):
    22:     -@article.target_article_links.each do |link|
    23:       li
    24:         =svg_symbol '#icon-page', class: 'icon'
    25:         =link_to link.source_article.title, collection_article_show_path(@collection.owner, @collection, link.source_article.id)
    26:         =="&nbsp;&mdash;&nbsp;#{link.display_text}"
  
app/views/article/_article_links.html.slim:25
app/views/article/_article_links.html.slim:22
```

When I view `link.source_article` it returns `nil`.  Is this an article-article link instead of a page-article link?  Are we not clearing them out properly?  What's allowing the data to get into this state?

PR #4175, opened 2024-06-25
Clean orphaned page article link records

WillNigel23:


app/models/article.rb
```diff
@@ -41,21 +33,20 @@ class Article < ApplicationRecord
...
-
   has_and_belongs_to_many :categories, -> { distinct }
   belongs_to :collection, optional: true
   has_many(:target_article_links, :foreign_key => "target_article_id", :class_name => 'ArticleArticleLink')
   scope :target_article_links, -> { include 'source_article' }
   scope :target_article_links, -> { order "articles.title ASC" }
 
   has_many(:source_article_links, :foreign_key => "source_article_id", :class_name => 'ArticleArticleLink')
-  has_many(:page_article_links)
+  has_many :page_article_links, dependent: :destroy
```
WillNigel23, line 43: This callback is in page.rb, but seems missing on article side.

db/migrate/20240625185153_clean_orphan_page_link_records.rb
```diff
@@ -0,0 +1,9 @@
+class CleanOrphanPageLinkRecords < ActiveRecord::Migration[6.1]
+
+  def up
+    page_article_links = PageArticleLink.where(article_id: nil).or(PageArticleLink.where(page_id: nil))
+
+    page_article_links.in_batches(of: 1000).destroy_all
```
WillNigel23, line 6: migration to clean existing bad records

app/models/page.rb
```diff
@@ -530,7 +529,7 @@ def clear_links(text_type)
     if self.page_article_links.present?
       self.clear_article_graphs
       # clear out the existing links to this page
-      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").delete_all
+      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").destroy_all
```
WillNigel23, line 532: Doing delete_all does not trigger active_record callbacks , therefore no validations.

Moving forward, unless we specifically want to avoid callbacks, we should use destroy instead of delete in order to avoid unvalidated changes.

WillNigel23:
#3993 and #3576

app/models/article.rb
```diff
@@ -81,7 +72,7 @@ def source_text
 
   def self.delete_orphan_articles
     # don't delete orphan articles with contents
-    Article.delete_all("source_text IS NULL AND id NOT IN (select article_id from page_article_links)")
+    Article.where('source_text IS NULL AND id NOT IN (SELECT article_id FROM page_article_links)').destroy_all
```
saracarl, line 75: I think this has the same issue caveat as #4217 :
Project owners can upload canonical subjects via a spreadsheet, and we don't want to delete those if they are not used.

If the subject record was created via a spreadsheet upload, then the Article record will have a provenance of the filename they uploaded. Every subject that was created via the transcription process will have a nil provenance. So when we "clean up orphaned subjects" at page save, we need to only delete articles with a nil provenance and no links.
(In production 6000/200,000 have a provenance.)

app/models/article.rb
```diff
@@ -146,7 +137,7 @@ def possible_duplicates
   def clear_links(type='does_not_apply')
     # clear out the existing links to this page
     if self.id
-      ArticleArticleLink.where("source_article_id = #{self.id}").delete_all
+      ArticleArticleLink.where("source_article_id = #{self.id}").destroy_all
```
saracarl, line 140: Same here.

db/migrate/20240625185153_clean_orphan_page_link_records.rb
saracarl: Same caveat here.

saracarl (changes requested):

saracarl:
Oops.  Just realized I'm wrong here -- we're not deleting Articles, but links.  Please ignore my review....

saracarl (commented):
Re-review to see if this might be causing #4269

app/models/article.rb
```diff
@@ -41,21 +33,20 @@ class Article < ApplicationRecord
...
-  has_many(:page_article_links)
+  has_many :page_article_links, dependent: :destroy
   scope :page_article_links, -> { includes(:page) }
   scope :page_article_links, -> { order("pages.work_id, pages.position ASC") }
 
-  scope :pages_for_this_article, -> { order("pages.work_id, pages.position ASC").includes(:pages)}
+  scope :pages_for_this_article, -> { order("pages.work_id, pages.position ASC").includes(:pages) }
 
-  has_many :pages, through: :page_article_links, counter_cache: true
+  has_many :pages, through: :page_article_links
```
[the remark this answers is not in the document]
WillNigel23: This is a bug in current rails version. This ticket provides more discussions on that

https://github.com/benwbrum/fromthepage/issues/4005

PR #4177, opened 2024-06-25
3624 - Upgrade rails to current version

### LINKED ISSUE #3624: Upgrade Ruby and Rails to current versions
We should upgrade both Ruby and Rails to current versions.  This may also help with #3623

WillNigel23:


benwbrum:
Obviated by #4237

Issue #1682, opened 2020-04-01
500 error viewing collection with no pages


benwbrum:
The "Maryland State Archives: 1978 Marriage Certificates" collection has all (I think) of it's pages deleted out of works (on purpose), but now if you go to the collection homepage or the settings you get a 500 error.

They'd like to delete the collection, but can't.  Can we just do this via the console?  Or make it private?

benwbrum:
This sounds like a problem with the `next_untranscribed_page` value on the `@collection`, but I might be wrong.

benwbrum:
Let's try to duplicate this first, by creating a collection, importing a work, transcribing a single page, then deleting the work's pages, then deleting the work.  View the collection landing page between each step.

WillNigel23:
@benwbrum it seems deleting pages/works thru UI it handles removing references to next_untranscribed_page both to work and collections. So IF the problem is indeed because of that then the pages were deleted in CLI where the after_destroy callbacks are not handled.

However I tried doing the same, where the page is deleted with the collection.next_untranscribed_page, but I experience no errors. 

If possible, can we get have the actual error here to debug easier?

benwbrum:
I'll get the exception for you.

benwbrum:
Just duplicated it and got this exception.  (I don't understand why the Completed 500 appears before the exception message, though.)

```
I, [2024-06-26T23:06:22.264499 #636467]  INFO -- : Oink Log Entry Complete
I, [2024-06-26T23:06:22.278382 #636467]  INFO -- : Started GET "/msa/marriage-certificates" for 136.62.254.224 at 2024-06-26 23:06:22 +0000
I, [2024-06-26T23:06:22.280939 #636467]  INFO -- : Processing by CollectionController#show as HTML
I, [2024-06-26T23:06:22.281187 #636467]  INFO -- :   Parameters: {"user_slug"=>"msa", "id"=>"marriage-certificates"}
I, [2024-06-26T23:06:22.287201 #636467]  INFO -- :   Rendered inline template (Duration: 1.0ms | Allocations: 81)
I, [2024-06-26T23:06:22.407696 #636467]  INFO -- :   Rendered collection/show.html.slim within layouts/application (Duration: 107.9ms | Allocations: 45823)
I, [2024-06-26T23:06:22.408505 #636467]  INFO -- :   Rendered layout layouts/application.html.slim (Duration: 109.1ms | Allocations: 46577)
[...]

PR #4186, opened 2024-06-27
1682 - Error when viewing notes with no page

WillNigel23:
#1682 

It seems the issue is regarding orphaned notes (when page_id: nil).

Approach done here is to render plain text instead of link when note.page.nil

BUT to be more elaborate, need answer to some of these questions:
SHOULD we allow page_id to be nullable in note? (In my local db dump there are 0 notes with page_id: nil. Looking at the notes model though it seems that it can be made for a work and not a page, not sure how this all works so please clarify).

### DIFF
--- a/app/datatables/note_datatable.rb
+++ b/app/datatables/note_datatable.rb
@@ -57,7 +57,7 @@ class NoteDatatable < AjaxDatatablesRails::ActiveRecord
           userpic:    userpic(record),
           user:       link_to(record.user&.display_name, user_profile_path(record.user)),
           note:       record.title,
-          page:       link_to(record.page.title, collection_display_page_path(record.collection.owner, record.collection, record.work, record.page)),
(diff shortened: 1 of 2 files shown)


benwbrum:
I think that you're probably right here.  While I can imagine a couple of scenarios in which users might want the comments they left on a project to be retained even after a work was deleted, we have no way to read them in the context of the page.

That said, users _can_ leave notes on a work through the metadata description feature, which would result in a note with no `page_id`, so on second thought this feature may actually be needed.

benwbrum:
While testing this PR by adding a note to the metadata screen on a metadata-and-transcription project, I spotted a bug or feature gap displaying the list of notes for a project if it included a note with no page (which is what happens when a note is added to a metadata screen) but the project was not metadata-only.

@WillNigel23 can you review this?

WillNigel23 (commented):
@benwbrum your recent changes looks good to me :green_circle:

PR #4187, opened 2024-06-27
4079 - UI to automate account upgrade

WillNigel23:


app/views/dashboard/_owner_header.html.slim
```diff
@@ -43,20 +43,20 @@
...
       |&nbsp;&middot;&nbsp;
       =t('.total_pages', pages: current_user.page_count)
-    -if (current_user.account_type == 'Trial') 
-      span 
-        |&nbsp;&middot;&nbsp; 
-        =t('.to_upgrade_contact')
+    -if (current_user.account_type == 'Trial')
+      span
+        |&nbsp;&nbsp;
+        =link_to t('.upgrade'), UPGRADE_FORM_LINK, class: 'button outline', target: '_blank'
```
saracarl, line 59: Can we change this so if there's no UPGRADE_FORM_LINK then "upgrade" isn't shown?  Seems more friendly to our open source users, and gives them control.

saracarl (changes requested):
One small change, but otherwise looks good!

PR #4188, opened 2024-06-27
4126 - Hide pages needs review when count is zero

WillNigel23:


app/views/collection/show.html.slim
```diff
@@ -116,14 +116,19 @@
       =label_tag 'search', t('.search_by_title_1'), class: 'hidden'
 
     -if @collection.active? && @collection.text_entry?
+      ruby:
+        pages_need_correction_count = Page.where(work_id: @collection.work_ids, status: [nil, Page::STATUS_INCOMPLETE]).count
+        pages_need_review_count = Page.where(work_id: @collection.work_ids, status: Page::STATUS_NEEDS_REVIEW).count
```
WillNigel23, line 121: Although this accomplishes the task, I think it might be best not to do this?

Question would be do we want this feature even at the cost of performance?

That being said I'm thinking of more optimal way to do this, let me know if you have ideas as well

benwbrum: We roll up this kind of data in the `work_statistics` table, which is a child of `works` and gets re-calculated on any page status change.  Rewriting this do something like this (pseudocode) would be worth the performance trade-off, even if the we have to add the computations (and migrations) to `WorkStatistic`.
`@collection.joins([:works, :work_statistics]).where("work_statistic.needs_review_pages > 0").present?`

WillNigel23: @benwbrum I see no issue for pages_need_review_count. However for pages_need_correction I think we need other than incomplete_pages_count correct?

I think no change is needed here after all. We can perhaps test the performance first before deciding if this needs further optimization?

saracarl (approved):
This effectively resolves the original issue.  If @benwbrum and @WillNigel23 are OK with the performance I think it can be merged.

I also opened #4213 to capture almost exactly the same scenario at the work screen.

app/views/collection/show.html.slim
benwbrum: I think that we can use `work_statistics` in this case by checking whether the sum of `work_statistics.total_pages` - (`work_statistics.completed_pages` + `work_statistics.needs_review_pages`) is non-zero.  It is not clear to me which approach would work better, but doing a couple of tests on a reasonably large collection and observing the rack miniprofiler would probably let us decide.

WillNigel23: I think doing the current one is much more efficient/accurate. The rollup counts does not seem to match the actual records on some particular works. I outlined my reasoning in slack, let me know if you have any other questions @benwbrum

benwbrum:
I just tested this locally on a work with one page complete, one not started, and no pages in needs review.  All three "Pages needing..." buttons showed up at the work level. even though no pages needed review or completion.

Screenshot from 2024-07-18 19-33-24
Screenshot from 2024-07-18 19-33-10

PR #4206, opened 2024-07-04
4202 fix splitter(vertical) CSS

atbah:


saracarl:
To test:
text, spreadsheet, field based, metadata
with all 4 orientations of the image/input: over, under, left, right
pay attention to scrolls
go to "next page" and ensure it's the same

saracarl (commented):
Not through my review yet, but this is a problem:
image
This only seems to occur in the "image at top" view.  I tried about 2/3 image to 1/3 text field, starting "image on the left" and then switched to "image at top" to get to this.  The splitter shows up in the middle of the image.  It seems to stay there, although one other time it went away(??).

saracarl (commented):
Ok, next problem:
Metadata creation lacks the slider and the borders have slipped:
image
This is on a page like: http://localhost:3000/dartmouthlibrary/wrangel-island-set/stefansson-wrangel-09-06-001/describe
(You will probably need to create some test data, since it's rare.  Here's how:
https://content.fromthepage.com/project-owner-documentation/configuring-metadata-description/)

saracarl (commented):
Translation is also missing the slider:
image
URL here:  http://localhost:3000/benwbrum/thornton-manuscripts/the-prose-life-of-alexander/translate/6295

Turn it on for testing in the settings for the *work* (not the collection, it's lower level) look for "Enable Translation".

saracarl (changes requested):
Ok, I've reviewed all the task types, in the 4 transcription pane orientations.  The above comments cover what I've found.  

It's a super cool feature and folks are going to love it once we finish it.

This concludes the review!

atbah:
Thank you for your review, @saracarl 
I will fix all of your comments ASAP.

saracarl (approved):

PR #4212, opened 2024-07-09
3193 autocomplete for subject linking

### LINKED ISSUE #3193: Advanced editor autocomplete for subject linking
Jon Fackrell suggests:

> We have a large and growing list of Topics (Subjects) and it's only growing. We also have a lot of volunteers and cycle through interns regularly and we're finding that it's difficult for them to master the list. They might not know exactly what's on the list, misspell words, or just tag using the plural vs singular form.
> 
> I'm wondering if you have any quick suggestions for helping with this, but I'm also wondering what the possibility is that you might be able to add an autocomplete popup with a list of subjects. I picture it working like Slack and other apps where you type the '@' and then start typing a username and then select from a filtered list. The main difference being that it would popup when you type '[[' and then let you filter subjects and allow you to insert it.

cf. #3064

atbah:
Closes #3193


benwbrum:
Example of "Henry" not narrowing down options to titles containing "Henry":

https://github.com/user-attachments/assets/f738b01b-b00f-4aa7-9095-30c0a9470bee

PR #4217, opened 2024-07-10
2313 - Abandon subject without losing transcription

WillNigel23:
Discussions so far with @benwbrum 

I'm playing around the functionality. Correct me in if I am wrong here, but I think these should be the cases:

1. I want to go back and make some changes
    a. We click on cancel button. This should not assign a category, BUT I would assume this should also NOT create the subject, otherwise why cancel in the first place.
    
2. I want to save my changes
    a. We click on save or continue button. THIS should assign categories and redirect back to same page (in case we want to edit some texts, add new subjects, etc).
    b. Regardless if they assign the category here, the subject SHOULD persist.
    
3. I want to save and continue to next page
    a. It creates and assigns subject and redirects to next page
    
I believe this should be the case, as was described by the user on the ticket

> I just spelled it wrong or left out a punctuation so I don't actually want to be creating a new subject...

Clicking cancel should NOT persist the subject at hand. 

That being said, I first went ahead and implement what was instructed in ticket where:
1. Clicking cancel will take me back to editing page and will not save assigned categories (but subjects will persist)
2. Clicking continue should redirect to the next appropriate page.

I took out the third button named `next` This is because past ticket made the `continue` dependent on the workflow. i.e. If save was clicked it should remain in same page, if done was clicked it will go to next page. So I believe only 2 buttons are relevant here (cancel and continue).

Tagging @saracarl  as well

### DIFF
--- a/app/controllers/transcribe_controller.rb
+++ b/app/controllers/transcribe_controller.rb
@@ -28,6 +28,8 @@ class TranscribeController  < ApplicationController
   end
 
   def display_page
+    rollback_article_categories(params[:rollback_delete_ids], params[:rollback_unset_ids])
+
     @collection = page.collection unless @collection
     @auto_fullscreen = cookies[:auto_fullscreen] || 'no';
     @layout_mode = cookies[:transcribe_layout_mode] || @collection.default_orientation
@@ -128,6 +130,7 @@ class TranscribeController  < ApplicationController
 
   def save_transcription
     old_link_count = @page.page_article_links.where(text_type: 'transcription').count
+    old_article_ids = @page.articles.pluck(:id)
 
     @quality_sampling = QualitySampling.find(params[:quality_sampling_id]) if params[:quality_sampling_id].present?
 
@@ -148,7 +151,7 @@ class TranscribeController  < ApplicationController
     save_to_needs_review = params[:save_to_needs_review] || params[:done_to_needs_review]
     save_to_transcribed = params[:save_to_transcribed] || params[:done_to_transcribed]
     approve_to_transcribed = params[:approve_to_transcribed]
-    
+
     if params['save'] || save_to_incomplete || save_to_needs_review || save_to_transcribed || approve_to_transcribed
       message = log_transcript_attempt
       # leave the status alone if it's needs review, but otherwise set it to transcribed
@@ -240,7 +243,7 @@ class TranscribeController  < ApplicationController
 
             next_page_id = @page.last? || save_button_clicked ? @page.id : @page.lower_item.id
             redirect_to action: 'assign_categories', page_id: @page.id,
-                        collection_id: @collection, next_page_id: next_page_id
+                        collection_id: @collection, next_page_id: next_page_id, old_article_ids: old_article_ids
           end
         else
           log_transcript_error(message)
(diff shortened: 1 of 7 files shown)


saracarl:
Ok, if there is a bracketed subject in the page, and the page is *saved*, then we need to save the state/fact/existence of that subject, even if it is wrong.  The scenario here is "I messed up, I realized it during the subject categorization, so I'm going to cancel and go back and fix it."  Subject categorization as a step takes place *after* the page is saved, so we'll have orphaned subjects if they cancel and go back and fix it.  

So here's what we think we should do:
1)  save the page, with the links, which creates the new subject.
2) left folks cancel or continue in the categorization page
3)  Then to fix "orphaned" subjects -- subjects with no links to them, which would occur after they fixed the typo in the subject on the page, we should clean up all orphaned subjects after each page save.  

So for your scenarios:
1a) would actually create the subject on the save, no category.  The subject *likely* gets cleaned up by edits and the next save, but that's not guaranteed.
2a)  yes.
2b)  Yes.  How would a user get into this state?  Saving the subject is correct, but it implies a "hanging" subject assignment -- neither finished nor cancelled.  We should treat that the same as a cancellation, which might make more sense and might be easier to implement.
3a)  Yes.

One caveat:  Project owners can upload canonical subjects via a spreadsheet, and we don't want to delete those if they are not used.  

If the subject record was created via a spreadsheet upload, then the Article record will have a provenance of the filename they uploaded.  Every subject that was created via the transcription process will have a nil provenance.  So when we "clean up orphaned subjects" at page save, we need to *only* delete articles with a nil provenance and no links.
(In production 6000/200,000 have a provenance.)

Removing `next` makes sense to me.

benwbrum:
I've just tested this manually, and it looks like--between this PR and #4175--we've missed an important piece of functionality; abandoning _subject creation_ without losing transcription.

Here's what I did:
* Transcribe a page
* Add a hyperlink around `[[Oats]]` 
* Hit save
* During category assignment, I realized that `Oats` should not be capitalized, so I hit Cancel
* I was redirected back to the transcription page
* However, if I look at the list of Uncategorized subjects, there is a subject for `Oats` with 0 pages.

benwbrum:
This works as designed, and is easy to use.

One question is whether--while we do this work--we want to eliminate orphaned subjects to handle this use case:
* User transcribes a page
* User indexes a mis-spelled word `[[Consitution]]`
* User realizes their mistake on the assign categories screen and presses Cancel
* User corrects the spelling of the word: `[[Constitution]]`
* User categorizes their new subject
* Now two subjects exist:
  * `Constitution` is categorized and links to the page they transcribed
  * `Consitution` is uncategorized and links to no pages at all

saracarl:
Because we create the subject *before* asking the transcriber to categorize it, I don't know of a straightforward way to keep track of this and delete the subject from an abandoned subject categorization.  There's no guarantee the subject needs to be deleted -- maybe something else made them cancel -- so either we:
1)  save the subject and it's in the system as an uncategorized subject with 0 references.  The owner can delete it, correct?
2)  wait to create the subject until after it's categorized.

I'm actually fine with the first solution.

benwbrum:
The alternative approach that would accomplish this is to simply delete _all_ subjects with no `page_article_links` or `article_article_links` or `provenance` at the beginning of a save action, somewhere around the same code where we create page article links and new article records.

saracarl:
That sounds great!  Let's do that.  

Is there a case where folks might create subjects by hand, rather than emergently?

benwbrum:
The only ways for subjects to be created is 
* Emergently (via linking in page transcriptions, page translations, or subject articles)
* Subject upload (which creates a `provenance` attribute)

PR #4218, opened 2024-07-10
2698 - Trim leading spaces on searches

### LINKED ISSUE #2698: Trim leading spaces on searches (work title in particular)
```I'm having an issue searching.  I've been deleting the duplicate entries I have.  For example, I deleted George Lanier (Petersburg) who had 3 pages, but he also had 4 pages which is correct.  When I use the main search box it tells me it can't find Lanier, but if I go to Petersburg document set he is there?  
```

Sara's investigation:

> This is weird.  
> 
> When I search for Lanier I don't see it either.  If I search for his full name: 
> Lanier, George.
> I do find him.
> 
> I thought maybe it was the "," after "Lanier" that was causing the search problem, but I did some searching for "Williams" and was able to find documents with a "Williams," as the last name.  So that's not it.
> 
> Then I tried "Lanier" again, and I did find Lanier, George in the whole collection.  So why didn't it work the first time I tried?  Here's what I noticed -- I had a space (" ") before Lanier from my copy and paste on the first search.  On my second, I typed it in.  Obviously we should trim leading spaces from search terms, but we don't seem to be.  That is my best guess for what happened with your search.  
> 
> Would you try searching for Lanier again, and type it in to make sure you don't get any extraneous spaces?
> 
> I'll open an issue to trim the leading spaces from search terms.
>

WillNigel23:
Was able to replicate the bug as detailed by the ticket. It was indeed the leading spaces on searches.

### DIFF
--- a/app/controllers/search_attempt_controller.rb
+++ b/app/controllers/search_attempt_controller.rb
@@ -1,64 +1,66 @@
 class SearchAttemptController < ApplicationController
-    def create
-        owner = current_user.nil? ? false : current_user.owner
-        query = params[:search]
-        # Some of these objects may be nil, based on the search type
-        work = Work.find(params[:work_id]) if params[:work_id].present?
-        collection = Collection.find(params[:collection_id]) if params[:collection_id].present?
-        document_set = DocumentSet.find(params[:document_set_id]) if params[:document_set_id].present?
-        
-        if params[:work_id].present?
-            search_type = "work"
-        elsif (params[:collection_id].present? || params[:document_set_id].present?) && params[:search_by_title].present?
-            search_type = "collection-title"
-            query = params[:search_by_title]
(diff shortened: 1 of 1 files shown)


app/controllers/search_attempt_controller.rb
```diff
@@ -1,64 +1,66 @@
...
     end
 
-    def show
-        @search_attempt = SearchAttempt.find_by(slug: params[:id])
-        
-        unless @search_attempt.nil?
-            if session[:search_attempt_id] != @search_attempt.id
-                session[:search_attempt_id] = @search_attempt.id
-            end
+    query = query&.strip
```
WillNigel23, line 21: This is mainly the fix. We could opt to trip ONLY leading spaces, but I would assume trailing spaces would present just the same problems so we trip both.

PR #4219, opened 2024-07-11
2222 add undo/redo functionlity

### LINKED ISSUE #2222: Advanced Editor Testing Feedback
Feedback from beta testing:
 

- [x] the microphone button for the dictation feature sits on top of the sup button (AAP)
- [ ] any chance an "undo" button can be added to delete accidental code inserts? (AAP)
- [x]  I'm wondering about the placement of the cursor when clicking on a tag. I usually like the cursor to appear between the opening and closing tag. After the closing tag would also be OK (LW)
- [x] Adding `<del>` would be super helpful -- I think more helpful than `<add>` but having them both would be great.
- [x] When adding the <unclear> tags, the cursor appears one character shifted to the right from where it would be most useful  (JH)
- [x] In terms of tags, the only one that I'm using extensively at this point is italic <i></i>.  If you added that to the menu, I'd definitely give it a try! (MR)
- [ ] As for bugs, the only thing I've noticed is that on my Microsoft platform I sometimes erase a few words/line by mistake and right-clicking the mouse offers me an "undo" function that brings the deletion back.  When it happened today, I still lost the words and still got the "undo" option on the screen, but when I clicked it nothing happened.  (MR)

atbah:
Closes #2222


benwbrum:
@anabeabas Can you take a look at the undo/redo buttons in this screenshot and give us your opinion on their location/appearance?  I'm not sure they should be on the same line as the Save/Preview buttons -- especially in top-to-bottom layout (in the second picture).

Screenshot from 2024-07-16 16-12-59

Screenshot from 2024-07-16 17-15-56

anabeabas:
@benwbrum I would actually put it next to the speech-to-text icon. I noticed that on most websites the undo and redo buttons tend to be on the leftmost side of the screen, and my eyes tend to look for those types of buttons on the left side of the screen too.

benwbrum:
Atef, can you try moving these icons next to the voice dictation icon, and also add tooltips to them?

saracarl:
Ok...  so these are actually "editor buttons", when you think about it.   And when you have other editor buttons in place, it's obvious where they should go:

image

Can we make these match the style of the add / date / del in the example, and put them to the left of any other buttons that are displayed in that spot right above the text field?  These should always show up, unlike the table or any of the advanced mark-up buttons.

saracarl:
Also, I'm not seeing tooltips.  

Based on this, I think we should move the voice dictation button to the same spot.  Outside of scope for this PR, unless it's an easy change to make.

saracarl (changes requested):
UI Decisions above.

PR #4222, opened 2024-07-11
4221 Fix accidental approve to needs-review pages

### LINKED ISSUE #4221: Clicking "Save" approves a page in needs-review
Nat Wooding writes:
> It looks like the buttons have changed again. I'm reviewing the St. John's vestry books and this morning, I've done a couple pages where I had to make minor corrections. When I open the page, Preview, Save, and Approve appear If I click "Save", Approve disappears. So, I'm not certain whether these pages are being flagged as being approved,

I have reproduced this; apparently pressing "Save" as you are reviewing a page accidentally approves the page rather than simply saving it in its current state.

benwbrum:
Closes #4221

### DIFF
--- a/app/controllers/transcribe_controller.rb
+++ b/app/controllers/transcribe_controller.rb
@@ -101,7 +101,7 @@ class TranscribeController  < ApplicationController
         end
         return
       end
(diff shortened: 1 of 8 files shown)


saracarl:
Worked as desired/designed in manual testing.

PR #4231, opened 2024-07-15
4179 - Github workflows

WillNigel23:


WillNigel23:
@benwbrum for codecoverage I suggest codecov, although we are free to choose whatever you prefer.

Also on the test for pry, that should be covered by rubocop tests. However that would come after upgrade so I can temporarily mimic the test for pry here

WillNigel23:
Need to set these variables in .env file

```
MYSQL_ROOT_PASSWORD=...
MYSQL_DEV_PASSWORD=...
MYSQL_SOCKET=...
```

benwbrum (approved):
This all looks fantastic!

Issue #4182, opened 2024-06-27
"recent activity" feed has not been updated for 10 days, even though activity is happening


saracarl:
This collection's: https://fromthepage.com/riksarkivet/i-demokratins-namn
"recent activity" feed has not been updated for 10 days, even though activity is happening.  The last update is one of these June 17th ones:
https://fromthepage.com/riksarkivet/i-demokratins-namn/30002886a/versions/32791751
(probably the 2nd Gunillaramstedt one)

Perhaps introduced by the following deployment:
![image](https://github.com/benwbrum/fromthepage/assets/708566/a3b253f3-fafc-4a09-9a7a-cd4f7e40dc2b)
which seems to be this PR:
https://github.com/benwbrum/fromthepage/pull/4160
I don't see anything there that would cause this.

Could this be related to the done button issues?

saracarl:
recent activity is fine on FindAProject and the one other collection I checked.

benwbrum:
We heard the same problem from another customer today.

PR #4234, opened 2024-07-17
Bypass validations when updating most recent deed date for #4182

benwbrum:
The activity stream on `collection#show` is cached on `@collection.most_recent_deed_created_at`, however this value was not being updated for several collections.  

On further review, this was caused by the addition of validations to check for valid XHTML in `collection.intro_block`.  Since [the after_save deed callback](https://github.com/benwbrum/fromthepage/blob/development/app/models/deed.rb#L90-L100) was calling `update`, validations were being checked.  If a collection failed validation (as was the case for these collections), we were not updating the column on the record.

Since there is no way a volunteer can address validation failures on `collection.intro_blcok` when they are merely transcribing a page, I've changed this logic to `update_columns`, which by-passes validations.

### DIFF
--- a/app/models/deed.rb
+++ b/app/models/deed.rb
@@ -89,13 +89,13 @@ class Deed < ApplicationRecord
 
   def update_collections_most_recent_deed
     if self.collection
-      self.collection.update(most_recent_deed_created_at: self.created_at)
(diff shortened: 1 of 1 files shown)


WillNigel23 (approved):
LGTM

PR #4235, opened 2024-07-17
1205 - Refactor model status constants

WillNigel23:


spec/models/page_spec.rb
```diff
@@ -11,35 +11,35 @@
...
   describe '#validate_blank_page' do
     let(:page) { build_stubbed(:page) }
-    it 'sets Page::STATUS_BLANK' do
-        page.status = Page::STATUS_BLANK
+    it 'sets :blank' do
+        page.status = :blank
         page.validate_blank_page
 
-        expect(page.status).to eq(Page::STATUS_BLANK)
+        expect(page.status).to eq('blank')
```
benwbrum, line 18: Here's another case in which I wonder if we can get access to the enums outside of the model, so that we won't have to hard-code string literals.

spec/models/work_statistic_spec.rb
```diff
@@ -316,15 +315,7 @@
...
-        statuses = %w[
-          transcribed
-          corrected
-          translated
-          blank
-          indexed
-          review
-        ]
-        statuses.each do |status|
+        Page.statuses.each_value do |status|
```
benwbrum, line 318: Does `statuses` produce a hash instead of an array/list?  (Wondering why it's necessary to call `each_value` instead of `each` here.)

spec/models/work_statistic_spec.rb
```diff
@@ -374,10 +365,10 @@
             ws.recalculate
             expect(ws.complete).to eq(0)
 
-            create(:page, work: work, status: Page::STATUS_TRANSCRIBED)
-            create(:page, work: work, status: Page::STATUS_BLANK)
-            create(:page, work: work, status: Page::STATUS_INDEXED)
-            create(:page, work: work, status: Page::STATUS_NEEDS_REVIEW)
+            create(:page, work: work, status: :transcribed)
```
benwbrum, line 368: Okay, so here's a place where we have access to `:transcribed` from a spec file, so don't have to use a string literal.  Is it because we're using it within `status:`?  _I'm so confused..._

benwbrum (changes requested):
I think my only concrete request is to deal wiith `nil` for _not started_, but I had a lot of questions about how enums work.

app/models/page.rb
```diff
@@ -83,19 +84,31 @@ class Page < ApplicationRecord
...
 
   serialize :metadata, Hash
 
-  scope :review, -> { where(status: 'review') }
-  scope :incomplete, -> { where(status: 'incomplete') }
-  scope :translation_review, -> { where(translation_status: 'review') }
-  scope :needs_transcription, -> { where(status: [nil]) }
-  scope :needs_completion, -> { where(status: [STATUS_INCOMPLETE]) }
+  STATUS_VALUES = {
+    blank: 'blank',
```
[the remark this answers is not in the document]
WillNigel23: We cannot add nil values in enum.

What we can do instead is, make a default value, let's say 'not_started' as you mentioned. And we can map it in the enum.

app/views/shared/_description_tabs.html.slim
```diff
@@ -39,9 +39,9 @@
           span.page-nav_info = t('.metadata')
           -if index.present?
             -next_works = @collection.works.to_a[index+1..]
-            -next_work = next_works.detect {|candidate| candidate.pages.where(status: [nil, Page::STATUS_INCOMPLETE]).first || candidate.description_status == Work::DescriptionStatus::UNDESCRIBED }
+            -next_work = next_works.detect {|candidate| candidate.pages.where(status: [nil, :incomplete]).first || candidate.description_status == Work::DescriptionStatus::UNDESCRIBED }
```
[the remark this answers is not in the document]
WillNigel23: Only because we are using `status:`

app/models/work_statistic.rb
```diff
@@ -125,16 +125,16 @@ def recalculate_from_hash(stats=nil)
...
-    self[:corrected_pages]    = stats[:transcription][Page::STATUS_TRANSCRIBED] || 0
-    self[:blank_pages]        = stats[:transcription][Page::STATUS_BLANK] || 0
-    self[:annotated_pages]    = stats[:transcription][Page::STATUS_INDEXED] || 0
-    self[:needs_review]       = stats[:transcription][Page::STATUS_NEEDS_REVIEW] || 0
-
-    self[:translated_pages]     = stats[:translation][Page::STATUS_TRANSLATED] || 0
-    self[:translated_blank]     = stats[:translation][Page::STATUS_BLANK] || 0
-    self[:translated_annotated] = stats[:translation][Page::STATUS_INDEXED] || 0
-    self[:translated_review]    = stats[:translation][Page::STATUS_NEEDS_REVIEW] || 0
+    self[:transcribed_pages]  = stats[:transcription]['transcribed'] || 0
```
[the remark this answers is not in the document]
WillNigel23: I'll modify this. But essentially, the reason I did this is the function used to group `stats` is grouping it by key_name in string literals. So that instead of `review` it uses `needs_review` string literal for the hash map. Let me think more on this

spec/models/page_spec.rb
WillNigel23: Yes, we can do `Page.statuses[:blank]`. Might have missed this as I had to refactor a lot

spec/models/work_statistic_spec.rb
WillNigel23: Yes, it produces a hash. We can do `Page.statuses.each do |key, value|` but it's essentially the same with each_value

spec/models/work_statistic_spec.rb
WillNigel23: Let me check this again. We should be able to just use `create(:page, work: work, status: :whatever_status)`, but I think I did use string literals for the other cases because of FactoryBot. But I will review and apply changes

app/models/page.rb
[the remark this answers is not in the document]
WillNigel23: @benwbrum I think `empty` describes it better instead of `not_started`, but it's your call.

WillNigel23:
Further notes for when to use `string` vs `hash`

When we want to do comparison, and we access the value via `page.status`. Doing `page.status` will give us the value from db, so we compare it via string comparison. (But in any case, we can just use the `page.status_blank?` and other magic functions generated by rails.

Use `hash` when used with ActiveRecord methods like `where` functions. It is handled by ORM that way and enum conversion is taken care of.

app/models/page.rb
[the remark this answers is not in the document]
benwbrum: How about `new` as a status in this case?

PR #4237, opened 2024-07-18
3624 - Upgrade rails

### LINKED ISSUE #3624: Upgrade Ruby and Rails to current versions

WillNigel23:
Retained all specs for (with some minor changes on deprecated syntax)


app/controllers/user_controller.rb
```diff
@@ -157,9 +153,27 @@ def authorized?
...
+      :real_name,
+      :orcid,
+      :slug,
+      :website,
+      :location,
+      :about,
+      :preferred_locale,
+      :help,
+      :footer_block,
+      notification_attributes: [
```
WillNigel23, line 168: We can pass in `accepts_nested_attributes` for notification at `user` model so we can update them together.

Further optimization needed in update_profile logic.

app/helpers/export_helper.rb
```diff
@@ -223,8 +223,8 @@ def write_work_exports(works, out, export_user, bulk_export)
 
 
   def work_to_xhtml(work)
-    @work = Work.includes(pages: [{notes: :user}, {page_versions: :user}]).find_by(id: work.id)
-    render_to_string :layout => false, :template => "export/show.html.erb"
+    @work = Work.includes(pages: [{ notes: :user }, { page_versions: :user }]).find_by(id: work.id)
+    render_to_string template: 'export/show', layout: false, formats: [:html], handlers: [:erb]
```
WillNigel23, line 227: This syntax seems to be not supported in newer versions of rails (I got multiple failures due to doing relative path with extention, ex: `deed/deed.html`, `export/show.html.erb`).


We instead pass in `formats` and `handlers` params

app/models/ia_work.rb
```diff
@@ -165,7 +165,7 @@ def ingest_work(id)
 
     self.save!
     # now fetch the scandata.xml file and parse it
-    scandata_url = "http://#{server}#{dir}/#{URI.encode(scandata_file)}" # will not work on new format: we cannot assume filenames are re-named with their content
+    scandata_url = "http://#{server}#{dir}/#{CGI.escape(scandata_file)}" # will not work on new format: we cannot assume filenames are re-named with their content
```
WillNigel23, line 168: URI.encode is deprecated, use CGI.escape instead. 

(This is also why we need to use forked version of gravatar gem, old one is not updated and is still using URI.encode)

app/views/user/update_profile.html.slim
```diff
@@ -50,39 +50,40 @@
...
           |&nbsp;
           =select_tag :dialect, {}, {class: 'w40'}
 
-    p.email_notifications =t('.select_notification')
-    table.form.email_notifications
-      tr
-        td =f.check_box("notifications[user_activity]", {checked: @user.notification.user_activity})
-        th =f.label :user_activity, t('.user_activity')
-      -if @user.owner
+    =f.fields_for :notification do |nf|
```
WillNigel23, line 65: We can do fields_for, this is so we don't have to manually add the names. This will fall under `notification_attributes` and we can pass in `@user.update(user_params)` and the notification will also be updated

config/i18n-tasks.yml
```diff
@@ -106,16 +106,34 @@ ignore_missing:
     - 'deed.deed.*'
     - 'article.graph.title'
     - 'thredded.*'
-  es: &NOT_EN
```
WillNigel23, line 109: aliases in yml files are now considered unsafe and not supported anymore.

config/locales/deed/deed-en.yml
```diff
@@ -6,29 +6,28 @@ en:
     collection_inactive: Collection Inactive
     collection_joined: Collection Joined
     deed:
-      html:
```
WillNigel23, line 9: No need to scope in .html anymore since we moved it to formats: [:html], it still looks at translations .deed instead of at .deed.html

config/rubocop/.rubocop-layout.yml
```diff
@@ -0,0 +1,47 @@
+Layout/EmptyLinesAroundModuleBody:
```
WillNigel23, line 1: I find these configs a good default startinig point for rubocop. But these are preferential so we can tweak more moving forward.

app/models/page.rb
```diff
@@ -39,54 +39,53 @@
...
+    work.work_statistic.update_last_edit_date if saved_change_to_source_text? || saved_change_to_source_translation?
   end
 
-  after_initialize :defaults
-  after_destroy :update_work_stats
-  #after_destroy :delete_deeds
-  after_destroy :update_featured_page, if: Proc.new {|page| page.work.featured_page == page.id}
-
-  serialize :metadata, Hash
+  serialize :metadata, type: Hash
```
WillNigel23, line 88: New syntax, old is causing error

spec/models/xml_source_processor_spec.rb
```diff
@@ -1,76 +1,78 @@
...
-  let(:old_link)  { build_stubbed(:article, title: 'Old Subject', collection: collection ) }
-    
     context 'subject linking not disabled (default)' do
       it 'builds the xml document' do
         expect(work.collection).to eq(collection)
         xml = page.wiki_to_xml(page, Page::TEXT_TYPE::TRANSCRIPTION)
         expect(Article.all.count).to eq(3)
         expect(PageArticleLink.all.count).to eq(4)
-        expect(xml).to eq(EXPECTED_XML)
+        expect(Nokogiri::XML(xml).to_s).to eq(Nokogiri::XML(EXPECTED_XML).to_s)
```
WillNigel23, line 43: For some reason eq() matching is failing due to inconsistent tabs/spaces, but contents are literally the same. Doing this ensures we compare the contents.

config/i18n-tasks.yml
WillNigel23: Reason: aliases in yml files are unsafe and rails moved away from supporting this

saracarl:
We're closing this in favor of #4429.

PR #4239, opened 2024-07-19
Use destroy instead of delete for notes #4238

### LINKED ISSUE #4238: 500 error deleting older notes
Payton at GRI reports that they are unable to delete older notes.  When I log in as her and try to delete the top notes on [this page](https://fromthepage.com/gri/stendahl-letters-qc/l000324/display/1678501), I see the server return a 500 error.

Looking in the logs, I see this:
```
I, [2024-07-19T01:26:44.694705 #1503028]  INFO -- : Started DELETE "/notes/32056132" for 50.223.147.195 at 2024-07-19 01:26:44 +0000
I, [2024-07-19T01:26:44.697143 #1503028]  INFO -- : Processing by NotesController#destroy as JS
I, [2024-07-19T01:26:44.697506 #1503028]  INFO -- :   Parameters: {"id"=>"32056132"}
I, [2024-07-19T01:26:44.707396 #1503028]  INFO -- : Completed 500  in 10ms (ActiveRecord: 1.7ms | Allocations: 4318)
F, [2024-07-19T01:26:44.710936 #1503028] FATAL -- :   
NoMethodError (undefined method `delete' for nil:NilClass):
  
app/controllers/notes_controller.rb:47:in `destroy'
app/controllers/application_controller.rb:64:in `switch_locale'
```

The [relevant line of code](https://github.com/benwbrum/fromthepage/blob/development/app/controllers/notes_controller.rb#L47) calls `@note.deed.delete`.  Is it possible that some notes don't have deeds?

benwbrum:
Closes #4238

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -31,7 +31,10 @@ class Note < ApplicationRecord
   belongs_to :page, optional: true
   belongs_to :work, optional: true
   belongs_to :collection, optional: true
-  has_one :deed, :dependent => :destroy
+
+  has_many :deeds, dependent: :destroy
+  has_one :deed, -> { order(created_at: :desc) }
+
   has_many :flags
 
   after_save :email_users
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -44,8 +44,7 @@ class NotesController < ApplicationController
 
   def destroy
     @note = Note.find(params[:id])
-    @note.deed.delete
-    @note.delete
+    @note.destroy
     respond_to do |format|
       format.json { head :no_content }
       format.html { redirect_back fallback_location: root_path, notice: t('.note_has_been_deleted') }


WillNigel23 (approved):

app/models/note.rb
```diff
@@ -31,7 +31,10 @@ class Note < ApplicationRecord
   belongs_to :page, optional: true
   belongs_to :work, optional: true
   belongs_to :collection, optional: true
-  has_one :deed, :dependent => :destroy
+
+  has_many :deeds, dependent: :destroy
+  has_one :deed, -> { order(created_at: :desc) }
```
WillNigel23, line 36: Played with this a bit more. It seems that doing `has_one :deed` is only fetching the first deed linked to the current note (deed.note_id == note.id)

If for some reason we created multiple deeds for current note, it will only ever get the first deed created.

To avoid orphaning the other deeds linked, we added `has_many :deeds, dependent: :destroy` so that all deeds linked to a note is deleted. We still kept the `has_one :deed` relationship so functionality is retained (except that I scoped it to get the LATEST deed. Let me know if we should always get the FIRST deed linked)

Tested and it successfully notes without orphaning deeds.
@benwbrum

benwbrum: Isn't this a natural feature of `has_one` associations?  Unless there's a `UNIQUE` constraint on the foreign key column in the database, there is no way to actually enforce a single child record per parent.

WillNigel23: It's just that we have lines like these
  def record_deed
    deed = Deed.new
    deed.note = @note # here
    deed.page = @page
    deed.work = @work
    deed.collection = @work.collection
    deed.deed_type = DeedType::NOTE_ADDED
    deed.user = current_user

    deed.save!
    update_search_attempt_contributions
  end
Can we assure that we are not creating multiple deed for a single note? That would then fall to the described behaviour I mentioned in comment @benwbrum 

As per enforcing single child record, we are still doing what we initially have before, just that I changed it so it checks in the last element instead of the first (which is not going to affect anyway if we indeed cannot make 2 deeds have the same parent note_id).

So really the only change is that, we allow all related deeds to be deleted on note destroy.

PR #4240, opened 2024-07-22
Markdown table in code mirror

### LINKED ISSUE #2753: CodeMirror button for inserting markdown tables
This might be a model for inserting spreadsheets or tables into textual documents that aren't consistent (using markdown):

https://www.tablesgenerator.com/markdown_tables

atbah:
Closes #2753


saracarl:
So this works well, even with other markup buttons turned on:
image
However, this is an "advanced" sort of thing that would easily confuse transcribers, so I think we want to make it part of the editor buttons a project owner has to turn on.  (For example here:  http://localhost:3000/umdtranscription/david-c-driskell-papers/edit/tasks#edit-buttons)

Let's call it "Table" with the text "Inserts a markdown table of transcriber specified rows and columns"
Put it after superscript in the list.
Instead of trying to describe it after the checkbox, let's just say "Documentation" with a link to https://content.fromthepage.com/project-owner-documentation/table-encoding/

PR #4244, opened 2024-07-24
4236 fix missing subject popup

### LINKED ISSUE #4236: Missing pop-up for subjects in read view
A few years ago, users were able to move their mouse over a subject link when reading a transcript and see a pop-up showing a preview of the subject article, as seen in the screenshot below.  Several months ago, this stopped working.

Screenshot from 2021-06-21 10-57-25

atbah:
Closes #4236

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -123,8 +123,7 @@ $.fn.tooltip = function(s) {
     var url = $element.data('tooltip');
     var $tooltip = $('<div>').addClass('tooltip');
 
-    $element.on('click.Tooltip', function(e) {
-      e.preventDefault();
+    function showTooltip() {
       var offset = $element.offset();
       var pos_top = offset.top + $element.height();
       var pos_left = offset.left;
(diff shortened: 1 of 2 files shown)


benwbrum:
This fixes the issue as described.  In testing, I noticed a UI inconsistency -- the pretty pop-up for subjects appears if you click the link, but no if you mouse-over.  Can we add the popup to the mouseover as well?

benwbrum:
Now we're getting the alt-text title _and_ the pop-up:
Screenshot from 2024-08-23 13-44-51

saracarl (changes requested):
So the first time a subject is hovered over on a page, the pop-up is an empty black box with a delay before the subject details are filled in.  In subsequent hovers, there is not delay/empty black box.  I don't think it used to behave that way.

saracarl:
Discussed with @benwbrum  and this is how it's always worked -- possibly my dev box is just slower than production.  We're merging this in.

saracarl (approved):

PR #4245, opened 2024-07-24
3161 - Owner login with no collection should go to 'Start a project'

WillNigel23:


saracarl (approved):
This looks great.  

Of course, testing it I discovered we aren't doing the right thing in the trial sign-up case, either, but I'll open a new issue with that.

PR #4246, opened 2024-07-25
3352 page navigator in fullscreen mode

### LINKED ISSUE #3352: Page navigator not available in full-screen mode
Transcriber suggests that the page navigator should be available in full screen mode as well, relocating it between the 'Layout' and 'Mark as blank' like this:

unnamed (1)

atbah:
Closes #3352


WillNigel23:
Specs are failing because, unlike previously where in a page we only have 1 html element that is a.page-nav_prev, your changes introduced multiple buttons like this. So capybara does not know which button in particular does this try to click

```
page.find('a.page-nav_prev').click
```

Need to adjust those affected lines to find the correct button.

WillNigel23 (changes requested):
Specs are failing because, unlike previously where in a page we only have 1 html element that is a.page-nav_prev, your changes introduced multiple buttons like this. So capybara does not know which button in particular does this try to click

```
page.find('a.page-nav_prev').click
```

Need to adjust those affected lines to find the correct button.

benwbrum:
I'm a little confused why we can't use the new page navigation partial within the page_tabs partial as well, since it seems like we're duplicating code here.

atbah:
@benwbrum It is hard to use the navigation partial within the page_tabs in full-screen mode. We need to add a new pagination navigator between the layout and the right buttons. It is obviously different section. When applying full-screen mode, the toolbar and editor views are floated to the front, causing the page_tabs to be obscured behind them. Please have a look at the screenshot. 
image

atbah:
Please let me know if you want to keep just one pagination component. If so, I can try to fix the issue by adjusting the CSS. Alternatively, I can update the spec to ensure it interacts only with the first pagination element.

benwbrum:
Okay, I think I understand the problem.  We can keep this as structured; either we can change the CSS class in the view to keep the specs from failing, or we can modify the spec tests to handle duplicate links with the same class.

WillNigel23 (approved):

atbah:
@benwbrum  I think you should check "Always show in fullscreen" checkbox to keep fullscreen mode when navigating pages. 
image

PR #4254, opened 2024-07-29
catch exceptions sending emails in owner newsletter rake task

saracarl:


benwbrum (approved):
Looks good.  Let's merge once the specs finish.

PR #4256, opened 2024-07-30
4041 redo adding collaborators to restricted works

WillNigel23:


saracarl (changes requested):
image


So either we need to list the collaborators under "Allowed Collaborators" or, since this is rarely used, I'm ok with just removing "Allowed Collaborators" and having the "Edit Collaborators" button.

benwbrum:
I just tried this out and would really like to see the collaborators listed on the screen.  When I tested, I found myself annoyed to have to press the "Edit Collaborators" button to see who was authorized, because I was used to the collection listing screens.

Screenshot from 2024-08-07 10-40-04
Screenshot from 2024-08-07 10-39-55
Screenshot from 2024-08-07 10-39-32

PR #4262, opened 2024-08-06
3937 - Links clicked in read views replace current tab

WillNigel23:


app/helpers/abstract_xml_helper.rb
```diff
@@ -286,7 +311,11 @@ def xml_to_html(xml_text, preserve_lb=true, flatten_links=false, collection=nil,
     my_display_html.gsub!('<p/>','')
     my_display_html.gsub!(/<\/?page>/,'')
 
-    return ActionController::Base.helpers.sanitize(my_display_html.strip, :tags => SANITIZE_ALLOWED_TAGS).gsub('<br>','<br/>').gsub('<hr>','<hr/>')
+    ActionController::Base.helpers.sanitize(
+      my_display_html.strip,
+      tags: SANITIZE_ALLOWED_TAGS,
+      attributes: SANITIZE_ALLOWED_ATTRIBUTES
+    ).gsub('<br>','<br/>').gsub('<hr>','<hr/>')
```
WillNigel23, line 318: It seems the sanitize function is stripping 'target' attributes on hrefs.
We can deal with this by whitelisting href and target, but passing in `attributes` is restrictive rather than additive so I declared SANITIZE_ALLOWED_ATTRIBUTES above.

I may have missed some attributes there but I prioritized obvious ones and those that are explicitly used here in xml_to_html function.

NOTE:
Users will still have to add target="_blank" or "new" as done here:
https://www.fromthepage.com/nbwm/logbooks-and-journals/kwm-482/display/32946693

Let me know if this is okay

spec/helpers/abstract_xml_helper_spec.rb
```diff
@@ -24,6 +26,11 @@
     expect(xml_to_html(@xml_text, false, true)).to include("Faith<span class=\"line-break\"></span>fully")
   end
 
+  it 'returns a <a> tag with preserved href and target attributes' do
+    expect(xml_to_html(@xml_text, true, true)).to include(@a_tag_with_attr)
```
WillNigel23, line 30: added specs to test new functionality

benwbrum:
This is tested and can be merged as soon as it passes spec tests.

PR #4273, opened 2024-08-14
2979 - Metadata template and download should be consistent

### LINKED ISSUE #2979: Metadata template & download should be consistent
Could we make the metadata template and the metadata export the same spreadsheet?  Or at least have the same fields in the same order as much as possible?

see also #2715

WillNigel23:


benwbrum (changes requested):
On testing, @saracarl and I noticed these things that need to be changed:
* We should move the metadata upload processing to a background task, since it takes a very very long time for large collections.
* Some fields which are internal to FromThePage (work slug, document set) are now appearing in the About tab for a work if we update them.  These should not be processed the same way that non-metadata fields are, and should not appear in About.

benwbrum:
Before upload:
Screenshot from 2024-11-06 07-39-55
After upload:
Screenshot from 2024-11-06 07-39-25

benwbrum:
I did a quick UI test and found that clicking on the "Export Metadata as CSV" button in the Export All Works tab gave me an exception.
Screenshot from 2024-11-13 09-52-55

PR #4274, opened 2024-08-15
4263 - Bootstrap css classes not working

WillNigel23:


app/views/shared/_landingheader.html.erb
```diff
@@ -13,9 +13,11 @@
     <link rel="preconnect" href="https://fonts.googleapis.com">
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
 
-    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.1/dist/css/bootstrap.min.css" rel="stylesheet"
```
benwbrum, line 16: It looks like these three lines changed from a hard-coded HTML reference to externally-hosted assets to a rails helper-based reference to assets which are local.  Was this what we intended?

PR #4283, opened 2024-08-26
4103 - Improve document set settings tab to match collection settings

WillNigel23:


spec/requests/document_sets_controller_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    it 'renders status and template' do
+      login_as owner
+      subject
+
+      follow_redirect!
+      expect(response).to have_http_status(:ok)
+      expect(response).to render_template(:index)
+    end
+  end
+end
```
WillNigel23, line 151: Ideally, we add all defined routes/functions inside document_sets_controller here. That way we have a clear test on all requests.

spec/interactors/document_set/update_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it 'updates document_set' do
+      expect(result.success?).to be_falsey
+      expect(result.errors).to include(
+        "Title can't be blank",
+        'Title is too short (minimum is 3 characters)'
+      )
+    end
+  end
+end
```
WillNigel23, line 48: Moving business logics to interactors also means we can isolate each specific functionalities into unit tests like this.

saracarl (approved):
Functionality is good.

benwbrum (changes requested):
There are enough problems with the translations that I think we should discuss the overall approach.  It looks like we are re-translating from English into DE/ES/FR/PT.  This runs the risk of replacing wording that has been reviewed by native speakers, and which we hope to match, with new machine wording which is more awkward to read, less appropriate for our problem domain, and inconsistent with other wording in the system.

I only flagged a few of the problem messages in my review comments, since it looks like we need to change overall approaches.

WillNigel23:
Changes:
1. Translation suggestions applied
2. Added spec (but only warns instead of fail) of commonly mis-translated words for 'work'

@benwbrum @saracarl

config/locales/document_sets/document_sets-de.yml
```diff
@@ -2,107 +2,116 @@
...
-      document_created: Dokumentenset wurde erstellt
+      document_created: Dokumentensatz wurde erstellt
     edit:
-      current_url_for_this_work: Die aktuelle URL für dieses Werk ist %{current_url}. Wenn Sie den Abschnitt zum Dokumentenset in der URL bearbeiten möchten, verwenden Sie bitte Kleinbuchstaben und Bindestriche zwischen den Wörtern.
+      current_url: 'Die aktuelle URL zum Dokumentensatz:'
       description: Beschreibung
-      description_message: Die Beschreibung wird auf der Projektübersichtsseite und in der Projektinfo angezeigt. Ein abgekürzter Teil der Beschreibung wird auf dem Startbildschirm der Organisation und unter „Projekt suchen“ angezeigt.
-      document_is_marked_public: Wenn das Dokumentenset als öffentlich gekennzeichnet ist, sind darin enthaltene Werke einsehbar, auch wenn die Sammlung auf privat gesetzt ist.
-      save_document_set: Dokumentenset speichern
+      description_message: Die Beschreibung wird auf dem Projektübersichtsbildschirm und dem Projektinfobildschirm angezeigt. Ein gekürzter Teil der Beschreibung wird auf dem Organisations-Startbildschirm und dem Bildschirm „Projekt suchen“ angezeigt.
```
benwbrum, line 9: This seems like a lot of changes -- Projectübersichtsseite (Project overview page) is turned into Projectübersichtbildschirm (Project overview screen).  Also in line 5, `Documentenset` is turned to `Documentensatz`.

config/locales/document_sets/document_sets-es.yml
```diff
@@ -2,52 +2,84 @@
...
     index:
-      actions: Acciones
-      aggregating_works_thematic_exhibits: Agregando obras a exposiciones temáticas
-      assign_works_to_document_sets: Asignar Obras a Conjuntos de Documentos
-      create_a_document_set: Crear un Conjunto de Documentos
-      delete: Eliminar
-      delete_document_set_confirm_message: "¿De verdad deseas eliminar este conjunto de documentos? Esto eliminará tu conjunto de documentos, pero no borrará ninguna de las obras en el conjunto de documentos."
-      document_sets_are_sub_sets: 'Los conjuntos de documentos son subconjuntos de los documentos dentro de una colección. Tienen varios usos, incluyendo:'
-      document_sets_for: Conjuntos de Documentos para %{title}
+      actions: Comportamiento
```
benwbrum, line 51: Similarly, `Acciones` is turned to `Comportamiento`, indicating a retranslation.

benwbrum (changes requested):
It looks like we still have a lot of new translations here.  I just spot-checked the German and Spanish, but I found enough.

Maybe compare the values in `development` with the values here with the same keys to revert the auto-translations?

spec/i18n_spec.rb
```diff
@@ -35,4 +36,25 @@
                     "Run `i18n-tasks check-consistent-interpolations' to show them"
     expect(inconsistent_interpolations).to be_empty, error_message
   end
+
+  describe 'translations' do
+    let(:incorrect_translations) { ['travail', 'travaux', 'trabalho', 'trabalhos', 'arbeit', 'arbeiten', 'trabajo', 'trabajos'] }
```
benwbrum, line 41: Woo hoo!  Hooray!  I love this!

benwbrum: Let's add Dokumentensetz to this list, too.

benwbrum (changes requested):
A couple of minor translation issues, but this is really exciting!

PR #4297, opened 2024-08-29
4275 - Fix pandoc error on \textquotesingle

WillNigel23:


saracarl (approved):
Tested locally with general pdf export and an underlined apostrophe.  Both worked, but that's not the best test because it was never broken locally for me.  Will test again after deployment.

PR #4306, opened 2024-09-02
4302 - Missing page breaks on facing pdf

WillNigel23:
https://github.com/jgm/pandoc/issues/2330

Pandoc seems to not support page break when converting from `latex` to `pdf`.
There are existing lua filters as workaround for this, but nothing made yet for a pdf output.

Ironically, our old approach of directly uploading from `md` to `pdf` supports `\newpage` and other tags, but we go back to initial issue of unsupported font  for \textsinglequote. So this PR will revert to old approach and fix the initial issue at #4297 

After much research on the matter, I have pinpointed the error to Xelatex being not fully compatible with fonts (fontenc for example, which is what is being used for pdfs). To resolve this I switched out the pdf engine from `xelatex` to `pdflatex` and everything worked correctly.

@benwbrum @saracarl is there any particular reason for using xelatex engine? If there is none I think this proposed solution should be acceptable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get -qq update
-          sudo apt-get install -y ghostscript poppler-utils graphviz
+          sudo apt-get install -y ghostscript poppler-utils graphviz pandoc texlive-full
           bundle install --jobs 4 --retry 3
       - name: Create db
         run: bundle exec rake db:create
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,7 @@ end
 group :test do
   gem 'capybara'
   gem 'database_cleaner'
+  gem 'rails-controller-testing'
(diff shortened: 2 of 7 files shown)


WillNigel23:
This is not fully ready yet, I plan on adding specs here @saracarl @benwbrum (unless you deem this critical to release asap).

But I would like to get your opinion before writing specs as it will vary depending on what we decide to go move forward with.

benwbrum:
We used `xelatex` as part of the development of a [different feature -- using LaTeX encoding for formulae within transcriptions](https://content.fromthepage.com/project-owner-documentation/encoding-formula-with-latex/), which we should test out as part of any of this work.

The choice of xelatex here was--I think--just because we already had been using that engine.

WillNigel23:
Rebased, and added more specs @benwbrum 

For the specs, I made sure to add/remove my changes locally and see if it does break (which it does, so I can confirm that this is not a false green rather really testing the pdf/doc export)

PR #4313, opened 2024-09-05
4288 fixed versions tab UI issue

### LINKED ISSUE #4288: Fix versions tab UI
We lose line breaks and formatting (i.e. links, but probably other formatting as well) on the right side of the version comparison screen in the versions tab:

image

Can we make sure the right side gets the same UI processing as the left?

atbah:
Closes #4288


benwbrum:
This fixes the problem with versions for document-based transcription projects perfectly.

Unfortunately, the diff view of spreadsheet projects loses all formatting, possibly due to this change.

In this screenshot, I updated a spreadsheet cell.  The diff view now shows only the difference and loses the context around it.
Screenshot from 2024-09-05 18-14-02

atbah:
@benwbrum I fixed the issue.

PR #4336, opened 2024-09-16
3890 UI for transkribus

saracarl:


spec/models/ai_job_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
WillNigel23, line 1: this should be 'spec_helper'

PR #4339, opened 2024-09-18
update newrelic and rubysaml gems

saracarl:
addresses https://nvd.nist.gov/vuln/detail/CVE-2024-45409


saracarl:
[Test failures](https://github.com/benwbrum/fromthepage/actions/runs/10921817876/job/30314772659?pr=4339) are all in xml_source_processor:
```
rspec ./spec/models/xml_source_processor_spec.rb:39 # XmlSourceProcessor#wiki_to_xml subject linking not disabled (default) builds the xml document
rspec ./spec/models/xml_source_processor_spec.rb:48 # XmlSourceProcessor#wiki_to_xml subject linking disabled builds the xml document
rspec ./spec/models/xml_source_processor_spec.rb:63 # XmlSourceProcessor#valid_xml_from_source builds the xml document
```

Gemfile.lock
```diff
@@ -493,7 +493,7 @@ GEM
     responders (3.1.1)
       actionpack (>= 5.2)
       railties (>= 5.2)
-    rexml (3.2.6)
+    rexml (3.3.7)
```
saracarl, line 496: Could this gem change be causing xml_source_processing errors?

WillNigel23: I have encountered this before, it seems the change in rexml gem removed the extra '\n' here. 

I'm not entirely sure if this is indeed due to some gem updates they are working on at their end, but if we check most of the difference are extra whitespaces and '\n', I don't think this is fatal for our codebase. @saracarl

WillNigel23 (approved):
Chomp ᑈ

PR #4346, opened 2024-09-23
4289 switch text source

### LINKED ISSUE #4289: Versions screen -- switch between text and source
So our versions screen, or at least parts of it, is going to start playing a bigger part in some of the upcoming AI driven features.  A first step of that is to support 2 different modes in the version comparison screen:

1)  The existing versions screen (fixed per #4288) that shows formatted text, with links and bolds/italics/etc.  This should come from the page's `transcription` field.
2)  A "source text" version comparison that instead of formatting the page, shows the tags the transcriber entered for both text formating and wiki links.  So a lines might look like:  
```
China GrovE [[Mississippi--Pike County|Pike Co]]
May 10<sup>th</sup> 1876
```
and show differences in the mark up as well was the actual text.  (i.e. make it easier to see when a link changes, which right now requires hovering and looking at the target URL of the link.)
This come from the `xml_transcription` field of the page.

We'll also want to do this for translation versions.  Those come from  `source_translation` and `xml_translation`. I believe those are just stacked on top of each other in the versions screen -- that's fine to keep the same way, but just toggle together with the transcription.  

Here's what I think the UI could look like (but happy to have other suggestions if you have a better idea), with one or two buttons next to the paired side by side/over under icons:
image
But maybe a paired "source" and "document" icon like we have paired side by side/over under icons?

The source/code icon we like is this one:  https://fonts.google.com/icons?selected=Material+Symbols+Outlined:code:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=html&icon.size=24&icon.color=%235f6368
And for the document icon:  https://fonts.google.com/icons?selected=Material+Symbols+Outlined:article:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=document&icon.size=24&icon.color=%235f6368

atbah:
Closes #4289


app/views/page_version/show.html.slim
```diff
@@ -70,35 +93,64 @@ table.diff-versions(data-fullheight='{ "bottom": 30, "cssrule": "min-height" }')
...
+                    div.source-code
+                      == @selected_version.source_translation
+                    div.html-code
+                      == xml_to_html(@selected_version.xml_translation)
             - else
               h4(data-diff-title="original" data-empty=t('.untitled')) = @previous_version&.title.presence || ""
               div.table-responsive(data-diff-transcription="original" data-empty=t('.no_transcription_provided'))
-                == xml_to_html(@previous_version&.xml_transcription)
+                div.source-code
+                  == @previous_version.transcription
```
WillNigel23, line 127: Seems unhandled nil error here. Please do safe operation @previous_version&.transcription to avoid error when @previous_version is nil

saracarl:
This is beautiful and works great.

PR #4352, opened 2024-09-26
4350 - Ingestor hangs fix

WillNigel23:


WillNigel23:
Working on  adding specs + minor refactors. But in case this is rushed then this fix problem.

benwbrum:
How are we doing on the test coverage and refactoring?  Is this ready to test?

WillNigel23:
Let's move this forward in the meantime @benwbrum , this PR contains general error handling so that ingestor do not crash. I would like for this to be merged asap to have better logs on future errors (which would greatly help refactoring and specs too)

PR #4362, opened 2024-09-30
4223 - Approved pages with no changes need a version

### LINKED ISSUE #4223: approved pages with no changes need a version
If you "approve" a page and there are changes in your approval, then we generate a version of the page in the "versions" tab.

If the page is perfect and you approve it with no changes, then we don't generate a version of the page.  We need to.

This problem is a result of changes made in #3921

WillNigel23:


benwbrum:
I just tested this and ran into a problem:
* Edit a collection with a review-optional workflow
* Edit a page and check the "needs review" box
* Save the page without making any changes.
  * The versions tab correctly shows a new version with a needs-review status
* Edit the same page and uncheck the "Needs review" box
* Save the page without making any changes
  *  _The versions tab does not show a new version with approved status_

WillNigel23:
Upon further investigation, it seems the page version object is actually created. But we are not showing it in the versions tab since `content_changed?` does not consider status changes.

I added to include status changes here in page_versions. Let me know if this fixed the problem

benwbrum:
This may also resolve #4433

PR #4363, opened 2024-09-30
4311 - Missing display name fix

WillNigel23:
The fix works in 2 parts:

We update the callback `update_display_name` to set real_name and display_name to `nil` in case the string value is blank. This will prevent new records with ' ' display_name values from being created.

Second is we update the display_name override to detect [:display_name] presence, otherwise use login. This will help us so that we don't need to do a migration to fix bad data

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -150,8 +150,9 @@ class User < ApplicationRecord
     end
   end
 
-
   def update_display_name
+    self.real_name = nil if self.real_name.blank?
+
     if self.owner
       self.display_name = self.real_name
     else
@@ -287,9 +288,9 @@ class User < ApplicationRecord
 
   def display_name
     if self.guest
-      "Guest"
+      'Guest'
     else
-      self[:display_name] || self[:login]
+      self[:display_name].presence || self[:login]
     end
   end
 


saracarl (changes requested):
This is fine, but it doesn't address what I suspect is the real cause of owner names being blank, which is the redirect to a different view when you encounter an error creating your trial account.

If you get errors sign up here (users/new_trial):
signup

You get redirected to here (users):
redirect

WillNigel23:
Owner names being saved in db as blank does not have much impact.

As long as real_name is set to something blank (i.e: nil, " " empty string, etc), we will be using 'self[:login]' which is always true.

At the end of the day, we save as nil if it is blank (which is the callback fix here).

WillNigel23:
On the redirect part, this seems like a separate bug if there are sign up errors in new_trial that redirects to a different sign up form route.

app/controllers/registrations_controller.rb
```diff
@@ -131,6 +128,15 @@ def after_update_path_for(resource)
     edit_registration_path(resource)
   end
 
+  def after_failed_sign_up_action_for(registration_type)
+    case registration_type
+    when :free_trial
+      render :new_trial
+    else
+      render :new
```
WillNigel23, line 136: If in case we add more registration_type, we can easily add it here

saracarl:
So there's one more change that's needed here to get to the desired behavior.  The "Organization Name" field is not required for a trial sign up.  It needs to be, so we actually force trials to have a display name (Organization Name = display name, just different label).

saracarl:
Tried this, and it did not require a Organization Name.  I was able to create a trial account without a display name.

WillNigel23:
Please try restarting running server.

WillNigel23:
This is ready @saracarl , let me know if there are any other change requests

PR #4365, opened 2024-10-02
4353 - Duplicate work fix

### LINKED ISSUE #4353: duplicate upload work
TN State Archives are seeing duplicate works after uploading:
duplicates

When I went to find an example, however, there was only one:
image
I'm asking if they deleted the duplicate.

I determined the file was just uploaded once, and had a successful upload processing.  The processing log, however, only shows creating pages for the first 40 pages, which is weird:

[view_processing_log.txt](https://github.com/user-attachments/files/17175432/view_processing_log.txt)

I also unzipped the uploaded file and looked for any duplicate folders; there were none.  

https://fromthepage.com/admin/view_processing_log?id=32026693

~/deployment/releases/20240920211105/public/uploads/document_upload/file/32026693

no deployments during or after -- this release folder is the current one.

WillNigel23:
Problem was `race` condition.

```

    begin
      Rake.application.init
      Rake.application.load_rakefile
      ...
    rescue => e
      ...
    ensure
      Rake::Task.clear
    end
```

We `Rake::Task.clear` so we don't load a rake task twice and execute twice (same problem by adding load_rakefiles in application.rb)

But problem here is, if we do 2 consecutive document_uploads (say, I have 2 tabs, and I submit both almost at the same time), one rake task is not yet finished executing so we still have not reached the `Rake::Task.clear` part and thus tasks are loaded more than once again.

What we need instead is real job queuing rather than the hacky way to avoid syscall I introduced. We should refactor this part soon, but for now reverting to old implementation.


saracarl (approved):
Reviewed this with the original PR for active job up next to it.  Looks good.  Waiting for tests to finish and will do some manual testing before merging.

PR #4381, opened 2024-10-09
Improve article coverage

WillNigel23:


app/controllers/article_controller.rb
```diff
@@ -1,89 +1,66 @@
...
+        @article = result.article
+        render :edit, status: :unprocessable_entity
       end
-    elsif params['autolink']
+    elsif params[:autolink]
       @article.source_text = autolink(@article.source_text)
+
       flash[:notice] = t('.subjects_auto_linking')
-      redirect_to :action => 'edit', :article_id => @article.id
+      render :edit
```
WillNigel23, line 60: We are doing `redirect_to` here before, which will reload and defeats the purpose of the unpersisted change. We should instead do `render :edit` so the changes apply.

@benwbrum

app/controllers/article_controller.rb
```diff
@@ -284,82 +217,38 @@ def record_deed
...
-        row = 0
-      end
     end
 
-    return vertical_articles
+    vertical_articles
   end
 
-  private
+  def find_or_create_category(collection, title)
```
WillNigel23, line 245: Moved this to a `private` function. It doesn't seem to be a route. Ideally we move helper functions like these to `private` and keep all public functions as `route/actions` only to help with readability.

app/models/category.rb
```diff
@@ -17,25 +17,17 @@
...
-#    end
-#    self.destroy
-#  end
+  # def destroy_but_attach_children_to_parent
+  #   self.children.each do |child|
+  #     child.parent = self.parent
+  #     child.save!
+  #   end
+  #   self.destroy
+  # end
```
WillNigel23, line 32: Can I remove these commented codes @benwbrum ?

app/models/document_set.rb
```diff
@@ -27,121 +27,107 @@ class DocumentSet < ApplicationRecord
...
-  has_many :works, -> { order 'title' }, through: :document_set_works
+  has_many :works, -> { order(:title) }, through: :document_set_works
+  has_many :pages, through: :works
+  has_many :articles, -> { distinct }, through: :pages
+  has_many :notes, -> { order(created_at: :desc) }, through: :works
+  has_many :deeds, -> (document_set) {
+    where(work_id: document_set.works.select(:id))
+      .includes(:work)
+      .reorder('deeds.created_at DESC')
+  }, through: :collection, source: :deeds
```
WillNigel23, line 49: We have all the necessary reference_ids/keys here. We can make `articles` and `deeds` an actual relationship rather than a function.

This allows for the possibility of even preloading these relationships which can help with performance.

app/models/document_set.rb
```diff
@@ -27,121 +27,107 @@ class DocumentSet < ApplicationRecord
...
+  delegate :text_language,               to: :collection
+  delegate :field_based,                 to: :collection
+  delegate :transcription_fields,        to: :collection
+  delegate :metadata_fields,             to: :collection
+  delegate :description_instructions,    to: :collection
+  delegate :facets_enabled?,             to: :collection
+  delegate :api_access,                  to: :collection
+  delegate :alphabetize_works,           to: :collection
+  delegate :institution_signature,       to: :collection
+  delegate :most_recent_deed_created_at, to: :collection
```
WillNigel23, line 130: The `delegate` syntax is just shorthand for what we are already doing.

config/routes.rb
```diff
@@ -157,10 +157,10 @@
   scope 'article', as: 'article' do
     get 'list', to: 'article#list'
     get 'tooltip', to: 'article#tooltip'
-    get 'delete', to: 'article#delete'
+    delete 'delete', to: 'article#delete'
```
WillNigel23, line 160: Use `delete` for this route

spec/factories/category.rb
```diff
@@ -1,3 +1,5 @@
 FactoryBot.define do
-  factory :category
+  factory :category do
+    title { "Category #{SecureRandom.uuid}" }
```
WillNigel23, line 3: I'm having trouble with unique id for categories (especially on old features spec) so this should work for now.

WillNigel23 (commented):
Please check the article specs I added (specifically the article_controller_spec, interactors/article/*_spec). I'm happy with the changes there, especially we have actual checking of DB records and not simply relying on frontend web displays for checking the correct changes when articles are updated/combined/deleted etc.

app/models/category.rb
benwbrum: Absolutely!

app/models/document_set.rb
benwbrum: That's a good point.  One question: this code was written before we started experiencing performance problems with `deeds`.  Do you think that that's a worry in the current code?  Can we avoid it in a new relation?

spec/factories/category.rb
benwbrum: That sounds fine to me.

app/controllers/article_controller.rb
benwbrum: Agreed 100%.

app/controllers/article_controller.rb
```diff
@@ -1,89 +1,66 @@
...
   def delete
-    if @article.link_list.empty? && @article.target_article_links.empty?
-      if @article.created_by_id == current_user.id || current_user.like_owner?(@collection)
-        @article.destroy 
-        redirect_to collection_subjects_path(@collection.owner, @collection)
-      else
-        flash.alert = t('.only_subject_owner_can_delete')
-        redirect_to collection_article_show_path(@collection.owner, @collection, @article.id)
-      end
+    result = Article::Destroy.call(
```
benwbrum, line 26: I'm confused by this -- this seems to skp a lot of the checks on links to the article being destroyed.  How does this new syntax address these?

app/controllers/article_controller.rb
```diff
@@ -1,89 +1,66 @@
...
-
-    @article.attributes = article_params
-    if params['save']
-      #process_source_for_article
-      if @article.save
-        if old_title != @article.title
-          rename_article(old_title, @article.title)
-        end
+    if params[:save]
+      result = Article::Update.call(
```
benwbrum, line 42: I'd also like to know more about this syntax -- why use `Article::Update.call` instead of `@article.update`?

app/controllers/article_controller.rb
```diff
@@ -1,89 +1,66 @@
...
+        article: @article,
+        article_params: article_params
+      )
+
+      if result.success?
         record_deed
-        flash[:notice] = t('.subject_successfully_updated')
-        if gis_truncated
```
benwbrum, line 76: Did we lose this warning message?

app/controllers/article_controller.rb
```diff
@@ -94,15 +71,14 @@ def article_category
...
+    render plain: 'success'
   end
 
   def combine_duplicate
-    #@article contains "to" article
-    params[:from_article_ids].each do |from_article_id|
-      from_article = Article.find(from_article_id)
-      combine_articles(from_article, @article)
-    end
+    Article::Combine.call(
```
benwbrum, line 78: I like moving `combine_articles` logic into the model, but am confused why we're doing this call syntax (again).

app/controllers/article_controller.rb
```diff
@@ -132,26 +105,17 @@ def show
...
-      # initialize for this rank if necessary
-      if count_per_rank[link_count]
-        # set this rank
-        count_per_rank[link_count] += 1
-      else
-        count_per_rank[link_count] = 1
-      end
-      #logger.debug("DEBUG: \tcount_per_rank[#{link_count}]=#{count_per_rank[link_count]}\n")
-    end
+      link_max = [link_count, link_max].max
```
benwbrum, line 112: This is beautiful!

app/controllers/article_controller.rb
```diff
@@ -191,43 +154,31 @@ def show
     session[:col_id] = @collection.slug
   end
 
-
   # display the article upload form
   def upload_form
   end
 
-
-  def find_or_create_category(collection, title)
```
benwbrum, line 200: Was this method dead code?  It's strange to exist in the controller...

app/controllers/article_controller.rb
```diff
@@ -191,43 +154,31 @@ def show
...
-  end
-
   # actually process the uploaded CSV
   def subject_upload
     @collection = Collection.find params[:upload][:collection_id]
     # read the file
     file = params[:upload][:file].tempfile
 
-#    csv = CSV.read(params[:upload][:file].tempfile, :headers => true)
+    # csv = CSV.read(params[:upload][:file].tempfile, :headers => true)
```
benwbrum, line 167: We can delete this line.

app/controllers/article_controller.rb
```diff
@@ -284,82 +217,38 @@ def record_deed
...
-      end
-    end
-    # write to the DB
-    to_article.save!
-    from_article.destroy
+  def authorized?
+    redirect_to dashboard_path unless user_signed_in?
   end
 
-  def gis_truncated?(params, dec)
```
benwbrum, line 330: Where has this logic moved to?

app/controllers/article_controller.rb
```diff
@@ -284,82 +217,38 @@ def record_deed
     update_search_attempt_contributions
   end
 
-  def combine_articles(from_article, to_article)
```
benwbrum, line 287: Where has this logic moved to?

app/controllers/article_controller.rb
WillNigel23: The goal here is to simplify the logic in controllers by adopting a higher-level architecture.
The idea is, the controllers should only be concerned with routing, handling params, responses, etc. Anything CREATE/DESTROY (or what we can consider anything commiting to the db) is handed off to an interactor.

Interactors are meant to be simple, straightforward, and single purpose. This allows us developers, at a single glance on the files, to know what the application does.
Say in this example, for articles:
/interactors/article/
   - combine.rb
   - destroy.rb
   - update.rb
   
Now we know these 3 actions are available for article. This is a very simple example, but I can imagine we can have single purpose interactors like for isolating the logic on `transcription/update`, `transcription/mark_as_blank`, `transcription/set_needs_review`, which will help simplify and isolate what happens when these are done.

I also love this quote from the interactor-rails team on why use an interactor:

```
SPOILER ALERT: Your use case won't stay so simple.

In [our](http://collectiveidea.com/) experience, a simple task like authenticating a user will eventually take on multiple responsibilities:

    Welcoming back a user who hadn't logged in for a while
    Prompting a user to update his or her password
    Locking out a user in the case of too many failed attempts
    Sending the lock-out email notification

The list goes on, and as that list grows, so does your controller. This is how fat controllers are born.
``` 

I would say that, save_transcription route is one example of a fat controller, where there are more than 5 nested ifs, some with obscure params/conditions that will cause a redirect/render and some branches does CRUD operations while others don't. 

I think a higher architecture will help simplify and untangle things here, which will make updating/fixing bugs easier. 
Admittedly, the examples I gave are technically working (no bugs), and so this refactoring to simplify things are lower on the priority for me. But at the same time I think this is essential for higher coverage -> higher confidence on tests -> higher confidence in rails upgrade not breaking things


BTW:
Article::Update is a class, check interactors/article/update.rb I simply moved the logic there so that it is simpler to read for us.

app/controllers/article_controller.rb
WillNigel23: This is not dead code (though it is untested). I merely moved it to private function as this is not a route. I suggest refactoring like this for all functions that are not routes in controller. We should be moving them to private functions

app/controllers/article_controller.rb
WillNigel23: I moved it to `interactors/combine`

app/controllers/article_controller.rb
WillNigel23: interactors/article/update.rb

app/models/document_set.rb
WillNigel23: I think this is a problem that went undetected, and as your db gets larger so does performance was affected.

The reason I suggest moving this in a relation is to preload it. Currently we have n+1 problems here whenever we call `deeds` for this model.

app/controllers/article_controller.rb
WillNigel23: I moved the checks to interactors/article/destroy.rb. It contains the same logic.

app/controllers/article_controller.rb
WillNigel23: Nope, it is all working as expected

app/models/document_set.rb
benwbrum: Let's do that, then.

app/controllers/article_controller.rb
benwbrum: Oh, interesting!

app/controllers/article_controller.rb
benwbrum: This makes sense for complex data manipulations like we see in the transcription controller and article controller (especially with things like `combine_duplicate`), though I'd prefer for us to not interactor-ize absolutely everything we touch.

app/controllers/article_controller.rb
benwbrum: Interactors -- got it!

app/controllers/article_controller.rb
benwbrum: Agreed completely.

app/controllers/article_controller.rb
benwbrum: Perfect.

spec/features/zz_convention_spec.rb
```diff
@@ -4,8 +4,8 @@
   before :all do
     @owner = User.find_by(login: OWNER)
     @collections = @owner.all_owner_collections
-    @collection = @collections.second
-    @work = @collection.works.last
+    @collection = @collections.find('imported-collection')
```
benwbrum, line 7: This is small, but it's a real improvement!

spec/interactors/article/combine_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe Article::Combine do
+  let(:user) { User.find_by(login: OWNER) }
+  let!(:collection) { create(:collection, owner_user_id: user.id) }
+  let!(:work) { create(:work, collection: collection, owner_user_id: user.id) }
+  let!(:from_related_page) { create(:page, work: work, source_text: '[[Duplicate]]', source_translation: '[[Duplicate]]') }
+  let!(:from_source_article) do
+    create(:article, collection: collection.reload)
```
benwbrum, line 9: What does `collection.reload` do here?

spec/interactors/article/combine_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe Article::Combine do
+  let(:user) { User.find_by(login: OWNER) }
+  let!(:collection) { create(:collection, owner_user_id: user.id) }
```
benwbrum, line 5: What's the difference between `let` and `let!`?

spec/interactors/article/combine_spec.rb
WillNigel23: at this point collection is an instance, after interactor the db changes have not been applied to this insttance of the object so we reload to requery from db the changes

spec/interactors/article/combine_spec.rb
WillNigel23: It eager saves, and does validations too. sometimes when no ! the object is not created when specs is run, or that it tried to  save, no error thrown but object failed to save so specs will still run with said objects that are not created.

spec/requests/article_controller_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+
+  describe '#tooltip' do
+    let!(:article) { create(:article, collection: collection, pages: [page], categories: [category]) }
+
+    let(:action_path) { article_tooltip_path(article_id: article.id) }
+    let(:subject) { get action_path }
+
+    it 'renders status and template' do
+      login_as owner
+      subject
```
benwbrum, line 22: Is this calling the subject mock defined in line 18?  I'm new to request specs.

benwbrum (approved):
I really, really like this!

benwbrum:
This is ready to merge

PR #4387, opened 2024-10-23
4372 - Notes filterable table component

WillNigel23:


app/assets/javascripts/plugins/filterable-table.js
```diff
@@ -0,0 +1,68 @@
+/*
+ * Author: willnigeldejesus@gmail.com
+ * Date:   22th Oct 2024
+ */
+
+(function($) {
+
+  $.fn.filterableTable = function(options) {
```
WillNigel23, line 8: Let me explain the approach here.

Right now I settled with least amount of architecture to make this work. Bulk of the work is going to be done server/rails side instead of here in js. In any case, for further customizations we can pass in `options` when initializing, but we can deal with that once we encounter the need for adding further customization.

So the approach is, for filterable tables, we have a form and the actual table is inside a form under a specific selector id of our  choosing (ex: "#notes-table").
Each filterable table will have by default, hidden fields for `:sort` and `:order`, params that we will use later in filtering/sorting in server side.

Then the simplest approach I could think of here is, we autosubmit when the form changes, make the xhr request, then server will give us an html partial as response and we jquery substitute the contents of '#table-selector' with the response of the server.

So the structure is
#form
   hidden_fields :sort, :order
   form_fields :per_page, :search, :etc
   #table // this is what we replace with server response partial

For sorting, we just attach event listeners for each sortable table-header, which is identifiable by 'data-sort' attr.

app/controllers/notes_controller.rb
```diff
@@ -1,8 +1,20 @@
 class NotesController < ApplicationController
   include ActionView::Helpers::TextHelper
+  DEFAULT_NOTES_PER_PAGE = 15
   PAGES_PER_SCREEN = 20
 
+  def index
+    filtered_notes
+
+    render partial: 'table' if request.xhr?
```
WillNigel23, line 9: Ideally we use turbo_streams here, but that will have  to wait until upgrade.

But this is similar pattern, as if we are doing format.html, and format.json, but in this case, if request.xhr we will render just the partial that we will use to replace the table contents.

app/controllers/notes_controller.rb
```diff
@@ -13,50 +25,118 @@ def create
...
+                        .where.not(last_note_updated_at: nil)
+                        .reorder(last_note_updated_at: :desc)
+                        .paginate(page: params[:page], per_page: PAGES_PER_SCREEN)
   end
 
-  def show
-    @note = Note.find(params[:id])
+  private
+
+  def filtered_notes
```
WillNigel23, line 87: For filtering/sorting actions, as much as possible we should do it server-side and as efficient as possible (avoiding n+1 queries if possible)

I think this is the simplest I can think of in order to have the notes table searchable by all columns. But if possible I'd like for us to just limit the search to note titles maybe? But if we really need to search with all available columns then this should work.

app/helpers/components/filters_helper.rb
```diff
@@ -0,0 +1,43 @@
+module Components::FiltersHelper
+
+  def fe_filter_table_wrapper(url:, selector:, sorting:, ordering:, static_params: {}, &block)
```
WillNigel23, line 3: This is the functional component approach I am taling about. Much like rendering a shared/partial but we hand it off first inside a ruby function so that we can do complex operations/params processing BEFORE handing it off to the partial. This way we avoid bulky ruby logic inside .html.slim files while still retaining the ability to re-use components easily.

`fe_` prefix for these functions just to denote that these are `frontend` component functions.

app/views/export/_table.html.slim
```diff
@@ -1,23 +1,27 @@
...
-      th[class="sorting #{params[:sort] == 'completed_count' ? sorting_class  : '' }" data-sort='completed_count']
-        =header
-      th[class="sorting #{params[:sort] == 'reviewed_count' ? sorting_class : '' }" data-sort='reviewed_count']
-        =t('export.index.review')
-      th.w100 =t('export.index.progress')
-      th(colspan="6") =t('export.index.export_as')
+table.datagrid.striped.dataTable.no-footer
+  thead
+    =fe_sortable_header(key: :title, sorting: @sorting, ordering: @ordering, classes: 'w80') do
+      =t('export.index.work_title')
```
WillNigel23, line 4: Using functional components, we simplified html.slim here from

```
    thead
      - sorting_class = "sorting_#{params[:order] || 'asc'}"
      th[class="w80 sorting #{params[:sort] == 'title' ? sorting_class : '' }" data-sort='title']
        =t('export.index.work_title')
      th[class="sorting #{params[:sort] == 'page_count' ? sorting_class : '' }" data-sort='page_count']
        =t('export.index.pages')
      -unless collection.subjects_disabled
        th[class="sorting #{params[:sort] == 'indexed_count' ? sorting_class : '' }" data-sort='indexed_count']
          =t('export.index.indexed')
      th[class="sorting #{params[:sort] == 'completed_count' ? sorting_class  : '' }" data-sort='completed_count']
        =header
      th[class="sorting #{params[:sort] == 'reviewed_count' ? sorting_class : '' }" data-sort='reviewed_count']
        =t('export.index.review')
      th.w100 =t('export.index.progress')
      th(colspan="6") =t('export.index.export_as')
```
to

```
    =fe_sortable_header(key: :title, sorting: @sorting, ordering: @ordering, classes: 'w80') do
      =t('export.index.work_title')

    =fe_sortable_header(key: :page_count, sorting: @sorting, ordering: @ordering) do
      =t('export.index.pages')

    -unless @collection.subjects_disabled
      =fe_sortable_header(key: :indexed_count, sorting: @sorting, ordering: @ordering) do
        =t('export.index.indexed')

    =fe_sortable_header(key: :completed_count, sorting: @sorting, ordering: @ordering) do
      =@header

    =fe_sortable_header(key: :reviewed_count, sorting: @sorting, ordering: @ordering) do
      =t('export.index.review')

    th.w100 =t('export.index.progress')
    th(colspan="6") =t('export.index.export_as')
```

We add in the benefit of making sure these components will produce similar looks as we are using partials, while maintaining the ability to add customizations like adding 'w80' class to some table headers here that need it.

app/helpers/components/filters_helper.rb
WillNigel23: Moreover here, I personally prefer named parameters so that we technically require the necessary variables, much like setting locals in partials. Though in terms of locals, we can easily miss when local_assigns was not set correctly I think.

WillNigel23: We can add params **opts though, if we really need more customizations in the future if we have optional parameters that does not make sense adding as explicit function params, but for now this architecture should serve what our datatables need.

app/views/export/index.html.slim
```diff
@@ -58,68 +54,3 @@ p
         }, 1000);
       });
     });
-
-    $(document).ready( function () {
```
WillNigel23, line 62: No need for custom js per data table now.

app/views/notes/_notes.html.slim
```diff
@@ -70,18 +70,27 @@
           } else if(xhr.status === 500) {
             var message = xhr.statusText;
             $error.text('Error: ' + message);
+            return false;
           } else {
             // Error occurred
-            var message = xhr.responseJSON.join('. ');
+            var message = xhr.responseJSON.errors.join('. ');
```
WillNigel23, line 76: I find that adding/editing notes does not show the flash messages. I added these and modified the json response so all responses in notes_controller now are technically json, but we can still get the flash messages.

app/views/notes/_table.html.slim
```diff
@@ -0,0 +1,50 @@
+table.datagrid.striped.dataTable.no-footer
```
WillNigel23, line 1: This is the pattern for the tables I think we can follow for filterable tables. Do we need a usage docs for this @benwbrum ?

config/locales/filter_table/filter-table-de.yml
```diff
@@ -0,0 +1,7 @@
+---
```
WillNigel23, line 1: Do the translations look okay to you?

WillNigel23:
Refactored export/list here to reuse the filterable table component. Works well with different types of models without creating custom note_datatable class :smile:

WillNigel23:
We still have some tables using old DataTables. I can refactor them on other tickets but for now focused on notes as this ticket is already large

benwbrum:
I'm really pleased with this.  The UI is good, and the performance is excellent. I spotted a few things that should be changed:

- [ ]  Let's get rid of the result number selector and use 50; 15 is too short for much of anything, and _All_ has the potential to bring the site down for huge collections.  _Update: I tested `All` on [a huge collection](http://localhost:3000/notes?sort=time&order=desc&collection_id=i-demokratins-namn&per_page=-1&search=) and the site is taking forever to load, mainly due to rendering `profile_picture`.
- [ ]  Clicking on the contents of a note--`A lot of figures need to be formatted` in the screenshot below--sends the user to the `show` route for a collection, giving a NoSuchMethod exception.  We should take the user to the page being displayed, anchored down to the note in the notes section.
- [ ]  Something's a little odd about the way we are wrapping the text for usernames.  We should try to wrap on whitespace or word boundaries, or expand the width of the column if this is impossible.  (See `fransalman` in the screenshot.)
- [ ]  We no longer have a link in the UI to this, but it would be nice to get `notes/list` for the whole site working again.  Currently we get a NoSuchMethod exception if I try to visit http://localhost:3000/notes/list
- [ ] I can start a search by tabbing away from the search field, but I cannot start a search by typing a term and hitting `Enter`

WillNigel23:
- [ ] Clicking on the contents of a note--A lot of figures need to be formatted in the screenshot below--sends the user to the show route for a collection, giving a NoSuchMethod exception. We should take the user to the page being displayed, anchored down to the note in the notes section.

My bad for this, I forgot there is no show route for a note. I think we should retain the existing implementation where the Note column is showing `plain text` title instead of a link. I think there exists notes without collection(where the collection was deleted already I think, but the notes still exist). So we should instead just keep the text not as a link.

- [ ] Something's a little odd about the way we are wrapping the text for usernames. We should try to wrap on whitespace or word boundaries, or expand the width of the column if this is impossible. (See fransalman in the screenshot.)

I see no screenshot. can you send it please? Not sure what needs to change in the appearance here

WillNigel23:
Added anchor on link click for notes title

benwbrum:
This is all fantastic.  My only remaining request is to change `http://localhost:3000/notes?collection_id=i-demokratins-namn to a more restful route like `/i-demokratins-namn/notes`

PR #4391, opened 2024-10-24
4374 syntax highlighting for unbalanced tags

### LINKED ISSUE #4374: Syntax highlighting for unbalanced XML/HTML tags
Currently the CodeMirror syntax highlighting paints text as red if a tag has been begun but not finished (i.e. `<add`), but not if a tag is unbalanced (i.e. `<add>` with no `</add>`).  We'd like to warn users that they have unbalanced tags if possible, so that they do not get surprised by error messages as in this video:

https://github.com/user-attachments/assets/198eebbf-6e09-475c-aa03-90a8700cd5d4

atbah:
Closes #4374 

@benwbrum FYI: The codemirror editor displays this errors for the unbalanced tags. 
image


saracarl (changes requested):
This works fine with TEI tags, but I'm worried that volunteers may insert (validly) HTML tags and those don't seem to be working as well:

image
specifically 

- the close br -- which I don't even know if that's valid html, but our parser wanted it closed 
- the valid h1 which we are throwing a warning on.

saracarl (approved):
This can be merged once tests pass.

Issue #4261, opened 2024-08-05
Export downloads cleaned before they ought to be


saracarl:
EM at CA State Lib is getting a 500 error trying to download an export she did 4 days ago on the 1st.  

Looking at [the download record](https://fromthepage.com/bulk_export/6220) the processing log has been cleaned, which implies that the export zip file has also been cleaned.

The [rake file for cleaning up exports](https://github.com/benwbrum/fromthepage/blob/bc39845853452e10de1eab504e25f21d8fb364f0/lib/tasks/bulk_export.rake#L5) is passed an argument of 7 days when the task runs in the crontab.  I double checked the logic, and that's what this code does.

The cron job that calls the rake task 
`00 5 * * * /bin/bash -l -c 'cd /home/fromthepage/deployment/current && RAILS_ENV=production bundle exec rake fromthepage:clean_bulk_exports[7] >> /tmp/cron_log_file7 2>&1'`

The code that "unlinks" (not actually deleting??) the export file:
https://github.com/benwbrum/fromthepage/blob/development/app/models/bulk_export.rb#L124

Looking at the /tmp directory, there are no downloads there:

> fromthepage@fromthepage:/tmp/fromthepage_exports$ pwd
> /tmp/fromthepage_exports
> fromthepage@fromthepage:/tmp/fromthepage_exports$ ls
> fromthepage@fromthepage:/tmp/fromthepage_exports$ 

Which is really weird because [an export that ran today is downloadable](https://fromthepage.com/bulk_export/6227) and says it's in that (empty) directory.  (So I"m thinking the empty /tmp directory is a red herring.)

It looks like exports run on August 2nd are also cleaned.  Exports run on August 3rd are not.

So why is the exports being if they are older than 3 days, rather than 7?

benwbrum:
We've done some more research on this and are stumped.  The strange thing is that, in production the rails console writes to `/tmp/fromthepage_exports` and we can see the outputs on the filesystem.  However, the web app does does not appear to write any files which are visible, even though it can display their results.

So running an export from the UI shows https://fromthepage.com/bulk_export/7007 including the log file
```


```

This should be pulled from a log file [see source](https://github.com/benwbrum/fromthepage/blob/337b29aea556b20056fbf7a46e047da7757b95cb/app/models/bulk_export.rb#L145) streamed from a syscall to a rake task
```
[...]

PR #4393, opened 2024-10-24
adding environment data and force buffer flushing for bulk exports

saracarl:


benwbrum:
This addresses--but may not fix--issue #4261

PR #4397, opened 2024-10-28
Replace link in configuration test mailer views with actual collection list link

bencomp:
There was a typo in the HTML mailer view, an extra 'a'. And the link that suggests it goes to the collection list is a link to the uploads. This PR fixes both.

### DIFF
--- a/app/views/system_mailer/config_test.html.erb
+++ b/app/views/system_mailer/config_test.html.erb
@@ -3,8 +3,8 @@
 </h2>
 
 <p>
-	Check the sender and from email addresses to make sure they're right.a
+	Check the sender and from email addresses to make sure they're right.
 </p>
 <p>
-	This link should go to the collection list: <%= link_to 'click me', { controller: 'admin', action: 'uploads', only_path: false } %>
+	This link should go to the <%= link_to 'collection list', { controller: 'admin', action: 'collection_list', only_path: false } %>.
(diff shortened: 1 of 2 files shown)


app/views/system_mailer/config_test.html.erb
```diff
@@ -3,8 +3,8 @@
 </h2>
 
 <p>
-	Check the sender and from email addresses to make sure they're right.a
+	Check the sender and from email addresses to make sure they're right.
 </p>
 <p>
-	This link should go to the collection list: <%= link_to 'click me', { controller: 'admin', action: 'uploads', only_path: false } %>
+	This link should go to the <%= link_to 'collection list', { controller: 'admin', action: 'collection_list', only_path: false } %>.
```
WillNigel23, line 9: LGTM

We would prefer using rails path helper rather than the `{ controller: ..., action: ... }` syntax here. In fact we are refactoring instances of that in other areas of the code.

WillNigel23 (changes requested):

app/views/system_mailer/config_test.html.erb
bencomp: Thanks for reviewing! I used the URL helper, since we need the full URL in the email.

WillNigel23 (approved):

PR #4399, opened 2024-10-28
4382 - Fix can't add collaborators as owner

WillNigel23:


benwbrum:
This works as requested.  However, in testing I observed a scenario which does not make sense:
* Edit a public collection
* Add user A as a project owner
* Block users
* Attempt to add user A as blocked

This results in user A being both a project owner (who should have full access to the project) and a blocked user (who should have no access to the project).  That configuration doesn't make sense, so can we remove the list of owners from possible users to block?

PR #4400, opened 2024-10-29
4380 - Warn owners that collections settings are overridden by works

WillNigel23:


saracarl (approved):
I especially like the "and 5 others" phrasing here.  

Fix the merge conflicts and we'll get this merged in.

PR #4401, opened 2024-10-30
Remove hardcoded 'www.' from "current URL"s in forms

bencomp:
It bothered me that "current URL" items in forms for updating a user profile or document set include `www.`, whereas our FromThePage does not have that. In fact, we don't have `www.` in our DNS records.

This PR updates the views to remove it. Based on a comment by @WillNigel23 on #4397 I used the URL helper where available, but I could not find a route for the document set.

### DIFF
--- a/app/views/document_sets/_edit.html.slim
+++ b/app/views/document_sets/_edit.html.slim
@@ -13,7 +13,7 @@
       tr
         th.hidden= t('.url')
         td 
-          p.nomargin.settings ==t('.current_url_for_this_work', current_url: "<b>#{Rails.application.routes.default_url_options[:host]}/#{@document_set.owner.slug}/#{@document_set.slug}</b>")
+          p.nomargin.settings ==t('.current_url_for_this_work', current_url: "<b>#{collection_url(@document_set.owner.slug, @document_set.slug)}</b>")
       tr
         td(colspan="2")
           =f.label :description, t('.description'), class: 'above'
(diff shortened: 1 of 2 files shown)


app/views/document_sets/_edit.html.slim
```diff
@@ -13,7 +13,7 @@
       tr
         th.hidden= t('.url')
         td 
-          p.nomargin.settings ==t('.current_url_for_this_work', current_url: "<b>www.#{Rails.application.routes.default_url_options[:host]}/#{@document_set.owner.slug}/#{@document_set.slug}</b>")
+          p.nomargin.settings ==t('.current_url_for_this_work', current_url: "<b>#{Rails.application.routes.default_url_options[:host]}/#{@document_set.owner.slug}/#{@document_set.slug}</b>")
```
WillNigel23, line 16: I believe we can use the url_helper for this too. That way no need to conditionally append `www.`

WillNigel23 (changes requested):

app/views/document_sets/_edit.html.slim
bencomp: I couldn't find the right helper. Which one do you recommend?

WillNigel23: I think this is `collection_path(@document_set.owner.slug, @document_set.slug)` if I am not mistaken.

WillNigel23: If we need the full url here then collection_url should work.

WillNigel23 (approved):

PR #4404, opened 2024-10-31
4343 - Fix collection level transcription changes not propagating down to works

WillNigel23:


app/controllers/work_controller.rb
```diff
@@ -193,75 +193,23 @@ def create
...
+    work = Work.find(params[:id])
+    result = Work::Update.call(work: work, work_params: work_params)
 
-    #if the slug field param is blank, set slug to original candidate
-    if work_params[:slug].blank?
-      @work.slug = @work.title.parameterize
-    end
+    @work = result.work
+    if result.success?
+      record_deed(@work, DeedType::WORK_ADDED, @work.owner) if result.original_collection_id != result.collection_id
```
saracarl, line 201: Doesn't work.update get called whenever the settings change?  We shouldn't be logging a WORK_ADDED deed for that.

saracarl: or maybe the question is when is a work "updated" vs "edit"ed?

benwbrum (approved):
This looks good.  We need to be sure to test
* Convention changes
* Collection changes
* Slug changes (including blanking it)
* Some attribute change that is none of those

WillNigel23:
Yes, will work on the specs here, I plan to be thorough as checking coveralls work#update is not tested 100%

PR #4405, opened 2024-11-05
4359 fixed comment issue on date field spreedsheet

### LINKED ISSUE #4359: "date" field in spreadsheet yellow with no data
If you create a spreadsheet in a field based project with a "date" field, the date field is always yellow because the data doesn't match the expected format, even if the date field is empty:

image

It should be transparent, not yellow, until data is entered.

atbah:
Closes #4359

### DIFF
--- a/app/views/shared/_handsontable.html.erb
+++ b/app/views/shared/_handsontable.html.erb
@@ -13,7 +13,8 @@
 
   edtfValidator = function (value, callback) {
     setTimeout(function(){
-      if (/^(19|20)\d{2}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])$/.test(value) || !value) {
+      // support YYYY-MM-DD format, restricting year to 1400-2000, months to 01-12 and das to 01-31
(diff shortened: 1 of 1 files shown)


app/views/shared/_handsontable.html.erb
```diff
@@ -13,7 +13,7 @@
 
   edtfValidator = function (value, callback) {
     setTimeout(function(){
-      if (/^\d\d\d\d(-\d\d)?(-\d\d)?$/.test(value)) {
+      if (/^(19|20)\d{2}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])$/.test(value) || !value) {
```
WillNigel23, line 16: Can you please add a comment here what this regex do? It will be hard for other devs to understand what this does.

Or store the regex on a descriptive variable.

WillNigel23 (changes requested):

app/views/shared/_handsontable.html.erb
benwbrum: I've added the comment

WillNigel23 (approved):

PR #4415, opened 2024-11-15
4414 - Fix typo in bulk export api

WillNigel23:


app/controllers/api/v1/bulk_export_controller.rb
```diff
@@ -17,60 +17,55 @@ def index
...
       else
         render status: 401, json: 'You must use an API token to access bulk exports'
       end
     end
 
-
     def start
       if @api_user
         collection_slug = params[:collection_slug]
-        if collection_slug
```
WillNigel23, line 30: This is redundant checking. The route requires /:collection_id so this will always be present. It should suffice that we check if collection.nil? instead.

app/controllers/api/v1/bulk_export_controller.rb
```diff
@@ -17,60 +17,55 @@ def index
...
+            status: bulk_export.status,
+            status_uri: api_v1_bulk_export_status_url(bulk_export.id),
+            download_uri: api_v1_bulk_export_download_url(bulk_export.id)
+          }
 
-            render status: 202, json: response.to_json
+          render status: 202, json: response.to_json
 
-          else
-            render status: 403, json: 'User #{@api_user} is not authorized to view #{collection.title}'
```
WillNigel23, line 70: This is one of the status typo. We are not interpolating the string here due to single quotes

spec/factories/user.rb
```diff
@@ -17,5 +17,24 @@
       sequence(:display_name) { |n| "admin_#{n}_login" }
       sequence(:login) { |n| "admin_#{n}_login" }
     end
+
+    factory :unique_user do
```
WillNigel23, line 21: Took the chance to add a temporary solution to one of my minor annoyances with Factories when combined with `transactional_fixtures = false` specs.

The issue is, doing just `sequence(:display_name) { |n| "admin_#{n}_login" }` without resetting the database each time we run specs IS FAILING at times, due to `n` here not accounting for duplicates (since each specs run does not delete the created data necessarily)

Using SecureRandom should work here. Initially I planned to add this to the already defined factory bots above, but it caused some legacy specs (especially Devise specs which needed to query  created user by `user_#{index}_login`). So I decided we should add a sub-factory here :unique_user so it will always create a unique user regardless if we reset test db or not.

spec/requests/api/v1/bulk_export_controller_spec.rb
```diff
@@ -0,0 +1,214 @@
+require 'spec_helper'
+require 'fileutils'
+require 'zip'
+
+describe Api::V1::BulkExportController do
```
WillNigel23, line 5: This should test each api route's if branches making sure everything works correctly.

Although, we are technically using factory bulk_export records here and not really checking the correct data processed (focusing only on response statuses)

I think it should be sufficient for the controller tests though. Given that this seems to be doing a lot of repetitive logic with the other bulk_controller, I am planning on writing a lib so that we just call that in the controller instead of doing the business logic in this controller (and then add the thorough data processed testing there)

app/controllers/api/v1/bulk_export_controller.rb
benwbrum: That sounds true, but what are we doing differently with a `collection_slug` parameter versus the usual `collection_id` (which should populate `@collection`?

app/controllers/api/v1/bulk_export_controller.rb
benwbrum: Let's fix it!

app/controllers/api/v1/bulk_export_controller.rb
```diff
@@ -17,60 +17,55 @@ def index
...
-            bulk_export.notes_csv = !(params[:notes_csv].blank?)
-            bulk_export.save
-            bulk_export.submit_export_process
+        if collection.show_to?(@api_user)
+          # try to parse the bulk_export
+          bulk_export = BulkExport.new
+          bulk_export.collection = collection
+          bulk_export.user = @api_user
+          bulk_export.status = BulkExport::Status::NEW
+          bulk_export.plaintext_verbatim_page = !params[:plaintext_verbatim_page].blank?
```
benwbrum, line 41: Can you remind me what the precedence is of `!` versus `.blank?` ?  I'm trying to figure out why we added parentheses here.

benwbrum (approved):
This looks great!

app/controllers/api/v1/bulk_export_controller.rb
```diff
@@ -17,60 +17,55 @@ def index
...
     def start
       if @api_user
         collection_slug = params[:collection_slug]
-        if collection_slug
-          collection = Collection.where(slug: collection_slug).first
-          if collection.nil?
-            render status: 404, json: "No collection exists with the slug #{collection_slug}"
-            return
-          end
+        collection = Collection.where(slug: collection_slug).first
```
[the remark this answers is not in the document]
WillNigel23: We only have :set_api_user in the filter. We also expect :collection_slug instead of :collection_id so it won't be set using our before filters.

I think what we have now is ideal. We want to throw the message `          render status: 404, json: "No collection exists with the slug #{collection_slug}"
`  since this is an API endpoint rather than a webpage. This allows user to know the error in their api calls.

app/controllers/api/v1/bulk_export_controller.rb
WillNigel23: `blank?` function produces boolean. We just invert the boolean value here.

Ideally I'd like to change this to `.present?` instead so we don't need to invert. But I kept this similar to old implementation as I may have thought this is a preferential thing. Do you want me to change it to `present?`

app/controllers/api/v1/bulk_export_controller.rb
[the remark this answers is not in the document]
benwbrum: You've convinced me.

app/controllers/api/v1/bulk_export_controller.rb
benwbrum: Sure!  I wasn't aware that `present?` existed universally.

benwbrum:
This is ready to sqash and merge as soon as the tests run

PR #4421, opened 2024-11-21
Install RuboCop Rails to actually run RuboCop in CI

bencomp:
I noticed that the "Run RuboCop" step in CI is always skipped, because RuboCop is not in the Gemfile(.lock). This PR adds RuboCop and RuboCop Rails. I followed <https://github.com/rubocop/rubocop-rails#rubocop-configuration-file> to add the `.rubocop.yml` file, although there is more to configure.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,1 @@
+require: rubocop-rails
--- a/Gemfile
+++ b/Gemfile
@@ -104,6 +104,7 @@ group :development, :test do
   gem 'pry-awesome_print'
   gem 'pry-byebug'
   gem 'rspec-rails'
+  gem 'rubocop-rails'
 end
 
 gem 'dotenv', group: [:development, :test], require: 'dotenv/load'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -321,6 +321,7 @@ GEM
       rexml
     kramdown-parser-gfm (1.1.0)
       kramdown (~> 2.0)
+    language_server-protocol (3.17.0.3)
     launchy (3.0.0)
       addressable (~> 2.8)
       childprocess (~> 5.0)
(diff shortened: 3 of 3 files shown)


WillNigel23 (commented):
We are in the process of upgrading rails at the moment. That includes getting rubocop as part of the CI process.

The updates done here will only do warning for syntax. We plan to get to this sometime after the rails upgrade (as there are syntax changes too that are different from older versions I believe). For now I think we should leave it as is.

bencomp:
Thanks for reviewing! However, I'm not sure I understand your comments.

RuboCop is already defined as a step in the CI process, i.e. the GitHub workflow. Currently it always fails because it isn't installed in the CI environment. This step is allowed to fail, so every time it runs, it produces an ignored error.
By default RuboCop runs as a linter that flags problems, without trying to fix them. In CI that is a good thing. Since there are very many violations of code style currently, it is probably a good thing to allow the RuboCop step to keep failing.
In this sense I'm leaving it as is.

I understand that Ruby 3 syntax is slightly different, but I don't think that matters here. As far as I can tell, RuboCop is smart and understands the version that you are using. I don't think this PR gets in the way of upgrading Rails.

Perhaps it is friendlier to contributors to eventually only run RuboCop on the code that they change/add in a PR, after deciding what (style) rules you want people to adhere to. Perhaps there are smarter ways of running RuboCop too, like [running it with Reviewdog in a separate job](https://github.com/reviewdog/action-rubocop?tab=readme-ov-file#example-usage).

benwbrum:
@WillNigel23 can you shephard this PR to completion?  We may have already added enough RuboCop support that this is obviated.

WillNigel23:
Updated to use default modern rails setup for rubocop (+ one custom rule)

Will require follow-up PRs:
1. Apply all rubocop lint suggestions
2. Require this step to throw error on CI.

PR #4429, opened 2024-11-25
3624 - Upgrade

WillNigel23:


WillNigel23:
# Some notable changes:

1. `.to_s(:db)` is deprecated. We now use `.to_fs(:db)` for formatting time string that is db compatible.
2. `File.exists?` and `Dir.exists?` methods are both now just `.exist?`
3. render syntax `render 'namespace/partial.html` is now not supported. We should remove the `.html` and instead pass it as `formats: [:html]`. Take a look at `notes/note.html`'s past implementation and compare to now.
4. For handling `.dot` files like `graph/graph.dot`, we introduced new template handler for `:dot` files. Like #3, we can pass it thru `formats: [:dot]`
5. I refactored interactors to inherit from `ApplicationInteractor` so that it follows a higher class architecture.
6. Some capybara methods are failing after update (due to not waiting for elements to fully load before accessing in tests). Adjusted these in legacy specs to use capybara methods that have built in `wait` implementation to avoid unnecessary timing errors.

## TODO (likely on a future ticket)
1. Add hotwire turbo and move away from handling ajax requests server side
2. Migrate old js files to use importmaps syntax

Gemfile
```diff
@@ -44,8 +47,8 @@ gem 'oink'
 gem 'pry'
 
 gem 'iiif-image-api', git: 'https://github.com/samvera-labs/iiif-image-api.git', branch: 'main'
-gem 'iiif-presentation', git: 'https://github.com/benwbrum/osullivan', branch: 'service_is_array'
-gem 'riiif', git: 'https://github.com/BrumfieldLabs/riiif.git', branch: 'quote-paths-for-shell'
+gem 'iiif-presentation'
+gem 'riiif'
```
WillNigel23, line 51: Checked these forks and compared with main, it seems the changes applied here in forks are done already in main (merged from another PR I guess).

So we should be good to get these from main repo.

Gemfile
```diff
@@ -75,10 +78,6 @@ gem 'terser'
 
 gem 'interactor-rails', '~> 2.0'
 
-group :assets do
-  gem 'uglifier'
-end
```
WillNigel23, line 80: Not used anymore

Gemfile
```diff
@@ -119,7 +118,7 @@ gem 'autoprefixer-rails'
 gem 'slim'
 
 # Gravatar Image Tag
-gem 'gravatar_image_tag'
+gem 'gravatar_image_tag', github: 'Tinix/gravatar_image_tag'
```
WillNigel23, line 121: Original gravatar_image_tag gem is not updated anymore. It contains deprecated functions like URI.encode.

This fork addressed that. Though eventually we may want to move away totally from this (as I think gravatar_image is just an API after all so we maybe don't need our own gem for this)

app/controllers/article_controller.rb
```diff
@@ -131,13 +131,17 @@ def show
...
+    dot_source = render_to_string(
+      partial: 'graph',
+      layout: false,
+      locals: {
+        article_links: article_links,
+        link_total: link_total,
+        link_max: link_max,
+        min_rank: min_rank
+      },
+      formats: [:dot]
```
WillNigel23, line 143: We added new mime_type :dot, which is handled like erb (technically how it is being handled before, but the updated syntax fails so we needed to add this)

app/controllers/deed_controller.rb
```diff
@@ -32,12 +32,12 @@ def list
 
     # Scope for date
     if params[:start_date]
-      start_date = params[:start_date].to_datetime.to_s(:db)
+      start_date = params[:start_date].to_datetime.to_fs(:db)
```
WillNigel23, line 35: to_s(:db) is deprecated, we now use to_fs(:db)

app/controllers/facets_controller.rb
```diff
@@ -95,7 +94,10 @@ def update
 
       redirect_to collection_facets_path(collection.owner, collection), notice: t('collection.facets.collection_facets_updated_successfully')
     else
-      render('collection/facets', :locals => { :@metadata_coverages => collection.metadata_coverages, :@errors => errors })
```
WillNigel23, line 98: global variables now can't be passed thru locals. We should just declare the globals here then

app/controllers/notes_controller.rb
```diff
@@ -28,7 +28,7 @@ def create
     if @note.save
       record_deed
       render json: {
-        html: render_to_string(partial: 'note.html', locals: { note: @note }, formats: [:html]),
+        html: render_to_string(partial: 'note', locals: { note: @note }, formats: [:html]),
```
WillNigel23, line 31: We pass in formats: [:html] instead of 'note.html' as that syntax is not supported anymore.

app/models/ahoy/event.rb
```diff
@@ -19,11 +19,11 @@ module Ahoy
...
     include Ahoy::Properties
 
-    self.table_name = "ahoy_events"
+    self.table_name = 'ahoy_events'
 
     belongs_to :visit, optional: true
     belongs_to :user, optional: true
 
-    serialize :properties, JSON
+    serialize :properties, coder: JSON
```
WillNigel23, line 27: New syntax. Old one is failing.

app/models/document_upload.rb
```diff
@@ -25,13 +25,13 @@ class DocumentUpload < ApplicationRecord
...
 
-  enum status: {
+  enum :status, {
     new: 'new',
     queued: 'queued',
     processing: 'processing',
     finished: 'finished',
     error: 'error'
-  }, _prefix: :status
+  }, prefix: :status
```
WillNigel23, line 34: New enum syntax

app/models/ia_leaf.rb
```diff
@@ -42,8 +42,8 @@ def iiif_image_info_url
   end
 
   def refresh_cache
-    unless File.exists?(cache_file_path)
-      unless Dir.exists?(cache_dir_path)
+    unless File.exist?(cache_file_path)
+      unless Dir.exist?(cache_dir_path)
```
WillNigel23, line 46: Function was renamed

app/models/ia_work.rb
```diff
@@ -165,7 +165,7 @@ def ingest_work(id)
 
     self.save!
     # now fetch the scandata.xml file and parse it
-    scandata_url = "http://#{server}#{dir}/#{URI.encode(scandata_file)}" # will not work on new format: we cannot assume filenames are re-named with their content
+    scandata_url = "http://#{server}#{dir}/#{CGI.escape(scandata_file)}" # will not work on new format: we cannot assume filenames are re-named with their content
```
WillNigel23, line 168: Deprecated function `URI.encode` is replaced by `CGI.escape`

config/i18n-tasks.yml
```diff
@@ -106,16 +106,34 @@ ignore_missing:
     - 'deed.deed.*'
     - 'article.graph.title'
     - 'thredded.*'
-  es: &NOT_EN
```
WillNigel23, line 109: Not supported anymore. This seems to be a security vulnerability that i18n have patched long ago.

config/initializers/mime_types.rb
```diff
@@ -2,3 +2,6 @@
 
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
+
+Mime::Type.register 'text/vnd.graphviz', :dot
+ActionView::Template.register_template_handler :erb, ActionView::Template::Handlers::ERB.new
```
WillNigel23, line 7: We register :dot handler

PR #4432, opened 2024-11-26
Use the appropriate Capybara methods.

Konboi:
Capybara provides some useful methods for E2E testing. These methods also wait automatically for target elements to appear. Therefore, using the provided methods instead of sleep usually makes tests more stable. For this reason, I tried replacing the existing code with these methods.

e.g.) 
[page.accept_confirm](https://rubydoc.info/github/jnicklas/capybara/Capybara/Session:accept_confirm) method


WillNigel23 (approved):
LGTM. :green_apple: 

 I'm sure there are still other capybara functions that can be improved. Nonetheless this is a good start!

PR #4441, opened 2024-12-05
4440 - Wrong recent edits showing up on collection due to cache key collision

WillNigel23:


app/views/collection/show.html.slim
```diff
@@ -160,8 +160,8 @@
       =link_to t('.show_more'), collection_notes_path(collection_id: @collection.slug), class: 'button outline round'
 
     h3 =t('.recent_edits')
-    -cache @collection.most_recent_deed_created_at do
-      =deeds_for({ :collection => @collection, :limit => 20, :types => DeedType.collection_edits, :long_view => true })
+    -cache [@collection.id, @collection.most_recent_deed_created_at, :collection_show] do
```
WillNigel23, line 163: I decided we should also add `:collection_show` as key to indicate the page this is being used. We can see in watchlist that it will cause another collision if we don't specify that this is for `:collection_show`

app/views/dashboard/landing_page.html.slim
```diff
@@ -1,7 +1,7 @@
 -unless params[:search]
   .carousel
     -@collections.each do |c|
-      -cache c do
+      -cache [c, :old_landing_page] do
```
WillNigel23, line 4: Here, passing the actual object `c` rather than `c.id`. This is to account for, when there are changes in `c`. 

If we just do `c.id` in key, then if there are changes, say new pages/works added, it will not be invalidated.

Passing `c` object, I believe, is being etag cached by rails by it's `updated_at`. So any changes in the collection SHOULD invalidate the cache here.

app/views/dashboard/new_landing_page.html.slim
```diff
@@ -1,7 +1,7 @@
 -unless params[:search]
   .carousel
     -@collections.each do |c|
-      -cache c do
+      -cache [c, :landing_page] do
```
WillNigel23, line 4: Here, passing the actual object `c` rather than `c.id`. This is to account for, when there are changes in `c`. 

If we just do `c.id` in key, then if there are changes, say new pages/works added, it will not be invalidated.

Passing `c` object, I believe, is being etag cached by rails by it's `updated_at`. So any changes in the collection SHOULD invalidate the cache here.

PR #4453, opened 2024-12-16
Call Mixpanel only when enabled

### LINKED ISSUE #4452: Mixpanel tracking is called on some pages while Mixpanel is disabled at installation level
PR is coming.

I noticed that in our installation (which has no Mixpanel integration set up) on some page there was still a piece of JavaScript that tried to track the page visit – which failed, because `mixpanel` was not defined on the page.

It turns out that whereas in many view templates this call is included only when `MIXPANEL_ID` is defined (in the `01fromthepage.rb` initialiser), in some templates it was not conditional.

bencomp:
This fixes #4452 by adding the condition for inclusion of the `mixpanel.track` calls to the views that did not have this.

### DIFF
--- a/app/views/contact/form.html.slim
+++ b/app/views/contact/form.html.slim
@@ -26,6 +26,5 @@
       <br>
       =button_tag t('.submit'), class: 'big'
 
-  javascript:  mixpanel.track("Contact Us");
-
-
+  -unless MIXPANEL_ID.blank?
+    javascript:  mixpanel.track("Contact Us");
--- a/app/views/contact/send_email.html.slim
+++ b/app/views/contact/send_email.html.slim
@@ -10,4 +10,5 @@
 
     =link_to t('.start_free_trial'), users_new_trial_path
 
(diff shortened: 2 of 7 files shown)


WillNigel23 (approved):
It seems there are errors in specs (but looks random error to me). I am rerunning the tests and if it passes, I think this should be good.

Tagging @benwbrum

bencomp:
Thanks for reviewing, @WillNigel23! After submitting this I found an issue suggesting everything related to Mixpanel needs to be removed, but I don't think it's a problem when we first merge this PR.

WillNigel23:
@bencomp I saw that in the issue you mentioned, and you did raise a good point. So that's really a question for @benwbrum  if we do want to totally remove everything related to mixpanel?

saracarl:
We can definitely remove everything related to mixpanel -- we aren't using it--but that work is lower priority than a lot of other work. I'm going to review and merge this one, and we'll let the existing issue hold the eventual work of removing mixpanel. (Although @bencomp  don't let me stop you if you feel like taking that.)

PR #4456, opened 2024-12-18
4424 - Feedback on settings update

WillNigel23:


saracarl (changes requested):
1. The UI for subject categories is broken/wonky:

image

2. I changed the title and the URL and received no confirmation.  When I moved on to another tab, it turns out they didn't save.  As far as I can tell, none of the settings are saving at all.

I'll review this again once #2 is fixed.

WillNigel23:
Can you please try again? Everything is working locally.

Some things that could have caused the error:
1. This requires some new gems installed. Please run bundle install first before running server
2. As per `1`, this means we need server restart on local development.

saracarl:
Weird.  The only thing I can figure out that may have happened is there were 2 4424 branches, and I got the older one.  This works great!  I really like it.

saracarl (approved):

PR #4459, opened 2024-12-31
4458 Correct tilesource for v3 manifests

### LINKED ISSUE #4458: v3 manifests do not use image tile source
Importing [this manifest](https://labs.history.state.gov/exist/apps/consular-cards/iiif/nara/211636167) works, but displaying the page images in OSD doesn't, as `TransrcribeHelper#osd_source` [looks]*https://github.com/benwbrum/fromthepage/blob/4b165f0d96cb8e9c9536820bc84b154f7e14255f/app/helpers/transcribe_helper.rb#L89-L94) for a service `@id`, doesn't find one, then renders ugly ruby hash for an image source in place of valid JSON.

We should move this logic into `ScCanvas`, adding an `image service` method that encapsulates this.

(Requested by @JoeWiz)

benwbrum:
Closes #4458

### DIFF
--- a/app/helpers/transcribe_helper.rb
+++ b/app/helpers/transcribe_helper.rb
@@ -86,12 +86,7 @@ module TranscribeHelper
       sources
     else
       if page.sc_canvas
-        if page.sc_canvas.sc_service_id
-          service_id = page.sc_canvas.sc_service_id.sub(/\/$/,'')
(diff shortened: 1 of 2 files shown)


saracarl (approved):
Works for the test case.  Opened #4460 to handle issues found during testing.

Issue #4438, opened 2024-12-04
Refactor to Load JavaScript Assets in <head> Tag for Efficient Script Execution


WillNigel23:
Currently, some JavaScript assets are loaded in the <body> tag, leading to redundant script executions on every page reload. By refactoring to load these assets in the <head> tag, we ensure that JavaScript functions are defined once during the initial page load and persist across page navigations.

WillNigel23:
I may have provided incomplete context here @bencomp.

Yes you are right, js fucnctions can not persist across page navigations. So the problem above described is indeed not happening at the moment.

However, as we move towards updating our rails versions and convention, we are now as well moving towards incorporating [turbo-rails](https://github.com/hotwired/turbo-rails) as this is a core part of newer versions of rails.

[This](https://github.com/hotwired/turbo-rails?tab=readme-ov-file#navigate-with-turbo-drive) part of the docs discusses how turbo will intercept page navigations and under the hood do `<body>` tag swaps instead. Given that we load js tags `inside` the body tags, this will now be problematic for us. In the current PR I have disabled that behaviour using `Turbo.session.drive = false`, so this will only be a problem for routes that uses turbo_stream responses. 

One existing example in our codebase where this is not ideal is [here](https://github.com/benwbrum/fromthepage/blob/468b85afc91afd0dffa613b50cb7bd2b27e61b94/app/assets/javascripts/ftp_interface.js#L129). We already declared it once on page load, doing body-swap will re-declare a `const` variable causing some js error.

> I noticed that, for example, Annotorious scripts are included on every page because it is in application.js – is that something you wanted to address with this issue?

as per this question, this explains a separate issue that I am also working towards on fixing (and in fact some parts of it are addressed in #4439). We are also incorporating [stimulus-rails](https://github.com/hotwired/stimulus-rails) for the same reason that it is a core part of newer rails versions. This will deal with only declaring instance js libraries when needed.

PR #4469, opened 2025-01-06
4433 - Page version username should match recent version

WillNigel23:


saracarl:
We won't test or merge this until #4438 is approved and merged, as this contains those changes, which haven't passed review yet.

PR #4482, opened 2025-01-14
4481 - Configuring screen ruler causes 500 error

WillNigel23:


benwbrum:
I just tested this, and it no longer creates the 500 error.

However, the original functionality no longer works after the changes we've made to the OSD over the last 18 months.  

Take a look at 2-3 minutes of [this video on the screen ruler feature](https://youtu.be/AyqoHR8FESQ?si=LeK3vZ_be5ddXH7-&t=923) to see what the original feature did.  At the time, I think that we used Annotorious as an OSD overlay in this row identification screen, but did not use Annotorious to display images for transcription.  Since then, we started using Annotorious on transcription screens to support the AI-Assist feature we built on Annotorious's text overlay module.  Part of AI Assist involved disabling Annotorious's default behavior drawing boxes on an image.  That conflict is probably the problem here.

PR #4492, opened 2025-01-23
4473 - Works list performance

### LINKED ISSUE #4473: Work List performance is slow due to n+1 queries
For collections or document sets with a large number of works, the Works List tab takes a long time to load.  [Kennicott Documents](https://fromthepage.com/the-grove-national-historic-landmark/kennicott-documents/works_list) takes 15 seconds, for example, executing 6535 sql queries to retrieve 2178 records.

While we can get rid of some functionality to speed up the page--Last Edit seems optional, for example--it looks like we're repeating a lot of queries that could be better handled by using `includes` or something similar:
Screenshot from 2025-01-08 10-14-18

WillNigel23:


benwbrum:
This is _blazingly fast_ now!

PR #4504, opened 2025-02-03
4129 - Fix order of search results

WillNigel23:


benwbrum:
One more thing to put in this PR: Currently the images do not hyperlink to the owner profile page, while the owner names do.  We want them both to link to the owner profile, since users will probably click on whatever is easiest.

PR #4511, opened 2025-02-05
4312 - Document sets need to show up in tag cloud

### LINKED ISSUE #4312: document sets need to show up in tag cloud
Our tag cloud in find a project doesn't include document sets.  It needs to.  We should propagate tags set at the collection level "down" to the document sets -- not in the UI or data model, but apply them to the document sets, and search and include document sets based on that data.

Zobedia at American ancestors reported this.

WillNigel23:


app/controllers/dashboard_controller.rb
```diff
@@ -224,7 +224,20 @@ def landing_page
 
   def browse_tag
     @tag = Tag.find_by(ai_text: params[:ai_text])
-    @collections = @tag.collections.unrestricted.has_intro_block.has_picture.not_empty
+    collections = @tag.collections
+                      .includes(:owner, { works: :work_statistic })
+                      .unrestricted
```
benwbrum, line 229: Imagine that you have a public document set from a private collection, as happens with the `cwrgm` user.  You would like the document set to appear in the list.

The `unrestricted` scope will prevent the public document set from appearing in the tag cloud.

benwbrum (changes requested):
This looks really good.  We do need to handle the (rather common) case of a public document set in a private collection.

saracarl:
Should there be a migration here?  I'm getting a
```
Mysql2::Error: Unknown column 'document_sets.is_public' in 'where clause'

```
When I try to load the FindAProject page.

PR #4534, opened 2025-02-19
4483 - Work added deed has duplicate info in template

WillNigel23:


saracarl (approved):
So when I added a work via upload to test this, I did not get a deed in the "recent edits" deed list:

image

But if I click on "Show More" and get the deed list/activity stream there I do:

image

This problem wasn't introduced by this PR, so I'm going to open a new issue for it and merge this in, as the original issue was addressed.

PR #4543, opened 2025-02-24
4522 - Delete notes on work/page/collection delete

WillNigel23:


db/migrate/20250224185703_add_fk_to_notes.rb
```diff
@@ -0,0 +1,9 @@
+class AddFkToNotes < ActiveRecord::Migration[6.1]
+  def change
+    change_table :notes, bulk: true do |t|
```
WillNigel23, line 3: @benwbrum @saracarl please run db:migrate on your local machine and commit all annotations and schema changes please :pray:

db/migrate/20250224184349_delete_orphaned_notes.rb
```diff
@@ -0,0 +1,7 @@
+class DeleteOrphanedNotes < ActiveRecord::Migration[6.1]
+  def change
+    Note.left_outer_joins(:page).where('pages.id IS NULL').in_batches(&:destroy_all)
```
WillNigel23, line 3: This took a while on local dev, might be worth scheduling this on a down time. And please let me know if we are good with this

saracarl:
The migration went pretty fast on my dev box; I think prod would be faster.  I don't think we need downtime.

PR #4549, opened 2025-02-27
ES Search fix highlight terms

WillNigel23:


app/models/search_attempt.rb
```diff
@@ -45,64 +45,11 @@ def update_slug
     update_attribute(:slug, "#{query.parameterize}-#{id}")
   end
 
-  def results
```
WillNigel23, line 48: bad merge, we are not using this anymore I believe.

PR #4560, opened 2025-03-03
4559 - Page reviewed count double counts fix

WillNigel23:


saracarl (changes requested):
This needs to address the same problem in "pages indexed".  (see issue)

WillNigel23:
Included it @saracarl , let me know if any other stats here need the same treatment

PR #4573, opened 2025-03-06
4562 - Improve ES test coverage

WillNigel23:


benwbrum:
I spent about an hour attempting to test this locally, but my configuration is so out of sync with production that I think we need to wait for #4582 to be merged into development before I can feel certain about this PR.

WillNigel23:
This is now unnecessary, closing.

PR #4578, opened 2025-03-13
4463 - Page transcription json

WillNigel23:


db/migrate/20250225194127_add_transcription_json_field_to_pages_table.rb
```diff
@@ -0,0 +1,7 @@
+class AddTranscriptionJsonFieldToPagesTable < ActiveRecord::Migration[6.1]
+  def change
+    change_table :pages, bulk: true do |t|
+      t.json :transcription_json, null: true
```
benwbrum, line 4: This is really cool -- I hadn't seen json columns in migrations before!

lib/tasks/migrate_transcription_json.rake
```diff
@@ -0,0 +1,49 @@
...
+    File.open(log_file, 'a') do |log|
+      log.puts "Starting migration from Collection ID: #{start_from_id} at #{Time.current}"
+
+      collections = Collection.where(id: start_from_id..Float::INFINITY)
+                              .where(field_based: true)
+                              .order(:id)
+
+      collections.find_each do |collection|
+        transcription_fields = collection.transcription_fields
+                                         .order(:line_number, :position)
```
benwbrum: We'll also need to order by `page_number`, which is only used by two customers.  That should come before `line_number`.

lib/tasks/migrate_transcription_json.rake
```diff
@@ -0,0 +1,49 @@
...
+            spreadsheet_columns_map: spreadsheet_columns_map
+          )
+
+          pages_to_update << TranscriptionField::Lib::Utils.parse_fields(page: page, field_cells: field_cells)
+
+          log.puts "Processed Page ID: #{page.id} in Collection ID: #{collection.id} at #{Time.current}"
+        end
+
+        unless pages_to_update.empty?
+          Page.import pages_to_update, on_duplicate_key_update: [ :transcription_json, :source_text ], batch_size: 1_000
```
benwbrum: Will `import` update the DB rows, or just insert new ones?  This method is new to me, so the documentation led me to believe that it was only used for batch inserts.

benwbrum: Another question -- this seems to be the only write operation in the migration.  Will we leave orphan/out-of-sync `TableCell` records for these pages?

app/views/export/_table.html.slim
```diff
@@ -56,12 +56,19 @@ table.datagrid.striped.dataTable.no-footer
...
               span= t('export.index.iiif')
-          -unless work.table_cells.blank?
-            -if work.table_cells.count < 10000
-              td.nowrap
-                =link_to export_table_csv_path(work_id: work.id), class: 'btnCsvTblExport btnExport'
-                  =svg_symbol '#icon-export', class: 'icon'
-                  span =t('export.index.table_csv')
+          ruby:
+            if @collection.field_based?
+              total_table_cells = work.pages.sum("JSON_LENGTH(transcription_json)")
```
benwbrum, line 61: What will this return for pure field-based projects (one record per page) vs. spreadsheet projects?

app/models/page.rb
```diff
@@ -20,6 +20,7 @@
 #  source_translation      :text(16777215)
 #  status                  :string(255)      default("new"), not null
 #  title                   :string(255)
+#  transcription_json      :json
```
benwbrum: I'm not sure that we can do this within our automatically-created model annotations, but it would be nice to document an example of what this might look like for a simple spreadsheet format (2 columns, 2 rows) vs a field-based project (3 fields).

app/models/collection.rb
```diff
@@ -330,10 +332,6 @@ def self.search(search)
     where(sql, "%#{search}%", "%#{search}%", "%#{search}%")
   end
 
-  def sections
```
benwbrum, line 335: Thanks for removing this.  I have no idea what I was thinking here.

lib/tasks/migrate_transcription_json.rake
WillNigel23: I believe I have added handling for that?

lib/tasks/migrate_transcription_json.rake
WillNigel23: The 'import' call is to do bulk inserts and it is from the gem activerecord import. It is more efficient compared to doing updates for each record.

WillNigel23: Let's deal with orphaned records later. I believe this migration will take a lot of time.


Also, we already implemented a way so that, if a page was not migrated to transcription_json yet, we will convert it on-demand.

app/views/export/_table.html.slim
WillNigel23: It should sum all outer keys for json if I recall correctly.

app/controllers/export_controller.rb
```diff
@@ -97,18 +97,48 @@ def work_metadata_csv
...
+    if collection.field_based?
+      result = Work::Table::ExportCsv.new(
+        collection: collection,
+        works: collection.works.includes(
+          { pages: [ :notes, { page_versions: :user } ] }, :deeds
+        ).where(id: @work.id)
+      ).call
+
+      csv_string = result.csv_string
+    else
```
benwbrum, line 112: This looks good.

app/controllers/export_controller.rb
```diff
@@ -97,18 +97,48 @@ def work_metadata_csv
...
+    if @collection.field_based?
+      result = Work::Table::ExportCsv.new(
+        collection: @collection,
+        works: @collection.works.includes(
+          { pages: [ :notes, { page_versions: :user } ] }, :deeds
+        )
+      ).call
+
+      csv_string = result.csv_string
+    else
```
benwbrum, line 134: This looks good.

app/interactors/work/table/export_csv.rb
```diff
@@ -0,0 +1,137 @@
+require 'csv'
+
+class Work::Table::ExportCsv < ApplicationInteractor
```
benwbrum, line 3: Would it make sense to test if the work is part of a field-based collection and error or bail out if it is not?  

Normally I don't think that this would be necessary, but--knowing that there are two possible code paths--it's not clear from the class name or comments that this work table export applies to field-based projects and not markdown-table projects.

WillNigel23: I'm not sure I understand. If I'm correct, this will never be called by a non-field based work 🤔

lib/tasks/migrate_transcription_json.rake
benwbrum: Can we split off a sub-issue for this migration, then?  94.6% of the records in `table_cells` will disappear once we delete the records, which is a pretty significant savings.

app/views/export/_table.html.slim
benwbrum: This might be better addressed on my comment on the page model, but I'm a little unsure about how we're representing the data here -- is it an array of hashes, or a hash?

app/interactors/work/table/export_csv.rb
benwbrum: Right, I get that -- this is just a question about future developers (including future-me and future-you) trying to debug a markdown table export problem, seeing this class and assuming (based on the class and module name) that this is the code that exports markdown tables.  It would be helpful to add a comment saying something like, "This module exports CSVs from field-based collections; the code that exports CSVs from document-based collections is at _filename_".

app/views/export/_table.html.slim
WillNigel23: => {"2280"=>"", "2281"=>[{"1"=>"My Title", "2"=>2, "3"=>"4"}]}

I will add a code comment as to the format of transcription_json

lib/tasks/migrate_transcription_json.rake
WillNigel23: https://github.com/benwbrum/fromthepage/issues/4877

db/migrate/20250225194127_add_transcription_json_field_to_pages_table.rb
WillNigel23: This is my first time working with this too!

app/models/page.rb
```diff
@@ -138,6 +140,33 @@ module TEXT_TYPE
   NOT_INCOMPLETE_STATUSES = COMPLETED_STATUSES + [ Page.statuses[:needs_review] ]
   NEEDS_WORK_STATUSES = [ Page.statuses[:new], Page.statuses[:incomplete] ].freeze
 
+  # Expected format for `transcription_json`
```
WillNigel23, line 150: @benwbrum this should be enough for annotations. Let me know if it is clear enough for the formatting.

app/interactors/work/table/export_csv.rb
```diff
@@ -0,0 +1,139 @@
...
+  def initialize(collection:, works:)
+    @collection = collection
+    @works = works
+    @owner = @collection.owner
+
+    super
+  end
+
+  def perform
+    raise ArgumentError, "Collection with id=#{@id} is not field-based." unless @collection.field_based?
```
WillNigel23, line 31: @benwbrum this should prevent wrong usage of this interactor and improves readability.

benwbrum:
@WillNigel23 I just started merging development into this branch, but discovered that a lot of the conflict is with refactoring work which I don't want to mess up.  I'm going to test the branch as-is, then ask you to resolve the merge conflicts when that's done.

WillNigel23:
Working on getting the merge conflicts resolved

benwbrum:
I've tested spreadsheet transcription and exports successfully.  Still working on markdown tables.

app/models/page.rb
```diff
@@ -90,6 +91,13 @@ class Page < ApplicationRecord
 
   serialize :metadata, type: Hash
 
+  # TODO: We need to upgrade our DB version to utilize native json column field.
+  # Right now we are technically using long-text field and serializing to JSON
+  if (col = columns_hash['transcription_json']) &&
```
WillNigel23, line 96: As we discovered recently, it seems our production DB does not support native json field columns.

As such we need backsupport for it. This should allow for that.

benwbrum:
I just tried testing the export code and found that it works fine for markdown-table projects, but does not appear to work correctly for field-based projects. 

I tried my local copy of `missouri-state-penitentiary-register-of-inmates`, which is a field-based project that has at least some data.  In development, the logic for detecting the presence of tables/fields works for the work `MSP Vol 105`

### Development
Screenshot from 2025-11-05 07-31-19


### This branch
Screenshot from 2025-11-05 07-38-04

Similarly, a Table/CSV export from `development` has several rows, while the export from this branch only has headers.

benwbrum:
There is still a problem exporting field-based projects.  If I `nil` the `transcription_json` for a work, then export it using the Table/CSV link on the `export#show` screen, the migration happens and I am able to open the CSV file with the correct data.  However, running the same export in the background via the work Download tab produces an empty CSV file.

db/migrate/20250225194127_add_transcription_json_field_to_pages_table.rb
saracarl: We just realized we need this in page versions as well as pages.

saracarl: (We didn't track this before because the diff was based on the page_xml, but now we want to compare field by field for some of the AI strategy work.  We won't need a new UI -- YET! :)  )

app/models/page.rb
```diff
@@ -90,6 +91,13 @@ class Page < ApplicationRecord
 
   serialize :metadata, type: Hash
 
+  # TODO: We need to upgrade our DB version to utilize native json column field.
```
saracarl, line 94: Please open an issue (include a link to documentation for the db field) and assign it to Ben and I.

saracarl: What happens if we have serialized JSON data in the database in a text field, but then you upgrade the db, will we need an additional migration?

app/models/page.rb
```diff
@@ -139,6 +147,33 @@ module TEXT_TYPE
...
 
+  # Expected format for `transcription_json`
+  # {
+  #   "transcription_field_id" => "content",
+  #   "transcription_field_id" => "content",
+  #   ...
+  # }
+  #
+  # For spreadsheet fields, we expect array of key-value pairs for
+  # each spreadsheet_column ID PER ROW
```
saracarl, line 158: This should handle multiple spreadsheets per page, but we want to double check the export and test that.

app/assets/javascripts/ftp_interface.js
```diff
@@ -381,6 +380,8 @@ function freezeTableColumn(topEl, tableEl, columnEl, mode='') {
   }
 }
 
+window.freezeTableColumn = freezeTableColumn;
```
saracarl, line 383: What does this do?  (and why in this PR?)

saracarl (changes requested):
Ok.  This is the (most recent!) code review.  The biggest problem we saw was that we now need transcription_json on page versions.

We'll add another comment for some manual testing we came up with as we did this review.

app/models/page.rb
saracarl: The code looked like it handled this.

saracarl:
To test manually:
- [x] Test 2 spreadsheets on a page
- [x] test viewing version comparisons
- [x] test exporting a "hybrid" spreadsheet/field based collection
- [x] test with multipage field based forms -- LVA
- [x] test page preview
- [x] edit transcription fields / preview on save

- [x] is there a cleanup (to clean up table cell data in Db instaed of json) rake task?  (if not, create an issue)
- [ ] run the rake task at 10AM on a Monday

app/models/page.rb
WillNigel23: I think we need additional migration, but should not be complicated

app/assets/javascripts/ftp_interface.js
WillNigel23: It's a code in the old implementation of handsontable. I moved it to somewhere that can still be accessed. I'll review what it does but if memory serves right I think this one is neeeded.

app/models/page.rb
saracarl: Let's add that to the issue as well.

saracarl:
Test Failure 1:

I have a spreadsheet configured project -- 2 spreadsheets on the same line of a field based project.  With data.  When the first work is exported as a text only pdf it fails with:

```
2025-11-12T13:06:22-06:00 Exporting work	32045342	[Letter from James Steele to Peter Steele, December 16, 1786]
rake aborted!
ActionView::Template::Error: Malformed XML: No root element (ActionView::Template::Error)
Line: 1
Position: 21
Last 80 unconsumed characters:
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/baseparser.rb:271:in `pull_event'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/baseparser.rb:249:in `pull'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/treeparser.rb:21:in `parse'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/document.rb:468:in `build'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/document.rb:105:in `initialize'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/abstract_xml_helper.rb:168:in `new'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/abstract_xml_helper.rb:168:in `xml_to_html'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/export_helper.rb:43:in `xml_to_pandoc_md'
/home/saracarl/dev/products/fromthepage/fromthepage/app/views/export/facing_edition.html.erb:34:in `block in _app_views_export_facing_edition_html_erb__1704431750000358296_37420'
```

Here's what the page data looks like:
```
3.3.5 :002 > print p.xml_text
<?xml version='1.0' encoding='UTF-8'?>    
      <page>
        <table class='tabular'><thead><th>field 1</th><th>field 2</th><th>field 3</th></thead><tbody><tr><td>a</td><td>2</td><td>false</td></tr></tbody></table><table class='tabular'><thead><th>field 4</th><th>field 5</th></thead><tbody><tr><td>c</td><td>1998-09-10</td></tr></tbody></table>
      </page> => nil 
3.3.5 :003 > print p.transcription_json
{"4897"=>[{"1185"=>"a", "1186"=>2, "1187"=>false}], "4898"=>[{"1188"=>"c", "1189"=>"1998-09-10"}]} => nil 
3.3.5 :004 > 

```

saracarl:
Test Failure 2:
When I ask for a CSV export at a work level:
image
I expect to get a fields_and_tables.csv in each work folder that contains only the fields/tables for that work.  
image

Instead, I get a csv that contains the entire collection's fields/tables.  (Duplicated in each work folder.)


image

benwbrum:
If you go to a field-based document set, and click the Export tab, you get this exception:
```
NoMethodError - undefined method `field_based?' for an instance of DocumentSet:
  app/views/export/_table.html.slim:60
  app/views/export/_table.html.slim:24
  app/views/export/index.html.slim:32
  app/helpers/components/filters_helper.rb:27:in `block in fe_filter_table'
  app/views/shared/components/_filter_table.html.slim:2
  app/helpers/components/filters_helper.rb:26:in `fe_filter_table'
  app/views/export/index.html.slim:31
  app/helpers/components/filters_helper.rb:6:in `block in fe_filter_table_wrapper'
  app/views/shared/components/_filter_table_wrapper.html.slim:2
  app/helpers/components/filters_helper.rb:5:in `fe_filter_table_wrapper'
  app/views/export/index.html.slim:26
```

app/models/page.rb
saracarl: @WillNigel23 please document this in an issue so we don't lose it!

saracarl:
@WillNigel23 this is very close, but @benwbrum would like `transcription_json` on page_versions as mentioned here before we deploy:  https://github.com/benwbrum/fromthepage/pull/4578#discussion_r2518916033
We'll need the db side and the page save side.

I'd like an issue for the clean up task so we don't lose it: 
https://github.com/benwbrum/fromthepage/pull/4578#discussion_r2518930741

Once we get those 2 things in place we'll merge and deploy.  We want to deploy early in the week (and probably not Thanksgiving week -- negotiable, though.).

PR #4587, opened 2025-03-20
4585 - Inline tables break plaintext export

WillNigel23:


app/models/xml_source_processor.rb
```diff
@@ -494,7 +494,7 @@ def pipe_tables_formatting(text)
     text.split("\n").map{|line| "|#{line}|"}.join("\n")
   end
 
-  def xml_table_to_markdown_table(table_element, pandoc_format=false)
+  def xml_table_to_markdown_table(table_element, pandoc_format=false, plaintext_export=false)
```
WillNigel23, line 497: The whole xml_source_processor as a module is imported/included in so many different ways and is depending on helper files that may or may not be imported depending on the context it is used.

As such I'd like to be able to clean this up in the future. I'll create an issue. But for now, to deal with critical bugs we can do this.

benwbrum: This looks like a good compromise!

PR #4592, opened 2025-03-25
Handle markdown table exports for heterogeneous tables

### LINKED ISSUE #4584: inline tables not exported
Inline spreadsheets are not being exported when a document level table/csv export is selected.  The export works without error, but only produces a header row; no data.

Here's the first page with a table on it:
https://fromthepage.com/carnegiemnh/holland-letter-archive/volume-1-6df1be04-2570-49e6-ba2b-d37c4b80dfbb/display/34233659

Here's the exported/empty spreadsheet.  Note it has a column named "falua"; searching the document does not find a "falua".

[fields_and_tables.csv](https://github.com/user-attachments/files/19290642/fields_and_tables.csv)

benwbrum:
Closes #4584

### DIFF
--- a/app/controllers/concerns/export_service.rb
+++ b/app/controllers/concerns/export_service.rb
@@ -474,7 +474,8 @@ module ExportService
       spreadsheet_count = input_types.count("spreadsheet")
       position = input_types.index("spreadsheet")
     else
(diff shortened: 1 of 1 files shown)


WillNigel23 (approved):
LGTM :green_apple:

Issue #4596, opened 2025-03-26
can't parse markup in table markdown


saracarl:
A markdown table with superscript (and other??) tags in it used to save, but no longer does.

This markdown:
```
A.B. Mon.<sup>r</sup> Gen<sup>t</sup>.  | Thur. | Friday | | Sat. | Monday | | Tuesday | | Wednesday | | Thurs.  | Bill. exam.<sup>d</sup> |    |
Names | E | M | E | M | M| E | M | E | M | E | M |  Remarks | S. D. | 
J. A. |   | X | X |   |   |   |   |   |   |   |   | excused |   |   
J A |   |   |   |  1 |   |   |   |   |   |   |  | Paid | 00 .. 1/2 
T. B |   |   |   |   |   |   |   |   |   |   |   |   |   | 
&c. |   |   |   | X |   |   | X | X |   |   |   | Paid | 0 .. 2 |
|   |   |   |   |   |   |   |   |   |   |   |   |   |   |
|   | X |   |   |   | X | X | X |   |   |   |   |   | 00 .. 4 |
|   |   |   |   |   |   |   |   |   |   |   |  | the whole | 1 .. 6 1/2 | 
```

when saved, produces this error:
```
There was an error parsing the mark-up in your transcript. This kind of error often occurs if an angle bracket is missing or if an HTML tag is left open. Check any instances of < or > symbols in your text. (The parser error was: #r Gent."> /home/fromthepage/deployment/shared/bundle/ruby/2.7.0/gems/rexml-3.4.1/lib/rexml/text.rb:132:in `check' /home/fromthepage/deployment/shared/bundle/ruby/2.7.0/gems/rexml-3.4.1/lib/rexml/attribute.rb:176:in `element=' /home/fromthepage/deployment/shared/bundle/ruby/2.7.0/gems/rexml-3.4.1/lib/rexml/element.rb:2380:in `[]=' 
```

When I change it to be (without superscript markup):

```
A.B. Mon.r Gent.  | Thur. | Friday | | Sat. | Monday | | Tuesday | | Wednesday | | Thurs.  | Bill. exam.d |    |
Names | E | M | E | M | M| E | M | E | M | E | M |  Remarks | S. D. | 
J. A. |   | X | X |   |   |   |   |   |   |   |   | excused |   |   
J A |   |   |   |  1 |   |   |   |   |   |   |  | Paid | 00 .. 1/2 
[...]

Issue #4570, opened 2025-03-06
wall of red error when saving a markdown problem


saracarl:
If you try to save a table formatted like this:

```
|
| Page
Boyce, L. J. | 30
Beasley, Elira | 97
Bolo, Clarence | 220
Anderson, Tony | 128
Alison, Lewis | 231
```

You get a truly awful "wall of red" error message:

![Image](https://github.com/user-attachments/assets/f29a8c84-c443-45ae-a6df-509abb9cbdaa)

This is very transcriber-unfriendly.

Can we just display the initial, human readable, error message?

We also seem to have gotten more restrictive on what we're willing to accept for tables, since this page saved originally just fine.

https://fromthepage.com/page_version/show?page_version_id=39652859

(Can we figure out why?)

Issue #4528, opened 2025-02-14
Text below a table isn't included in pdf exports unless there are at least 3 blank lines between the table and the text.


saracarl:
This works, in that it exports the table as well as the Goodwyn and Goodwyn line in a pdf:
```
a

| No. |Tare |No.  |  |
| --- | --- | --- | --- |
|2728  |101  |1114  |<i>Thomas Pollitt.</i>  |
|1894  |106  |1156  |<i>Robert Westmoreland.</i>  |
|1051  |106  | 807  |<i>James Bradby</i>  |
|1380  |124  | 1015 |<i>John Barns.</i>  |
|652  |155  |1191  |<i>Young M<sup>c</sup>Lamore.</i>  |
|2493  |108  |1127  |<i>John Kelly.</i>  |
|2523  |137  |1027  |<i>James Bradley.</i>  |
|1344  |125  |1162  |<i>Bobert Cunningham</i>  |



<i>GOODWYN and GOODWYN.</i>
```

But this doesn't:

```
a

| No. |Tare |No.  |  |
| --- | --- | --- | --- |
|2728  |101  |1114  |<i>Thomas Pollitt.</i>  |
|1894  |106  |1156  |<i>Robert Westmoreland.</i>  |
|1051  |106  | 807  |<i>James Bradby</i>  |
|1380  |124  | 1015 |<i>John Barns.</i>  |
|652  |155  |1191  |<i>Young M<sup>c</sup>Lamore.</i>  |
|2493  |108  |1127  |<i>John Kelly.</i>  |
|2523  |137  |1027  |<i>James Bradley.</i>  |
|1344  |125  |1162  |<i>Bobert Cunningham</i>  |

<i>GOODWYN and GOODWYN.</i>
```
The goodwyn and goodwyn line is missing completely from the pdf, although it does show up in the read view in the app.

PR #4600, opened 2025-03-31
4570 - Fix wall of red error when saving markdown table

WillNigel23:


WillNigel23:
Fixes #4596, #4570

app/models/xml_source_processor.rb
```diff
@@ -242,10 +242,8 @@ def process_linewise_markup(text)
           cells = clean_line.split(/\s*\|\s*/, -1) # -1 means "don't prune empty values at the end"
           current_table[:rows] << cells
           rowline = ''
-          cells.each_with_index do |cell, i|
-            head = current_table[:header][i]
-            role_string = " role=\"#{head}\""
-            rowline += "<td#{role_string}>#{cell}</td> "
+          cells.each_with_index do |cell, _i|
+            rowline += "<td>#{cell}</td> "
```
WillNigel23, line 246: Before this change, it was
```
          cells.each_with_index do |cell, i|
            head = current_table[:header][i]
            role_string = " role=\"#{head}\""
            rowline += "<td>#{cell}</td> "
```
where head and role_string was unused. I thought this was unintentional so I added it back in #4528 , but it seems this caused the errors we are having.

benwbrum: That's amazing!  I have no idea what the `role` attribute was meant for, but am sure that removing it won't break anything.

PR #4626, opened 2025-04-24
4619 - Transcribers with no activity should redirect to FindAProject

WillNigel23:


app/controllers/application_controller.rb
```diff
@@ -297,8 +297,10 @@ def after_sign_in_path_for(resource)
       else
         dashboard_startproject_path
       end
-    else
+    elsif current_user.deeds.any?
```
WillNigel23, line 300: We have this: session[:user_return_to]

So I assume that should still take precedence instead of having no deeds correct?

PR #4643, opened 2025-05-05
3936 - Option to include notes in pdf exports

WillNigel23:


benwbrum:
I would like to remove the bullet point in the note listing, since the italicized username gives the reader enough information to know that a new note is starting.  This should also give us more room in the comments.

Would it be possible to reduce the size of the font for notes, so that they appear smaller than the text font?

benwbrum:
Is there any way to retain the width of the notes, or reduce the page margin around them?  Otherwise this looks fantastic:

Screenshot from 2025-05-14 14-55-06

saracarl (changes requested):
There's a problem with this that didn't show up in the earlier version Ben reviewed.  

I tried the configurable PDF export on a 2 page letter:
image

My export has "double pages" -- one for the content, and one for the notes (even if there are no notes on that page):
image
[text.pdf](https://github.com/user-attachments/files/20350374/text.pdf)

So I don't think we want notes on a separate page with a duplicate page title; I think want them immediately after the text of the page.

I definitely think we don't want a empty page with the same name if there are no notes.

WillNigel23:
After investigation, I found out the issue @saracarl  is experiencing here is not due to the notes export, but a bad nil check for source_translation.

```
### <%= page.title %>
<% unless page.status_blank? %>
<%= raw(xml_to_pandoc_md(page.xml_text, @preserve_linebreaks, true)) %>
<% end %>

<% if @work.supports_translation && !page.xml_translation.blank?%>
### <%= page.title %>
<%= raw(xml_to_pandoc_md(page.xml_translation, @preserve_linebreaks, true)) %>
<% end %>
```
We print the page.title twice, once for xml_text and the other for xml_translation.

Problem here is page.xml_translation.blank? is returning false, due to it's content being this:
`"<?xml version='1.0' encoding='UTF-8'?>    \n      <page>\n        <p/>\n      </page>\n"`

We should instead check for source_translation.blank? to ensure xml_translation does not have anything worth showing.

Even though not directly related to notes export, I decided to add the fix here and noted here in comment that we dealt with this here in case duplicate issue arise.

PR #4645, opened 2025-05-07
4464 - Collection data export task

WillNigel23:


app/interactors/database/export/dump_builder.rb
```diff
@@ -0,0 +1,220 @@
...
+    return @users if defined?(@users)
+
+    direct_owners = User.where(id: collections.select(:owner_user_id))
+    owners = User.where(id: collection_owners.select(:user_id))
+    collaborators = User.where(id: collection_collaborators.select(:user_id))
+    reviewers = User.where(id: collection_reviewers.select(:user_id))
+    blocked = User.where(id: collection_blocks.select(:user_id))
+    scribes = User.where(id: transcribe_authorizations.select(:user_id))
+
+    @users = direct_owners.or(owners)
```
benwbrum, line 211: We're probably also going to need to import user records who are linked from `deeds` and `page_versions` as well, unless I'm missing something.

benwbrum (changes requested):
A couple of minor changes.

app/interactors/database/export/dump_builder.rb
WillNigel23: Checking

WillNigel23:
Few notes here how I manually tested this locally.

# Database Export

1. run `rake fromthepage:export:database_dump[slug-1,slug-2]` where params passed are collection slugs separated by ", "
2. This will create a directory at `Rails.root.join('tmp/dumps', Time.current.to_s)` We are going to need the current_time for the import later.

# After Export
1. Shut down rails server
2. Go to `database.yml`, change the database name for development `diary_development_2` for example
3. Run `rails db:create && rails db:migrate` to setup new db
4. Start rails server and confirm that there are no data present


# Database import
1. Using the current_time key we found earlier, run `fromthepage:import:database_dump[current_time]`. Ex: `fromthepage:import:database_dump[1747245366]`
2. If successful, check in running server if records were created with correct assets.

PR #4666, opened 2025-05-27
4582 cleanup es implementation

WillNigel23:


WillNigel23:
# Elasticsearch Development Environment Setup Instructions

1. Download elasticsearch (preferrably version 8.15.0)
  - Extract the files. You should have folder `elasticsearch-#{version-number}`
  - Find `/config/elasticsearch.yml` and add these configs:
  ```
  xpack.security.enabled: false
  xpack.security.http.ssl.enabled: false
  xpack.security.transport.ssl.enabled: false
  ```
    -- You can also specify here at which port you want elasticsearch to run, but the default is at :9200
  - Go to /bin, and run `./elasticsearch`
    -- This should open a process. To confirm if it worked, go to http://localhost:9200 and you should see something like image:
    -- Leave the process running
  - If there are any issues, refer to https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-from-archive-on-linux-macos
  
2. Initializing indices
- Once elasticsearch is running, we can run rake tasks to initialize the FTP indices:
- First, set these environment variables:
```
ELASTIC_ENABLED='true'
ELASTIC_HOST_URL='http://localhost:9200' # Change to the port you indicated in ES config if you use custom port
ELASTIC_SUFFIX='dev' # this can be left blank if we do not want suffix on our indices.
```
- Run `bundle exec rake 'fromthepage:es:setup:init'`
  -- This will setup initial templates, scripts, and pipelines
- Run `bundle exec rake 'fromthepage:es:data:build'`
  -- This will initialize the indices under ftp_#{index_name}_#{suffix}
  -- To confirm if this worked, run `bundle exec rake 'fromthepage:es:data:stats'` to get indices and their stats.
- Run `bundle exec rake 'fromthepage:es:data:reindex'`
  -- This will reindex based on local database data. 
  -- This process will take a while given a large DB
  -- While this is running, you can go to `http://localhost:9200/_cat/indices` and refresh to see if indices are being added.

3. Run `rails s` and confirm thru `/findaproject` that elasticsearch is working correctly.

saracarl (changes requested):
Testing "tobacco" on FindAProject, I get this error:
NoMethodError at /findaproject
undefined method `text_entry?' for nil:NilClass
in app/views/elastic/_work_partial.html.slim
on line 2.  
It's probably bad data, but we need to account for it.

saracarl:
We've lost the tabs for the different types of returned objects -- users, collections, works, pages.  This is "brumfield" from the FindAProject page:
image
Suspect we've lost them everywhere; here's a project owner profile page search, which should have collections, works, and pages:
http://localhost:3000/benwbrum/search?term=graves&button=
image
Just checked -- collection search is also missing the tabs.  
See production for examples of how this should look -- tabs should only be present if there is an object of that type returned from the search.
(Check work search's UI, too.)

I did confirm that production and my dev environment are using separate suffixes -- yay!

We seem to have lost es_sync as a task:
saracarl@kateryna:~/dev/products/fromthepage/fromthepage$ bundle exec rake 'fromthepage:es_sync'
rake aborted!
Don't know how to build task 'fromthepage:es_sync' (See the list of available tasks with `rake --tasks`)
Did you mean?  fromthepage:owner_stats

Nothing looks to be the equivalent of sync here:
rake fromthepage:es:data:build                                                                           # Create Elasticsearch indices
rake fromthepage:es:data:query[id]                                                                       # Querries Elasticsearch via ID
rake fromthepage:es:data:reindex                                                                         # Reindex elements
rake fromthepage:es:data:reset                                                                           # Delete Elasticsearch indices
rake fromthepage:es:data:stats                                                                           # Gets Elasticsearch indices data
rake fromthepage:es:setup:cleanup                                                                        # Clean up Elasticsearch components
rake fromthepage:es:setup:init                                                                           # Initialize Elasticsearch configuration

However, a newly saved page gets indexed.  So maybe that's OK?  We should think through this, as it seems to be an architectural change.  (What happens to a page save if ES is slow to respond?)
Confirmed that newly saved collections and works get indexed.

saracarl:
OK.  Today's review.  Searched "brumfield", which worked in the previous iteration, but is failing here:
```
NoMethodError - undefined method `collection' for #<User:0x000055739b224458>
Did you mean?  collections
               collections=:
  app/views/elastic/_search_result.html.slim:19
  app/views/elastic/_tabbed_results.html.slim:54
  app/views/elastic/_tabbed_results.html.slim:53
  app/views/dashboard/landing_page.html.slim:19
  app/controllers/application_controller.rb:67:in `switch_locale'

```
When the first thing returned is a user.

Can we add a space here:
image
between "searching inside" and "Ben W. Brumfield"?

Otherwise, looks good.

Issue #4267, opened 2024-08-08
Handle ADD, INS, DEL, SUP consistently


benwbrum:
We need to make sure that we're using semantic HTML tags correctly for `add`, `ins`, `del`, `sup`, `marginalia`.

### Transcript
```xml
this is just a portion of hte text

this is the rest of the text.

This is a <del>DEL</del> tag.

This is an <ins>INS</ins> tag. (Not supported in our documentation, but autocompletes?)

This is an <add>ADD</add> tag.

This is a <sup>SUP</sup> tag.
```

### Rendering
![Screenshot from 2024-08-08 08-33-56](https://github.com/user-attachments/assets/7c8a8b72-02dd-4d49-89f8-a655b356ba97)

### HTML
(both export and display)
```html
<div class="page-preview ltr" lang="en" dir="ltr">
      <p>this is just a portion of hte text</p>
      
      <p>this is the rest of the text.</p>
      
      <p>This is a <del>DEL</del> tag.</p>
      
      <p>This is an <ins>INS</ins> tag. (Not supported in our documentation, but autocompletes?)</p>
      
      <p>This is an <span class="addition">ADD</span> tag.</p>
      
      <p>This is a <sup>SUP</sup> tag.</p>
    </div>
```

### TEI-XML
```xml
<div xml:id="OTP33267572">
	<fw type="pageNum">FL19466726</fw>
	<p corresp="TTP33267572P0" xml:id="OTP33267572P0">this is just a portion of hte text </p>
	<p corresp="TTP33267572P1" xml:id="OTP33267572P1">this is the rest of the text. </p>
	<p corresp="TTP33267572P2" xml:id="OTP33267572P2">
This is a

		<del>DEL</del>
tag.

	</p>
	<p corresp="TTP33267572P3" xml:id="OTP33267572P3">
This is an

		<ins>INS</ins>
tag. (Not supported in our documentation, but autocompletes?)

	</p>
	<p corresp="TTP33267572P4" xml:id="OTP33267572P4">
This is an

		<add>ADD</add>
tag.

	</p>
	<p corresp="TTP33267572P5" xml:id="OTP33267572P5">
This is a

		<add place="above">SUP</add>
tag.

	</p>
</div>
```

benwbrum:
Marginalia is rendered


### HTML
```html
<p>This is a <span class="marginalia">{MARGINALIA}</span> tag.</p>
```

### TEI
```xml
[...]

PR #4678, opened 2025-06-03
Support bold and ins tags for #4591 #4267

### LINKED ISSUE #4591: Add bold editor button.
We also need to look up what the TEI equivalent to HTML `B` is.

benwbrum:
Closes #4591 and #4267


WillNigel23 (approved):
LGTM. I would ideally add all this logic to xml/lib/utils, but this should work fine as well.

Other than that, would be good to add these cases on special_tags.txt and special_tags.xml as test cases so we are sure these does not cause failed exports.

PR #4683, opened 2025-06-03
3854 - Allow owners to delete subjects with links

WillNigel23:


saracarl:
We're not sure why the un-deleted link ever worked, because it was a capital Churned, but just linked with [[Churned]]:
image

PR #4685, opened 2025-06-05
4174 - PDF import with not text layer, but text import checked, leads to "correction" instead of "transcription"

WillNigel23:


saracarl (changes requested):
This didn't work:
image

benwbrum:
Closed elsewhere.

PR #4691, opened 2025-06-11
4633 - Add featured date to collections and document sets

WillNigel23:


app/controllers/dashboard_controller.rb
```diff
@@ -280,27 +280,19 @@ def landing_page
...
+                                   .featured_projects
+                                   .order(featured_at: :desc)
+                                   .limit(20)
 
     respond_to do |format|
       format.html do
-        @new_projects = (@new_projects + @new_document_sets).sort do |a, b|
+        @new_projects = (new_collections + new_document_sets).sort do |a, b|
           a_date = a.is_a?(Collection) ? a.created_on : a.created_at
           b_date = b.is_a?(Collection) ? b.created_on : b.created_at
```
saracarl, line 297: I think we're going to want to change these to featured_at, rather than created_at.

saracarl (changes requested):
This PR addresses one of the 2 concerns of the issue, in that collections that are private and then turned public get themselves on the New Projects list.  However, we also want a way to "bump" collections onto that list and this doesn't work for that.  If I'm public, then private, then public again (which is how I think we'd "bump"), we don't update the featured date.  (The alternative is updating that date in the console -- ugh -- or building a UI.)

app/controllers/dashboard_controller.rb
WillNigel23: Nice catch!

I rebased. Please delete local branch thanks! :pray:

PR #4709, opened 2025-06-23
4700 create honeypot traps

WillNigel23:


db/migrate/20250625171905_create_honeypot_visits_table.rb
```diff
@@ -0,0 +1,16 @@
+class CreateHoneypotVisitsTable < ActiveRecord::Migration[6.1]
+  def change
+    create_table :honeypot_visits do |t|
+      t.string   :ip_address, null: false
+      t.string   :ip_subnet,  null: false
+      t.string   :browser
+      t.text     :user_agent
+
+      t.index [:ip_subnet, :created_at]
```
benwbrum, line 9: This looks really good.  Let's throw in a foreign key to record `visit_id` so that we can get to the ahoy records if we need to.

benwbrum:
Unfortunately, we ended up not needing this code due to the Cloudflare deployment.

PR #4724, opened 2025-06-25
4702 - ContentDM push retry

WillNigel23:


lib/contentdm_translator.rb
```diff
@@ -136,6 +136,25 @@ def self.fts_field_for_collection(collection)
...
+    delay = 300
+
+    begin
+      ContentdmTranslator.export_work_to_cdm(work, username, password, license)
+    rescue Net::ReadTimeout => e
+      print "Net::ReadTimeout: Retrying in #{delay} seconds... (#{e.message})"
+
+      sleep(delay)
+
+      delay *= 2
```
benwbrum, line 150: This will only do a back-off four times, with a max time-out of 6 minutes.  We're seeing some longer outages lately, so I'd prefer to raise the back-off to somewhere around 4-6 hours.

benwbrum (changes requested):
I've got a couple of minor changes to make, but this looks really good otherwise.  I'm looking forward to testing it!

PR #4730, opened 2025-06-30
4725 - Fix wrong filtering of public collections on owner profile

WillNigel23:


app/models/user.rb
```diff
@@ -387,13 +387,12 @@ def visible_collections_and_document_sets(user)
 
     if user
       collaborator_collections = self.all_owner_collections.where(:restricted => true).joins(:collaborators).where("collection_collaborators.user_id = ?", user.id)
-      owned_collections = self.owned_collections
```
WillNigel23, line 390: This is redundant now. We already have `public_collections = self.unrestricted_collections` and `collaborator_collections = self.all_owner_collections.where(:restricted => true).joins(:collaborators).where("collection_collaborators.user_id = ?", user.id)` which when combined should comprise the set of all unrestricted together with collaborator collections

Issue #4662, opened 2025-05-19
Background long-running subject name updates


benwbrum:
Currently, there are two instance in which `article.title` may be renamed, causing a bulk update of transcription text in pages referencing the article:
* Actually renaming `article.title` via the subject Settings tab, which updates the wikilinks referring to the article to use the new canonical name.
* De-duplicating two articles, which updates the wikilinks within pages referencing the merged article, pointing to the target article's canonical name.
Both of these actions happen within the browser request life-cycle.  The problem is that if thousands of pages link to that article, the request may time out, leaving orphan references to the article.  This happened in the Wilford Woodruff Papers recently, where an article mentioned on 4000 pages was renamed, and only 1500 pages were updated.

We need to move the bulk page update into a background task.

benwbrum:
Was this fixed by #4663 ?

saracarl:
This is affecting Dartmouth, WWP, and CWRGM.  It is not resolved yet.

benwbrum:
We're still running into this in production, even after deploying #4663.  I think that the next step is to pull one of the big collections down to attempt it locally.

benwbrum:
From Lindsey at CWRGM:

> On tags that I have tried to combine and it didn’t work (some are marked to be deleted), do I attempt to combine them again? For example, all references to the Mississippi Circuit Court are supposed to be combined under the [[[Mississippi--Judiciary. Circuit Courts]]](https://fromthepage.com/cwrgm/cwrgm-rev2/article/32118791/edit) tag, which has been annotated and is in two Organizations' child categories. But the following tags did not combine with it properly:
> [Courts & Judicial Systems--Circuit Courts](https://fromthepage.com/cwrgm/cwrgm-rev2/article/32242354/edit)
> [TO_BE_DELETED:Mississippi--Judiciary. Circuit Courts](https://fromthepage.com/cwrgm/cwrgm-rev2/article/32240009/edit)
> [TO_BE_DELETED:Mississippi--Judiciary. Circuit Courts](https://fromthepage.com/cwrgm/cwrgm-rev2/article/70054/edit)
> 
> Do I try combining them again?
> 
> Also, is there a way for me to try and find and delete tags that I have previously deleted but keep popping up again from old documents every few days, like Secretaries or Clerks? I can’t give links to these examples, however, because I can’t see them in FTP right now but I suspect they are still there in documents.

benwbrum:
An update from Jon of the WWP project:

[...]

PR #4735, opened 2025-07-02
4734 - Article rename race condition fix

### LINKED ISSUE #4734: Article Renaming race conditions
Reproduction Steps:
1. From transcribe, create a subject (this will create an article)
2. Edit article name
3. While rename job is running (and without reloading transcribe page), make another change to the source text and click save
  - This will create race condition on the page, which dettaches the page_article_links and creates new article.

Proposed solution is to consider finding article matches based on previous versions created within the hour before deciding if we are dealing with renamed articles or not.

WillNigel23:


app/models/xml_source_processor.rb
```diff
@@ -345,34 +345,55 @@ def update_links_and_xml(xml_string, preview_mode=false, text_type)
     # first clear out the existing links
     # log the count of articles before and after
     clear_links(text_type) unless preview_mode
-    processed = ""
+
+    candidate_articles = collection.articles.left_joins(:article_versions)
```
benwbrum, line 349: I'm not sure why we're joining `articles` to `article_versions` here, since we don't appear to use it below.

app/models/xml_source_processor.rb
```diff
@@ -345,34 +345,55 @@ def update_links_and_xml(xml_string, preview_mode=false, text_type)
...
 
+      article = candidate_articles.find_by(title: title)
+
+      if article.nil?
+        article_version = ArticleVersion.where(article_id: candidate_articles.select(:id))
+                                        .where(title: title)
+                                        .where('created_on > ?', page_update_timestamp)
+                                        .order(created_on: :desc)
+                                        .first
+        article = article_version&.article
```
benwbrum, line 377: I think that we might be able to use the left join above here by doing something like this:
```ruby
candidate_articles.where('article_versions.title': title)
                                 .where(where('article_versions.created_on > ?', page_update_timestamp)
```
etc.  

It's possible that in that case we'd get the correct article, rather than having to pluck it from the article_version, since it would already be in the join.

benwbrum (changes requested):
I really love this approach.  Minor changes requestd.

app/models/xml_source_processor.rb
WillNigel23: Forgot to remove this one.

benwbrum:
I'm not quite sure why this didn't work, but here's what I did:
* Brought up a collection `khs/cwgk-migration` that had thousands of pages and edited `Commonwealth of Kentucky` which had hundreds of pages mention it
* Changed the title to `State of Kentucky`.
* Clicked the Overview tab of the article
* Clicked on the bottom page of in the listing on the right side
* The Page Overview had a mousover (drawn from the `title` attribute of `Commonwealth of Kentucky`, indicating that the XML text was still not updated
* The Tooltip pop-up showed `State of Kentucky`, indicating that the title of the subject article was the new title
* Click the Transcription screen
* See `[[Commonwealth of Kentucky|Ky]]` in the text, indicating that the `source_text` has not been updated.
* Hit Save and got this:
Screenshot from 2025-07-03 18-20-18

benwbrum:
This shows the same behavior as my previous test.  If I change the title of a subject, then save a page that hasn't been updated yet, it still prompts me to categorize a new-looking subject with the old title.

benwbrum:
see also #4662

saracarl:
This is affecting WWP, CWRGM, and Dartmouth.

benwbrum:
I finally was able to get a reproduction video!

https://github.com/user-attachments/assets/eace40b4-4682-4347-95de-8a462392743c

PR #4762, opened 2025-07-16
4760 - Integrate DelayedJob for job monitoring

WillNigel23:
Will be adding this in wiki once PR is merged:

h2. Async job monitoring

FromThePage provides optional asynchronous job adapter `delayed_job` which will store async jobs to db which allow admins to monitor and manage job logs.

To enable, set environment variable `ENABLE_DELAYED_JOB=true`. Once set, running `rails server` should move all async job processes thru delayed_job adapter. This will only enqueue the jobs in DB. In order to start processing them, we need to start a worker. Run `bin/delayed_job run` to start a worker. 

For production environment, every `redeploy` or server restart, we will need to run `bin/delayed_job stop` then `bin/delayed_job start` to restart the workers.

More info about [delayed_job](https://github.com/collectiveidea/delayed_job) here.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -37,8 +37,8 @@ module Fromthepage
     config.upload_host=nil
 
 
-  # load overrides for Thredded and other engines
-  # config/application.rb
+    # load overrides for Thredded and other engines
+    # config/application.rb
     overrides = "#{Rails.root}/app/overrides"
     Rails.autoloaders.main.ignore(overrides)
 
@@ -47,9 +47,14 @@ module Fromthepage
         load override
       end
     end
+
+    if Settings.enable_delayed_job
+      config.active_job.queue_adapter = :delayed_job
+    else
+      config.active_job.queue_adapter = :async
+    end
   end
 
-
-  #uncomment for development of SSO
-  #Rails.application.config.action_controller.forgery_protection_origin_check=false
+  # uncomment for development of SSO
+  # Rails.application.config.action_controller.forgery_protection_origin_check=false
 end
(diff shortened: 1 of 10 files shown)


config/application.rb
```diff
@@ -47,9 +47,14 @@ class Application < Rails::Application
...
+    else
+      config.active_job.queue_adapter = :async
+    end
+  end
 
-  #uncomment for development of SSO
-  #Rails.application.config.action_controller.forgery_protection_origin_check=false
+  # uncomment for development of SSO
+  # Rails.application.config.action_controller.forgery_protection_origin_check=false
 end
```
saracarl, line 60: settings changes are high risk.  Let's comment this back out, please.

config/application.rb
```diff
@@ -47,9 +47,14 @@ class Application < Rails::Application
         load override
       end
     end
-  end
```
saracarl, line 50: settings changes are high risk.  This worries me; need to check the raw file rather than the diff view.

saracarl: This seems ok -- have we been running around with an extra `end` in this file?

saracarl (changes requested):
So this doesn't actually delay any jobs, it just puts the infrastructure in place to delay them in the future?

If enabled, can we create an admin tab "jobs" with the view `delayed_jobs` view loaded within the UI?

config/application.rb
WillNigel23: It is commented. I just added space, rubocop seems to complain when there are no spaces after '#'

WillNigel23:
The gem name is `DelayedJob`, but essentially it just allows `ActiveJob` calls to be:
1. Saved in Database instead of just in-memory
2. Give us dashboard for managing/checking launched async jobs.
  - It also allows us to re-run failed jobs (in case we encounter timeouts/random error)
3. Isolate job logs from server logs.

benwbrum:
Closing based on @WillNigel23's comment about using `solid_queue` instead.

PR #4775, opened 2025-07-23
4774 - ES rake task to catch-up

WillNigel23:


benwbrum:
I'm not sure I really understand the scenario in which we'd run this.  Is this to be done after we've indexed under a new suffix, then switched over, to catch up the records updated between the end of the index job and the switch-over?

PR #4796, opened 2025-07-28
4567 - Fix elasticsearch user query

WillNigel23:


app/interactors/elasticsearch/multi_query.rb
```diff
@@ -332,6 +332,8 @@ def filtered_count
       @type_counts[WorksIndex.index_name.to_sym]
     when 'page'
       @type_counts[PagesIndex.index_name.to_sym]
+    when 'user'
+      @type_counts[UsersIndex.index_name.to_sym]
```
WillNigel23, line 336: I found a bug for when we are scoping by users, the pagination on the page is not scoped to user counts (except uses total_counts). This should fix it!

benwbrum: Nice!

benwbrum:
I still have concerns about deployment, especially since we have such a long backlog of changes to deploy next time.  Would you mind creating a Rails migration that (if ES is enabled) does the UserIndex delete/create/import so that we can be sure it is executed?

PR #4802, opened 2025-07-30
4800 - Subjects elasticsearch architecture

WillNigel23:
# Setup
Inside rails console:
```
ArticlesIndex.create
ArticlesIndex.import Article.includes(:collection, :categories, { works: :document_sets })
```

# Testing
To simplify things I have added `self.es_search` functions inside necessary models that uses es_search so this is not necessarily limited to articles.
Inside rails console:
### Standard query
```
Article.es_search(query: 'Your query string')
```
This will return articles belonging to public collections or belonging to works that belongs to public docset.

### Logged in user query (optional user param)
```
Article.es_search(query: 'Your query string', user: current_user)
```
This will return:
1. Articles belonging to public collections or belonging to works that belongs to public docset.
2. Articles that is owned by current_user even if it is private
3. Articles that belongs to collection that current_user is blocked will not be returned

# Deployment
We can either run the same rails console command as above in setup, OR run reindex rake. I'd rather we run the console version as it will only deal with articles.

### DIFF
--- a/app/chewy_indexes/articles_index.rb
+++ b/app/chewy_indexes/articles_index.rb
@@ -1,0 +1,24 @@
+class ArticlesIndex < Chewy::Index
+  include ChewyConcern
+
+  index_name formatted_index_name('article')
+  default_import_options batch_size: 1000
+
+  index_scope Article
+
+  field :db_id, type: 'integer', value: -> { id }
+  field :collection_id, type: 'integer'
+  field :docset_id, type: 'integer', value: -> { works.flat_map(&:document_sets).pluck(:id) }
+  field :category_ids, type: 'integer', value: -> { categories.map(&:id) }
+  field :owner_user_id, type: 'integer', value: -> { created_by_id }
+  field :is_public, type: 'boolean', value: -> {
+    !collection&.restricted ||
+      works.flat_map(&:document_sets).any? { |doc_set| ['public', 'read_only'].include?(doc_set.visibility) }
+  }
+
+  field :title, type: 'text', analyzer: 'general' do
+    field :no_underscores, type: 'text', analyzer: 'general_no_underscores'
+  end
+
+  field :content_english, type: 'text', analyzer: 'text_english', value: -> { source_text }
+end
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -70,6 +70,54 @@ class Article < ApplicationRecord
   edtf_date_attribute :begun
   edtf_date_attribute :ended
 
+  update_index('articles', if: -> { ELASTIC_ENABLED && !destroyed? }) { self }
+  after_destroy :handle_index_deletion
+
+  def self.es_search(query:, user: nil, is_public: true)
+    blocked_collections = []
+    collection_collabs = []
+    docset_collabs = []
+
+    if user.present?
+      blocked_collections = user.blocked_collections.pluck(:id)
+      collection_collabs  = user.collection_collaborations.pluck(:id)
+      collection_collabs += user.owned_collections.pluck(:id)
+      docset_collabs      = user.document_set_collaborations.pluck(:id)
+    end
+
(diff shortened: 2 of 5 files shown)


app/chewy_indexes/articles_index.rb
```diff
@@ -0,0 +1,21 @@
...
+  index_name formatted_index_name('article')
+  default_import_options batch_size: 1000
+
+  index_scope Article
+
+  field :db_id, type: 'integer', value: -> { id }
+  field :collection_id, type: 'integer'
+  field :category_ids, type: 'integer', value: -> { categories.map(&:id) }
+  field :owner_user_id, type: 'integer', value: -> { created_by_id }
+  field :is_public, type: 'boolean', value: -> {
```
benwbrum, line 14: We are starting to see a handful of customers using article text very extensively.  I think that we should support the same kind of full-text search for articles that we do for pages.

app/chewy_indexes/articles_index.rb
```diff
@@ -0,0 +1,21 @@
...
+    !collection&.restricted ||
+      works.flat_map(&:document_sets).any? { |doc_set| ['public', 'read_only'].include?(doc_set.visibility) }
```
benwbrum, line 16: This is pretty elegant.  What will happen when works are moved from one document set to another, changing the theoretical visibility of the article?

app/models/article.rb
```diff
@@ -71,6 +71,56 @@ class Article < ApplicationRecord
...
+          bool: {
+            must_not: [
+              { terms: { collection_id: blocked_collections } }
+            ],
+            # At least one of the following must be true
+            should: [
+              { term: { is_public: is_public } },
+              { term: { owner_user_id: user&.id || -1 } },
+              { terms: { collection_id: collection_collabs } },
+              { terms: { docset_id: docset_collabs } }
```
benwbrum, line 112: I see this in the terms, which parallels what we do with works and pages, but I don't see `docset_id` in the chewy index.  Is this dead code?  It'd be nice to be able to search for articles that are only mentioned in private collections and private docsets, if I'm able to see the docset.

(Updating the docsets for each article as works are moved from one docset to another might be a headache, however.)

app/chewy_indexes/articles_index.rb
WillNigel23: As long as one public collection/docset exists then the article is considered public.

app/models/article.rb
WillNigel23: We can add this later on. I think I've moved most of the logic this is happening to an interactor already so it should be easier to change. Check this: https://github.com/benwbrum/fromthepage/blob/f961b1e08d128afb4506ec9ae5f56ac21c507a55/app/jobs/elasticsearch/collection/sync_job.rb

It should sync when collection is updated. I think I call this both in collectiion and document sets already. So all that's left is for adding this logic in adding/removing works from docset.

WillNigel23: I think docset-id is defined in pages/works index.

saracarl:
As part of this PR, can we look at how subjects are ranked in the index/search results?  They should be weighted in such a way that they show up after collections but before works (if the search term is in the subject title).  If the search term is in the article, it should show up below works with the search term in their title or description, but above pages with the search term in their page text.

WillNigel23:
> As part of this PR, can we look at how subjects are ranked in the index/search results? They should be weighted in such a way that they show up after collections but before works (if the search term is in the subject title). If the search term is in the article, it should show up below works with the search term in their title or description, but above pages with the search term in their page text.

Upon further review, this should be done in [#4997](https://github.com/benwbrum/fromthepage/issues/4997), or on another sub-PR integrating this base subject architecture to the multiquery implementation.

@saracarl

WillNigel23:
@benwbrum this was rebased

PR #4819, opened 2025-08-05
4757 - Uploaded metadata should update additional metadata attributes

WillNigel23:


app/interactors/work/metadata/import_csv.rb
```diff
@@ -15,6 +15,22 @@ class Work::Metadata::ImportCsv < ApplicationInteractor
     '*Uploaded Filename*'
   ].freeze
 
+  ADDITIONAL_METADATA_HEADERS = {
+    author: 'Author',
+    recipient: 'Recipient',
+    location_of_composition: 'Place of Creation',
+    document_date: 'Document Date',
```
benwbrum: Document date has a validation rule that constrains it to the EDTF format.  Our users love the idea of EDTF but sometimes struggle with getting it correct.  What happens when a user uploads a spreadsheet that has an `document_date` value that fails EDTF validation?  I suspect that the work will not be updated -- in this case, it would be acceptable to place the date in `original_metadata` and leave the `document_date` value blank, but there might be better ways to handle this.  Regardless, I'd like to see a test case for this as well.

benwbrum (changes requested):

benwbrum:
I have tested this and believe that it is ready to merge as soon as `development` is unlocked.

PR #4827, opened 2025-08-06
4826 - Specs improvement PR

WillNigel23:


app/models/work.rb
```diff
@@ -609,6 +609,12 @@ def user_can_transcribe?(user)
       self.scribes.include?(user)
   end
 
+  def is_public?
+    return true if collection.nil? || !collection.restricted?
```
benwbrum, line 613: I'm curious about the style on this.  Why use this format rather than a three-way `||` operation?  (Not a criticism, just wondering.)

WillNigel23: I usually do early returns in functions. This is just one of them. I find it clearer to do this than multiple nesting if then statements.

That being said this one can be written just as you said, three conditions. I believe both are easy to read given that this is very simple check. But on larger blocks of logic it is way easier for me to read when there is early returns.

benwbrum: That's fair -- I was just curious.

app/models/collection.rb
```diff
@@ -253,56 +253,11 @@ def disable_messageboards
     self.save!
   end
 
-
-  def use_as_template(new_title)
```
benwbrum, line 257: Hey, where did this code go?

WillNigel23: #3631
as discussed this was unfinished implementation. I suggest we remove it here and move it someplace else like in libs when we get the change to finish the implementation

benwbrum:
This is ready to merge

PR #4833, opened 2025-08-12
4832 - Integrate solid_queue

### LINKED ISSUE #4832: Integrate solid queue

WillNigel23:
# Rationale

This PR integrates 2 gems into our pipeline:

1. [solid_queue](https://github.com/rails/solid_queue)
- This comes with modern rails versions. It would be good to use this as our job queuing engine as it is deeply integrated with rails
2. [mission_control-jobs](https://github.com/rails/mission_control-jobs)
- This provides web-based dashboard for solid_queue jobs. Also deeply integrated with rails.

The reason for adopting these are as follows:
1. Currently we queue our jobs in-memory. Meaning if for some reason the server restarts, we lose all unfinished enqueued jobs.
2. Better logging and job management. Currently we have async jobs that are misbehaving ([race conditions, failures, etc](https://github.com/benwbrum/fromthepage/issues/4734)).  It is very hard to determine the reason for these bugs. It will help if we can keep track of all jobs that finished/failed. This allows for that.
3. solid_queue provides more features that would be useful to add to our pipeline (cronjobs can be moved here, failed jobs can be retried, adding multiple job workers for faster processing, etc) 

# Setup
1. This PR includes db migrations to add solid_queue-related columns in DB. Right now it is added in the current database table (`diary_#{RailsEnv`). It is possible to move it to its own DB, I'll let @benwbrum decide if we want that overhead or not.
2. Once migrations are finished, we need to run `bin/jobs` to start a worker. This will listen on db to process enqueued jobs. Like in delayed_jobs prototype before, we should probably add step in deployment to start/run a worker service. 
Probably by running a nohup instance of `bin/jobs` to keep things really simple. Tagging @saracarl , we may need to play around with this as I have not tested this in a production environment.
3. We have `logs/queue.log` for worker logs, but job logs like rails stacktraces will be found in mission_control web-UI dashboards.
4. Access the web-UI panel at `/solid_queue`. This will require an ftp-admin account.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -7,7 +7,7 @@ class AdminController < ApplicationController
 
   def authorized?
     unless user_signed_in? && current_user.admin
-      redirect_to dashboard_path
+      redirect_to main_app.dashboard_path
     end
   end
 
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,8 @@ gem 'acts_as_tree'
 gem 'interactor-rails', '~> 2.0'
 gem 'friendly_id'
 gem 'thredded', '~> 1.1'
+gem 'solid_queue', '~> 1.2'
+gem 'mission_control-jobs'
 
 # Authentication / Security
 gem 'devise'
@@ -49,6 +51,7 @@ gem 'roo'
 gem 'charlock_holmes'
 gem 'nkf'
 gem 'net-pop', github: 'ruby/net-pop'
+
 gem 'forty_facets'
 gem 'chewy'
 
@@ -144,3 +147,4 @@ group :test do
   gem 'webmock', require: false
   gem 'with_model'
 end
+
--- a/Gemfile
+++ b/Gemfile
@@ -147,4 +147,3 @@ group :test do
   gem 'webmock', require: false
   gem 'with_model'
 end
-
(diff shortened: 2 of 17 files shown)


app/controllers/admin_controller.rb
```diff
@@ -7,7 +7,7 @@ class AdminController < ApplicationController
 
   def authorized?
     unless user_signed_in? && current_user.admin
-      redirect_to dashboard_path
+      redirect_to main_app.dashboard_path
```
benwbrum, line 10: Good catch!

benwbrum (approved):
We still need to test this locally, but the code looks good.

benwbrum:
I've done some limited testing, and this looks really good.  Before we merge it into development, I'd like a link to it from the Admin dashboard or tabs somewhere.  It would be nice to give it a bit of the look-and-feel of the rest of the app, but since it's on the admin side we don't care that much.

WillNigel23:
Steps to do in production:
1. Run a worker.
  - We can do this via `bin/jobs`. Preferably this is ran via `nohup` so it would persist on closed terminals.
  - `nohup bin/jobs >> log/jobs.log 2>&1 &` 
2. Add these two lines in production.rb:
```
  config.active_job.queue_adapter = :solid_queue
  config.solid_queue.logger = ActiveSupport::Logger.new(Rails.root.join('log/queue.log'))
```

PR #4860, opened 2025-08-25
4859 - Refactor needs review logic

WillNigel23:


app/interactors/transcribe/lib/needs_review_handler.rb
```diff
@@ -0,0 +1,75 @@
+class Transcribe::Lib::NeedsReviewHandler < Transcribe::Lib::BaseHandler
```
WillNigel23, line 1: Normally we would like to only add tests for `interactor` tests rather than handlers. But this will be temporary just to assert the logic to be correct.

We want to create something like Transcribe::SavePage where all of the handlers are called and tested instead of this pattern.

spec/interactors/transcribe/lib/needs_review_handler_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      it 'does not change status to needs_review' do
+        expect { result }.not_to(change { Deed.count })
+        expect(result.status_needs_review?).to be_falsey
+      end
+
+      context 'when page is already needs_review' do
+        let!(:page) { create(:page, work: work, status: :needs_review) }
+
+        it 'changes status to new' do
```
saracarl, line 71: @WillNigel23 can you explain this?  It looks potentially off to us.

spec/interactors/transcribe/lib/needs_review_handler_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      it 'does not change translation status to needs_review' do
+        expect { result }.not_to(change { Deed.count })
+        expect(result.status_needs_review?).to be_falsey
+      end
+
+      context 'when page is already needs_review' do
+        let!(:page) { create(:page, work: work, translation_status: :needs_review) }
+
+        it 'changes translation status to new' do
```
saracarl, line 118: double check this when previous comment is resolved.

saracarl (commented):
This looks really good, but we have some questions.

spec/interactors/transcribe/lib/needs_review_handler_spec.rb
WillNigel23: This is the `toggle` action. When we click on the `saves_to_need_review` button and the page is already in review, it will remove the needs_review status.

I did not change the corresponding legacy spec and I believe this was tested there as well.

saracarl: So the existing legacy spec doesn't test the "transcribed -> needs review --> edit & save again" flow that I see:
https://github.com/benwbrum/fromthepage/blob/53d26b8e5cff9edcc6950ac6136ee2fbe7626912/spec/features/needs_review_spec.rb#L298-L304

If a page is in "needs_review" and it is edited & saved, it should have the following 2 states:
-- page status is transcribed (not new as this test has)
-- the page continues to be needs_review

If a page is in "needs_review" and is edited (or not) and approved (reviewed? -- approve button clicked), it should have the following 2 states:
-- page status is "transcribed" or "indexed"
-- page is no longer needs_review

The only thing that changes the needs_review status is clicking the "approved" button (in the required review scenario) or unclicking the "needs review" checkbox (in the optional review scenario).

WillNigel23: This one's old PR and I lost memory for my reasoning of things here, but after a quick review I think this is just correct.

In `needs_review_handler_spec`
If you check `line 61` of the spec, we set `needs_review = '0'`
`Line 63`  tests what happens if we `uncheck` the `needs_review param` BUT the page is NOT `status_needs_review` , which means NOTHING changes.

`Line 68` testswhat happens if we `uncheck` the `needs_reivew param` AND the page IS `status_needs_review`, it correctly resets the status back to `new`

The refactor is only handling the `needs_review` branch previously in the controller. Therefore the final page status is actually changed when used in the controller logic (thus all tests passed because we retained behavior)

Then try checking the diff (the removed one) in the `transcribe_controller`
```
  def needs_review
    if params[:type] == 'translation'
      if @page.work.collection.review_workflow == true && @page.translation_status_new?
        @page.translation_status = :needs_review
        record_deed(DeedType::TRANSLATION_REVIEW)
      elsif params[:page]['needs_review'] == '1'
        unless @page.translation_status_needs_review?
          @page.translation_status = :needs_review
          record_deed(DeedType::TRANSLATION_REVIEW)
        end
      else
        if @page.translation_status_needs_review?
          @page.translation_status = :new
          record_deed(DeedType::TRANSLATION_REVIEWED)
        end
        nil
      end
    elsif params['save_to_needs_review'] && @page.work.collection.review_workflow
      unless @page.status_needs_review?
        # don't log a deed if the page was already in needs review
        @page.status = :needs_review
        record_deed(DeedType::NEEDS_REVIEW)
      end
    else
      if params[:page]['needs_review'] == '1'
        unless @page.status_needs_review?
          @page.status = :needs_review
          record_deed(DeedType::NEEDS_REVIEW)
        end
        # if @page.translation_status == 'blank'
        #  @page.translation_status = nil
        # end
      else
        if @page.status_needs_review?
          @page.status = :new
          record_deed(DeedType::PAGE_REVIEWED)
          nil
        end
      end
    end
  end
```
Essentially we are testing this branch of the past behavior:
```
      else
        if @page.translation_status_needs_review?
          @page.translation_status = :new
          record_deed(DeedType::TRANSLATION_REVIEWED)
        end
        nil
      end
```

Moreover, we actually use the handler in `Line 86` of transcribe controller. If you follow the code, you will see that the proper setting of status is then handled by the logic after it. 

So this is just correct, we merely isolated the logic happening in the `def needs_review`

app/interactors/transcribe/lib/needs_review_handler.rb
```diff
@@ -0,0 +1,75 @@
...
+        @page.status = :needs_review
+        record_review_deed(DeedType::NEEDS_REVIEW)
+      end
+    elsif needs_review? && !@page.status_needs_review?
+      # Needs review checked and status is not needs review
+      @page.status = :needs_review
+      record_review_deed(DeedType::NEEDS_REVIEW)
+    elsif !needs_review? && @page.status_needs_review?
+      # Needs review unchecked and status is needs review
+      @page.status = :new
```
saracarl, line 61: This is probably wrong, too.

app/controllers/transcribe_controller.rb
```diff
@@ -53,48 +53,6 @@ def mark_page_blank
...
-        unless @page.status_needs_review?
-          @page.status = :needs_review
-          record_deed(DeedType::NEEDS_REVIEW)
-        end
-        # if @page.translation_status == 'blank'
-        #  @page.translation_status = nil
-        # end
-      else
-        if @page.status_needs_review?
-          @page.status = :new
```
saracarl, line 90: This doesn't make any sense.

saracarl (commented):
@benwbrum to review this.  I don't see why we'd ever set a page status to new in any review workflow.

PR #4879, opened 2025-09-04
4872 - Integrate active storage for document uploads

### LINKED ISSUE #4872: Integrate active_storage for document_uploads

WillNigel23:


app/models/document_upload.rb
```diff
@@ -22,8 +22,24 @@ class DocumentUpload < ApplicationRecord
...
 
-  validates :collection_id, :file, presence: true
+  validates :collection_id, presence: true
 
+  ACCEPTED_FILE_TYPES = [
+    'application/pdf',
+    'application/zip'
+  ].freeze
+
+  FE_ACCEPTED_FILE_TYPES = [
```
benwbrum, line 32: I don't see this used anywhere -- is it dead code?

app/models/document_upload.rb
```diff
@@ -22,8 +22,24 @@ class DocumentUpload < ApplicationRecord
...
+    'application/x-zip',
+    'application/x-zip-compressed'
+  ].freeze
+
+  validates :attachment, attached: true, on: :create
+  validates :attachment, content_type: ACCEPTED_FILE_TYPES, on: :create
+
+  has_one_attached :attachment
+
+  # TODO: We will soon deprecate DocumentUploader in favor of ActiveStorage
```
benwbrum, line 42: I assume that this is in the other PR?

config/importmap.rb
```diff
@@ -10,5 +10,5 @@
 pin_all_from 'app/javascript/plugins', under: 'plugins'
 pin_all_from 'app/javascript/controllers', under: 'controllers'
 pin_all_from 'app/javascript/actions', under: 'actions'
-pin_all_from 'app/javascript/vendor', under: 'vendor', preload: false
```
benwbrum, line 13: I'm curious what this did?

benwbrum (commented):
With the exception of the possible dead code constant, this looks fine to me.

app/models/document_upload.rb
WillNigel23: This is used in the following PR #4884. I remember having to split the ticket since it was huge, this must be an artifact of that.

Either way it will be used in the following PR so let's keep it here.

app/models/document_upload.rb
WillNigel23: Not yet. https://github.com/benwbrum/fromthepage/pull/4685 is the initial PR, but then we decided to split it into smaller pieces so it is easier to review and release. So this first, then 4884, then eventually the third (and probably 4th if necessary)

So to outline, what we did is:

1. 4872 - Add support for active storage on DocumentUpload
2. 4884 - Add support for direct upload (to linode in our case)
3. (No ticket yet) - Move the ingestor rake to solid_queue (which is in part done here: https://github.com/benwbrum/fromthepage/pull/4685)
4. (No ticket yet, and not in the closed PR mentioned in 3) Fully deprecate carrierwave `DocumentUploader`

config/importmap.rb
WillNigel23: Just some optimization. preloading will tell the browser to fetch and load the js dependencies under vendor before page load.

PR #4896, opened 2025-09-11
4880 - Latex export refactor

### LINKED ISSUE #4880: Latex export refactor (tables duplicated in exports)
ADAH is reporting that exporting some of their Supreme Court documents produces duplicate tables.  

Here's a page in FromThePage (must log in as ADAH to see it):
https://fromthepage.com/adah/alabama-supreme-court-case-files-complete/hunley-et-al-v-hunley/display/33042904

which looks fine in the UI:

Image

In our plaintext export (which is what is sent to CDM, where ADAH sees the problem) here:
https://fromthepage.com/export/work_plaintext_verbatim?work_id=32053043
(search "Apraise Bill and Inventory of estate Richard  Hunley Decd")
you can see the duplicate tables:

Image

I suspect this might be caused by having 2 tables on the page -- that we append the 2nd table to the end of the first and then also present the 2nd table.

(Also affects the Word export, etc -- anything downstream.)

WillNigel23:


WillNigel23:
This is an incomplete prototype. I am only creating a PR so you can test it with me along the way.

There are many unhandled features yet (ie but not limited to preserve_lb, flatten_links, etc)

But the general idea is:
1. We already have a valid xml format for source text. It is reliably correct when viewing as html.
2. Since it is reliable already, we can ditch the intermediate step of converting to `md` format. This is where most of our pain points occur.
3. Convert it directly to `tex` format. This gives us FULL control on how we want to format it as a document. We can directly remediate if there will be incompatible tags (ie nested effects which are breaking when we go from md to pdf)

How to test:
1. The steps are a bit janky as it is incomplete so please bear with me. Go to collection export page:
image
2. Click on `Download` on pdf export.
3. You will be greeted by a plain text page instead of a downloadable file. Copy the contents
image
4. Paste it into a tex editor. I suggest overleaf.
image

In a few hours of development and testing I was able to get to this point. Fully functional table, albeit slightly different than our usual formatting. We can dial in the look later, the most important thing for me now is that we are able to generate a decent looking document from it already.
image

saracarl:
This looks really interesting, and the example table above looks better than we what we're generating now.  I think anything that makes fixing and formatting all of our esoterica text better is a great idea.  

How would I test this for Word exports?

WillNigel23:
@saracarl Well for now, everything is converted directly to latex. The goal is we still use pandoc but to covnert from latex to docs/pdf

benwbrum:
I'm really excited about this approach.  So far both the approach documented and the code itself look great!

app/interactors/work/export/lib/utils.rb
```diff
@@ -0,0 +1,238 @@
+class Work::Export::Lib::Utils
+  LINEBREAK_ELEMENT = "\\\\\n"
```
benwbrum, line 2: This might be more readable in single quotes, since you wouldn't have to double-escape some of the backslashes.

app/interactors/work/export/lib/utils.rb
```diff
@@ -0,0 +1,238 @@
...
+    when 'table'
+      process_table(element, preserve_lb, flatten_links)
+    when 'tr', 'row'
+      element.elements.to_a('th|td').map { |td| process_element(td, preserve_lb, flatten_links) }.join(' & ') + " \\\\\n"
+    when 'td'
+      content
+    when 'link'
+      process_links(element, content, flatten_links)
+    when 'add', 'ins'
+      "\\textsuperscript{\\underline{#{content}}}"
```
benwbrum, line 79: Does this actually work?  If so, it's very exciting!

app/interactors/work/export/lib/utils.rb
```diff
@@ -0,0 +1,238 @@
...
+      # TODO: PDF and DOCX. Add lua filter to handle HTML to have tooltips
+      "\\underline{\\textit{#{content}}}\\{#{abbr}\\}"
+    when 'reg'
+      orig = element.attributes['orig']
+
+      # TODO: PDF and DOCX. Add lua filter to handle HTML to have tooltips
+      "\\underline{\\textit{#{content}}}\\{#{orig}\\}"
+    when 'footnote'
+      "\\footnote{#{content}}"
+    when 'head'
```
benwbrum, line 97: Are there any better stylings available to us in LaTeX for section headers?  I realize that this might be what we did before, but it really could be improved.

benwbrum: Actually, what's the difference between `head` and `entryHeading`?

benwbrum (commented):
I love this rewrite -- it's so clean!

app/interactors/work/export/lib/utils.rb
WillNigel23: I believe '\n' is different from "\n" in rails. It appears different at least when in editor. 

But essentially this aims to add 

```
\\

```

in the produced latex.

app/interactors/work/export/lib/utils.rb
WillNigel23: It does based on my testing so far

app/interactors/work/export/lib/utils.rb
WillNigel23: It is very limited as far as I can tell, I moved back to using pandoc again so it is more manageable for us, and that decision in itself limits our style options

WillNigel23: I just matched what I am seeing in the transcribe previewer. And entryHeading and are slightly different

saracarl:
OK.  We tested this with the following page transcription -- a pretty complicated one -- and then exported the attached PDF. 

```
8

should run smoothly. The dogs are
picking up and should be in shape
soon. One walrus and numerous ducks 
seen. When in the dory, I wounded one
duck with the shotgun but he got away
Light westerly breeze. Clear + getting
colder

<head><date when="1921-09-26">Sept 26 Monday</date></head>

<add>Max 32.5
Min 18.0</add>

All hands busy all day west of the camp
cutting and splitting logs for the sides
and roof of the house. It is surely a
great sight to see all the fine dry wood
after being on the islands north of [[Canada]]
when drift wood in scarce. Our [[Blackjack, Ada, 1898-1983|seamstress]]
is busy making clothing and we had a 
great many clothing skins so our clothing
outfit should be comfortable. A large
number of ducks and gulls seen. Cold
Clear and blowing a fresh breeze from
the North

<head><date when="1921-09-27">Sept 27 Tuesday</date></head>

<add>Max 30
Min 17.5</add>

A howling wind from the East accomp-
nied by a heavy fall of wet snow kept us
close to the tent today. Nothing done but 
necessary chores.

<head><date when="1921-09-28">Sept 28 Wednesday</date></head>

<add>Max 26.8
Min 19.3</add>

Spent most of the day working on the 
house. A small amount of ice hove in sight
```

[1_12_1943_waac_20250924154005.pdf](https://github.com/user-attachments/files/22518303/1_12_1943_waac_20250924154005.pdf)


Here are the problems we noticed:
1)  the downloaded file doesn't have a .pdf suffix
2)  Word wrap is not handled correctly; there are no spaces where line breaks were in the original
image
3)  We need more spacing between the header and the preceeding paragraph:
image
4)  Compared with the version in development, we really like the bold typeface on head, but think we should skip the underline.

WillNigel23:
Nice catch on the missing spaces on linebreaks. I certainly did not consider that!

For the underline typeface on `head` I just matched it with it's HTML counterpart. Given that I am trying to make everything go thru this pandoc formatter (yes, including transcribe overview page), it would be good if we can decide on a consistent formatting for that. But in the meantime it is easy enough to just use `bold` for now as we are not using this on HTML yet.

I'll address the other issues.

saracarl:
It makes sense to match the PDF formatting and the HTML formatting.  I think bold probably works, at least for now.

WillNigel23:
Had to rebase @saracarl @benwbrum

WillNigel23:
Please refer to dockerfile for dependencies needed on server. As far as I can tell, as long as `texlive-full` is installed we should be covered in prod env. But just to be safe, these are must-haves

```
      texlive-base \
      texlive-latex-base \
      texlive-latex-recommended \
      texlive-fonts-recommended \
      texlive-latex-extra \
      texlive-xetex \
      texlive-luatex \
      pdf2svg \
      fonts-cmu \
 ```

**Comment by WillNigel23** [comment]
# Simplify Export Pipeline

## Summary  
This PR simplifies our export pipeline by removing unnecessary intermediate steps and making the process more testable, reusable, and maintainable.  

Previously, the pipeline was:

`xml -> markdown -> latex -> pdf/docx`

With this change, the pipeline becomes:  

`xml -> latex -> pdf/docx`

---

## Motivation & Improvements  

### Why remove Markdown?  
- **Limited formatting support**  
  - Custom XML tags didn’t translate well into Markdown.  
  - Workarounds required heavy regex cleanups (Regular Expressions: Now You Have Two Problems :trollface: ).  
  - These cleanups often broke downstream LaTeX generation.  

- **Direct LaTeX is simpler**  
  - Since we convert to LaTeX anyway, skipping Markdown prevents double translation losses.  
  - We now handle custom tags directly in LaTeX.  
  - Pandoc limitations (e.g. page breaks, missing fonts) are handled via Lua filters.  

- **Better testability & reusability**  
  - The previous implementation relied on syscalls and was hard to test in isolation.  
  - We introduced a dedicated interactor with single responsibility.  
  - This makes the pipeline easier to reason about, test, and extend.  

### Dependency improvements  
- Reduced dependency bloat in CI (from ~8GB to ~2GB).  
- Instead of `texlive-full`, we identified the minimal required set:  
```
texlive-base
texlive-latex-base
texlive-latex-recommended
texlive-fonts-recommended
texlive-latex-extra
texlive-xetex
texlive-luatex
pdf2svg
fonts-cmu
```

This makes the setup lighter and easier to maintain.  

---

## What This PR Does  
1. Removes the Markdown step from the export pipeline.  
2. Implements a single-responsibility interactor for better architecture and testability.  
3. Defines the minimal LaTeX dependencies required for PDF/docx exports.  

---

## Next Steps / TODO  
1. **Bulk export support** – extend this new pipeline to bulk exports and deprecate the old one.  
2. **Bug testing** – tested with known problem transcriptions, but more edge cases may surface.  
3. **HTML export unification** – currently still uses `xml -> html`. Plan is to hand this off to Pandoc as well, so all formats share the same consistent pipeline. This will likely require a few more Lua filters.

**Comment by saracarl** [comment]
I'm testing this locally against a copy of the Jeremiah White Graves diaries and I'm getting this:

```
2026-01-29T14:07:16-06:00 Exporting work	14646	Diary of Jeremiah White Graves, Volume 2, Book 9
rake aborted!
ActionView::Template::Error: Malformed XML: No root element (ActionView::Template::Error)
Line: 1
Position: 21
Last 80 unconsumed characters:
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/baseparser.rb:271:in `pull_event'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/baseparser.rb:249:in `pull'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/parsers/treeparser.rb:21:in `parse'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/document.rb:468:in `build'
/home/saracarl/.rvm/gems/ruby-3.3.5/gems/rexml-3.4.4/lib/rexml/document.rb:105:in `initialize'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/abstract_xml_helper.rb:168:in `new'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/abstract_xml_helper.rb:168:in `xml_to_html'
/home/saracarl/dev/products/fromthepage/fromthepage/app/helpers/export_helper.rb:43:in `xml_to_pandoc_md'
/home/saracarl/dev/products/fromthepage/fromthepage/app/views/export/facing_edition.html.erb:34:in `block in _app_views_export_facing_edition_html_erb___1042448849166234167_222720'
```

PR #4915, opened 2025-09-22
4914 - CDM import does not reindex properly

### LINKED ISSUE #4914: Search issues on newly uploaded works (CWRGM)
Lindsey reports:

> I have a few more examples of this issue. This morning I uploaded new scans of existing documents to the public CWRGM set, and the previous copies are in the In Omeka set. They both show up under the Sets page but not from the search bar. When using the search bar, the In Omeka set documents seem to come up, but the newly uploaded documents do not. Here are the identifier numbers for a few I tested:
> mdah_757-932-09-05
> mdah_757-942-03-13
> mdah_757-942-07-01
>

WillNigel23:


app/controllers/sc_collections_controller.rb
```diff
@@ -257,6 +257,15 @@ def convert_manifest
       ocr_text = ocr ? 'and OCR text ' : ''
       flash[:notice] = t('.metadata_is_being_imported', ocr_text: ocr_text)
     end
+
+    # TODO: This patches CDM imports. It would be good to refactor it into interactor to make testing easier
+    Elasticsearch::Collection::SyncJob.perform_later(
```
benwbrum: How will the `perform_later` interact with the rake task?  Do we run into a possibility of race conditions, where the rake task is still running after the SyncJob completes?

WillNigel23: Nice catch! Fixing

benwbrum (approved):
This looks really good.

PR #4940, opened 2025-09-29
4939 - SolidQueue Worker Service

### LINKED ISSUE #4939: SolidQueue worker service

WillNigel23:
This is to ensure we start a worker every time we redeploy.

This is an outline `solid_queue.service` config file, I'll let Ander Kiereg check this one:

```
[Unit]
Description=Solid Queue worker for FromThePage
After=network.target

[Service]
Type=simple
EnvironmentFile=#path to env files
WorkingDirectory=/home/fromthepage/deployment/current
ExecStart=/bin/bash -lc 'bundle exec rails solid_queue:start'
User=fromthepage
Restart=always
RestartSec=5
StandardOutput=#path to solid_queue logs. This probably should be a linked file instead of actual file to `current` dir as we are `current` on every deploy.
StandardError=syslog
SyslogIdentifier=solid_queue

[Install]
WantedBy=multi-user.target
```

@saracarl In the meantime I started a solid_queue worker in production. We can find the worker's PID at fromthepage.com/solid_queue at `Workers` tab. We can kill it once this one is resolved.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -47,7 +47,16 @@ namespace :deploy do
 
   after :publishing, :restart
 
-  after :restart, :clear_cache do
+  after :restart, :restart_solid_queue
+
+  desc 'Restart Solid Queue service'
+  task :restart_solid_queue do
+    on roles(:app) do
+      execute :sudo, :systemctl, 'restart solid_queue'
+    end
+  end
+
+  after :restart_solid_queue, :clear_cache do
     on roles(:web), in: :groups, limit: 3, wait: 10 do
       # Here we can do anything such as:
       # within release_path do


saracarl:
@akierig please review...

akierig:
```
[Unit]
Description=Solid Queue worker for FromThePage
After=network.target

[Service]
Type=simple
WorkingDirectory=/home/fromthepage/deployment/current
Environment="PATH=/home/fromthepage/deployment/current/bin:$PATH"
ExecStart=/bin/bash -lc 'bundle exec rails solid_queue:start'
User=fromthepage
Restart=always
RestartSec=5
StandardOutput=file:/var/log/solid_queue/solid_queue.log
# would it be preferable to use, e.g., StandardError=file:/var/log/solid_queue/solid_queue_error.log ?
StandardError=syslog
SyslogIdentifier=solid_queue

[Install]
WantedBy=multi-user.target
```

edit: @WillNigel23 was correct, we need to set the $PATH

config/deploy.rb
```diff
@@ -47,7 +47,16 @@
 
   after :publishing, :restart
 
-  after :restart, :clear_cache do
+  after :restart, :restart_solid_queue
+
+  desc 'Restart Solid Queue service'
+  task :restart_solid_queue do
+    on roles(:app) do
+      execute :sudo, :systemctl, 'restart solid_queue'
```
akierig, line 55: I can't see if the fromthepage [as the user doing the deploy] has sudo to do `systemctl restart solid_queue` but we can grant it that very limited permission with a file in `/etc/sudoers.d/` that grants `/usr/bin/systemctl restart solid_queue`

saracarl:
@akierig Can you create this service on prod?  If we use this to restart, will it kill the one we started on the command line?  Guessing not.  @WillNigel23 and I just figured out that if we have 2 workers running we run into problems, so we'll want to kill the old one if this doesn't do it.  We can get the PID here:  https://fromthepage.com/solid_queue/applications/fromthepage/workers?server_id=solid_queue

Then we can check if fromthepage needs sudo or not and make that change if we need to.

I want to double check that the restart successfully kills the old worker when both are started via the service.

After we do those two things, we can merge this in and deploy because both pieces will be there for the capistrano change.

WillNigel23:
I think it will kill the running worker.

The difference with running it as service vs running it as nohup is, when we run as nohup we would have to manually kill it via it's process_id. Service will daemonize the task itself, and restarting it should kill the process.

`bundle exec solid_queue:start` in itself is a synchronous process, it does not detach from the caller.

akierig:
just so I have it down somewhere, this is what we'd need in `/etc/sudoers.d/fromthepage`

```
fromthepage ALL=NOPASSWD: /usr/bin/systemctl restart solid_queue
```

saracarl:
@benwbrum and I looked at this again, and we have concerns with where this is logging.  

```
StandardOutput=file:/var/log/solid_queue/solid_queue.log
# would it be preferable to use, e.g., StandardError=file:/var/log/solid_queue/solid_queue_error.log ?
StandardError=syslog
```

Is this just the service log (i.e. restart logging?) or is it the actual queue logging?  Right now, running solidqueue by hand, we get a queue.log in `deployment/shared/log` .  Jury is still out, but it seems kind of useful.  Would convention dictate it goes with the app logs in `deployment/shared/log` or with the system logs in `/var/log/solid_queue/`?  My gut says that 6 months from now, when we're troubleshooting a queue problem, we won't remember that there are solid_queue logs in /var/log but will notice them in the app log directory.

(Note to self -- we'll add this as /etc/systemd/system/solid_queue.service once we've resolved the above.)

akierig:
> @benwbrum and I looked at this again, and we have concerns with where this is logging.
> 
> Is this just the service log (i.e. restart logging?) or is it the actual queue logging? Right now, running solidqueue by hand, we get a queue.log in `deployment/shared/log` . Jury is still out, but it seems kind of useful. 

then we can just yank that part. Systemd related errors will run to `journald` and we'll see if it keeps logging where it is now. If it doesn't, we can try setting those to where `solid_queue` logs now.

saracarl (approved):

PR #4944, opened 2025-09-30
4943 - ActiveRecord::Deadlocked: Mysql2::Error on rename_job

### LINKED ISSUE #4943: ActiveRecord::Deadlocked: Mysql2::Error on rename_job
Image

The  decision to migrate to solid_queue paid off for chasing the unexplainable article_rename_job.

It seems, depending on timing, there is a chance to hit a deadlock on page updates. After `retrying` failed jobs, it finished successfully.

In my opinion this is an unavoidable consequence of backgrounding the rename process without `locking` the relevant records. That being said, I think `retry` approach will solve the problem.

We catch the deadlock error, then we retry up to n times. 

We are continuing to monitor jobs if there are any other errors other than the deadlock, but so far all good.

@benwbrum @saracarl

WillNigel23:


benwbrum (approved):
This looks good to me.

app/jobs/article/rename_job.rb
```diff
@@ -1,10 +1,14 @@
 class Article::RenameJob < ApplicationJob
   queue_as :default
 
+  retry_on StandardError, wait: 5.seconds, attempts: 3
```
saracarl, line 4: @WillNigel23 Can we get a PR to up to this 5 like we talked about on Wednesday?  Thx.

PR #4952, opened 2025-10-02
4945 - Refactor transcribe preview action

### LINKED ISSUE #4945: Refactor save_transcription's `preview` action

WillNigel23:


benwbrum:
This looks good, but while testing it I realized that "mark as blank" doesn't really make sense from the preview mode, nor does Autolink.  We should probably hide any actions that aren't either "save" or "edit".

PR #4970, opened 2025-10-07
4965 - Subject counts are negative

### LINKED ISSUE #4965: subject counts are negative
Some subject page counts are negative for `cwrgm-rev2`, even though there are multiple pages linking to them.

Image

Image

WillNigel23:
The main problem why it is negative is we are using `pages_count` counter_cache, which is flaky to begin with and further exacerbated by moving some processes to async job. I don't want to put my faith into counter_caches, so the solution is to use direct count query.

This is where we encounter another problem. As it is, `article#list` page is already VERY slow as we are querying all categories together with it's child categories, their articles, and their counts, then displaying them all in html while hiding some results. This may work well for smaller collections, but for cwgrm-rev2 for example, this takes more than 10 seconds to load! How much more adding queries to pages_count?

So the solution outlined here is we should only load articles per active category. This will lessen load times significantly, plus we are optimizing the queries happening here. To achieve the existing responsiveness, we implement custom stimulus controllers for the `accordion` effect AND add support to `filter_table` for list option filters. Overall this should be faster and have accurate pages_count.


benwbrum (changes requested):
If counter caches are unreliable, we should try to remove them everywhere.  I don't see any modifications to the models that would indicate removing them from the association and tables.  I would also like to see if we are displaying a page count in other places, like perhaps in the potential duplicate list in `article#show`.

benwbrum:
I'm seeing this error:
```
IndexError - element size differs (17502 should be 17503):
  app/models/article.rb:79:in `sort_vertically'
  app/controllers/article_controller.rb:41:in `list'
```

app/controllers/article_controller.rb
```diff
@@ -11,16 +11,39 @@ def tooltip
...
     end
 
-    @uncategorized_articles = sort_vertically(
-      articles.where(categories: { id: nil }).reorder(:title)
-    )
+    @pages_count_map = @articles.left_joins(:page_article_links)
+                                .group('articles.id')
+                                .pluck('articles.id, COUNT(page_article_links.id)')
+                                .to_h
+    @articles = Article.sort_vertically(@articles)
```
WillNigel23, line 41: Please comment this out @benwbrum and let me know how fast it goes.

PR #4980, opened 2025-10-14
4976 - Subject upload should populate all attributes of article records

### LINKED ISSUE #4976: Subject upload should populate all attributes of article records
Recently we [added several new attributes to subjects](https://github.com/benwbrum/fromthepage/pull/4598/files#diff-24503fd25ed68e6ebceee4951bc4f9b255b197278d4aa9d86ef9d5afe3f26bea).  Currently the subject CSV upload [only accepts the old attributes](https://github.com/benwbrum/fromthepage/blob/development/app/controllers/article_controller.rb#L300), and we have a customer who would like to use the upload for creating or updating subjects with the research they have already done in a spreadsheet.

Email from Kylie (note that "Description" is our `source_text`:
> I wanted to follow up on our meeting of a few weeks ago about the question regarding being able to import subject-related metadata directly into FtP via a spreadsheet (with appropriate field headings). 
> 
> I understand that Bruno had spoken to you about this already, and that it should be possible to ingest the information via the “Upload Subjects” function—could you please confirm that this is indeed correct? 
>  
> Additionally, could you please clarify whether any and all non-autogenerated or non-backend fields be updated via such import? That is, if we make edits to a field in the CSV, will those changes carry over upon import to FromthePage? 
> 
> To give you an overview of what we intend to include: we hope to be able to populate and activate (for public view) the following fields via this import:
> 
> People:
> * Title
> * Birth date
> * Death date
> * URIs (Getty Vocabularies; Wikidata)
> * Bibliography
> * Description (for select entities only)
> 
> Places, Businesses, Institutions:
> * Title
> * URIs (Getty Vocabularies; Wikidata)
> * Bibliography
> * Description (for select entities only)
> 
> Objects
> * Title
> * URI (Getty Vocabularies)
> * Description (for select entities only)
> 
>

WillNigel23:


app/interactors/article/import_csv.rb
```diff
@@ -0,0 +1,64 @@
...
+  ].freeze
+
+  ADDITIONAL_HEADERS = [
+    'bibliography',
+    'birth date',
+    'death date',
+    'disambiguator',
+    'race description',
+    'sex'
+  ].freeze
```
benwbrum, line 20: In addition to these headers, we should include the ones that apply to organizations or places:
```
#  begun            :string(255)
#  bibliography     :text(65535)
#  birth_date       :string(255)
#  death_date       :string(255)
#  ended            :string(255)
#  latitude         :decimal(7, 5)
#  longitude        :decimal(8, 5)
#  race_description :string(255)
#  sex              :string(255)
#  source_text      :text(16777215)
#  title            :string(255)
#  uri              :string(255)
```

app/interactors/article/import_csv.rb
```diff
@@ -0,0 +1,64 @@
+class Article::ImportCsv < ApplicationInteractor
+  REQUIRED_HEADERS = [
```
benwbrum, line 2: I'm not quite sure how the logic will work, but note that `article`, `uri`, and `category` are all actually optional.  A project owner should be able to upload a list of subject titles without any other attributes.

app/views/article/upload_form.html.slim
```diff
@@ -1,4 +1,3 @@
-=javascript_include_tag 'ftp_interface'
 .litebox-embed(style="width:550px")
   h1= t('.upload_subjects')
   p== t('.upload_subjects_description', link: (link_to t('.csv_file_with'), article_upload_example_path, {target: "_blank", alt: t('.example_csv')}))
```
benwbrum: We need to add the new headers to the example CSV file as well so that users don't have to guess at the headers.

benwbrum (changes requested):
We need to include a few new attributes in the upload and also add the new attribute headings to the example file.

app/views/article/upload_form.html.slim
benwbrum: @WillNigel23 We still need to add the new headers to the downloaded CSV file.

WillNigel23:
Had to rebase this @benwbrum

benwbrum:
I still don't see the additional headings on the template file we produce in the subject upload screen:
Screenshot from 2025-11-07 14-13-56

WillNigel23:
Whoops, looks like I missed it in the example download

WillNigel23:
@benwbrum ready to review again.

saracarl:
When I tested this with [CHP Import.csv](https://github.com/user-attachments/files/23780444/CHP.Import.csv)

I got the following message:
image

despite the spreadsheet having HEADING, CATEGORY, URI, and ARTICLE.

saracarl:
tested well.

PR #4981, opened 2025-10-15
4979 - Add permission check on subjects upload

WillNigel23:


WillNigel23:
Take note, we should merge #4976 first and rebase this to development before merging.

PR #4992, opened 2025-10-21
4986 - New collaborators duplicate

### LINKED ISSUE #4986: "New Collaborators" repeats same collaborator many times
Krystyna at the WWP reports:

I've also noticed on the collaborators page under the New Collaborators page that it's repeating, see below. It doesn't cause us problems - just letting you know it seems like a glitch.

Image

I see the same thing with the Graves project.  I'm guessing it repeats based on the number of contributions.

Image

I also notice that when the collaborators list loads, it says [email protected] after the collaborator name, but then eventually replaces that with their email.  (separate issue?)

WillNigel23:


WillNigel23:
> I also notice that when the collaborators list loads, it says [email protected] after the collaborator name, but then eventually replaces that with their email. (separate issue?)


I can't seem to replicate this part of the issue @benwbrum @saracarl

PR #4993, opened 2025-10-21
4985 - Flash alert bug on landing page

### LINKED ISSUE #4985: Sign Out flash alert never closes
A transcriber complained that the "Signed Out Successfully" message never goes away once you sign out:

Image

I tested and she is right.  It also blocks the menu items.

Can we make this one a flash that closes after a while, like the sign in message?

WillNigel23:


WillNigel23:
I was very confused as to why this was not working. I thought perhaps the `stimulus` controller was not working due to the sign out redirect being intercepted by `ajax` or `turbo` and having weird interactions with it. I tried debugging more and found out that the `stimulus` js is not loaded regardless of request type.

Upon inspecting the page I saw that we are using a custom layout and headers for the root_path, and we left out the import of stimulus controllers. I added it there to get the flash to work, it should not cause much in terms of performance so it should be good.

saracarl:
Makes sense, and looks like it still works with all the fancy landing page js.

PR #5019, opened 2025-10-27
5009 - Bulk CDM import es index

### LINKED ISSUE #5009: ES work index not updated after bulk CDM import
A rarely used rake task `bulk_import_cdm` imports one or more IIIF manifests from CONTENTdm.  When CWRGM does this, however, the works created are not added to their ElasticSearch index, so when their students search for a document they have been assigned by a metadata value, they can't find it.

Could we either reindex the collection after a bulk import, or make sure that whatever logic fires on work modification is actually happening here?

WillNigel23:


saracarl (approved):

benwbrum (approved):
This is a really good implementation.

PR #5020, opened 2025-10-27
5008 - Metadata refresh logs include 503 errors

### LINKED ISSUE #5008: Metadata refresh log emails include 503 errors
Metadata refresh emails include 403 errors thrown by CONTENTdm occasionally.  Since CONTENTdm dependencies are fragile, we should catch these, wait, and retry, then include a friendlier message in the logs since they are sent to project owners.  (See attached file for an example sent by Lindsey at CWRGM)

(cf #4878)

[refresh_logs.txt](https://github.com/user-attachments/files/23109047/refresh_logs.txt)

WillNigel23:


app/interactors/work/metadata/refresh.rb
```diff
@@ -22,14 +29,23 @@ def process_batches(works)
...
+        if ERRORS_TO_RETRY.any? { |msg| e.message.include?(msg) } && retries < MAX_RETRIES
+          wait_time = RETRY_DELAYS[retries]
+          log("#{e.message} - retrying in #{wait_time}s (attempt #{retries + 1}/#{MAX_RETRIES})")
+          sleep(wait_time)
+          retries += 1
+          retry
+        else
+          @success = false
+
+          log("Failed to refresh metadata for #{work.slug}")
```
WillNigel23, line 45: I think there's still value to keeping the 503 errors in logs just so we know it happened and can point to CDM error.

That being said I added retry and delay here.

@benwbrum

saracarl (approved):

benwbrum (approved):
This looks really good to me.

PR #5023, opened 2025-10-28
5000 - OCR change causes innaccurate work statistics

### LINKED ISSUE #5000: Page counts and percent transcribed off (with the same value) for most works in a collection
Many (but not all!) of the works in this collection are reporting incorrect page counts.  

For instance, [this work](https://fromthepage.com/nharl/admiral-sir-isaac-coffin-s-lancasterian-school/ms99-folder4-item1) has one page in reality:

Image

but in the collection list work view it [shows as having 4 pages](https://fromthepage.com/nharl/admiral-sir-isaac-coffin-s-lancasterian-school):

Image

And then the metrics also don't make sense -- there's no way to calculate a 50% transcribed metric on a work of 1 page.

I'm pretty sure this is a duplicate of #4778 but it's slightly different enough (more pages rather than too few) that I'm opening it as a separate issue.  (And that one was "fixed" by resaving a page in each work before we could identify the cause.)

I think what's happening, somehow, is that "all" (??) works are getting the page count and percent complete/transcribed from one of the works (like, when it's updated on a individual work, we somehow apply that stat to all the works).  This is relatively easy to see if you scan through [the works list](https://fromthepage.com/nharl/admiral-sir-isaac-coffin-s-lancasterian-school/works_list).

There are some works that get their correct metrics -- I think they have pages saved (and work statistics recalculated) *after* whatever causes this problem has occurred.  This implies that we might be able to identify the deed that caused the problem.  I'm going to try that now...

OK.  I think that's it.  If I look at the[ works list and order by "most recently transcribed"](https://fromthepage.com/nharl/admiral-sir-isaac-coffin-s-lancasterian-school/works_list?sort=activity&order=desc&per_page=200&search=&show=all) I see that every work has 4 pages and 50% transcribed *except* for the most recently edited one:

Image

So I would hypothesize that the last edit before "Folder 1 Marriage Records, Item 3" would be the culprit.  That works is "Folder 5 Correspondence William Coffin Letters In, Item 4":
Image

BUT that doesn't quite make sense because "Folder 5 Correspondence William Coffin Letters In, Item 4" is only a single page document, not a 4 page document that is only 50% complete.  

The next work on the deed list, "Folder 4 Correspondence Gorham Coffin Letters Out, Item 2" is also not a 4 pager.

I did not find a page for which the "4 pages, 50% complete" was accurate until I got to the 10th most recently edited work on this works list:  ["Folder 2 Incorporation, Item 1"](https://fromthepage.com/nharl/admiral-sir-isaac-coffin-s-lancasterian-school/ms99-folder2-item1).

So I think that means my theory on this coming from the most recently edited is incorrect.

If you search the activity stream for the most recent edits to "Folder 2 Incorporation, Item 1" it's quite a ways down:

Image

And I don't see any uniqueness to these edits; MaryV both transcribes and edits other works earlier in the "stack".

I don't have any other ideas what could be causing this!  I do suspect that we could "fix" it by saving a page in each document, but I would really like to figure out what caused it.

WillNigel23:


saracarl (approved):
Reviewed and tested -- looks great!

PR #5028, opened 2025-10-30
Update annotations

benwbrum:
Update auto-generated files.


WillNigel23:
This does not look correct. Why does this PR contain article/lib/delete.rb ? If I recall correctly it should be changes only for db/schema.rb (and maybe some annotation files)

benwbrum:
Thanks -- you're totally right, I missed that in my commit.

PR #5030, opened 2025-11-03
Revert "Update annotations"

WillNigel23:
Reverts benwbrum/fromthepage#5028

@benwbrum I missed it initially, but for some reason the `transcription_json` column is considered `text` instead of json column in db/schema.rb. It clearly should be json right? Is this an artifact of difference in db versions? In either case we should consider reverting...


WillNigel23:
We will not revert afterall.

PR #5031, opened 2025-11-03
4997 UI and implementation for subject search

### LINKED ISSUE #4997: UI and Implementation for subject search
We would like to add subject search to the existing searches within the site-wide, owner, collection, and document set scopes.  The search should search against subject title and article text, but not any of the structured fields like `birth_date` or `bibliography`.

We'll need a new partial to display the subject with the appropriate collection (or docset) breadcrumb at the bottom for context.  I don't think we have any options for thumbnail images, so we're stuck with a text-only display.  We should be able to handle subjects with no article (just display Title), subjects with an article (display the article or a truncated version of it, as we do with page search results), and maybe should also indicate what category the subject is in.

WillNigel23:


saracarl:
Only works for collection search, but logging problems in an issue and merging this anyway.

PR #5037, opened 2025-11-05
5036 - Flash message unhandled nil

### LINKED ISSUE #5036: 500 error with nil flash message
```
F, [2025-11-05T16:15:56.133831 #894765] FATAL -- :   
ActionView::Template::Error (undefined method `html_safe' for nil):

Causes:
NoMethodError (undefined method `html_safe' for nil)
     5:   <div class="flash flash-<%= type %>" data-controller="flash" data-flash-type-value="<%= type %>" role="<%= aria_attrs[:role] %>" aria-live="<%= aria_attrs[:'aria-live'] %>" aria-atomic="<%= aria
_attrs[:'aria-atomic'] %>">
     6:     <span class="flash_message">
     7:       <%= svg_symbol(icon, class: 'icon') %>
     8:       <span><%= message.html_safe %></span>
     9:     </span>
    10:     <a class="flash_close" data-action="click->flash#close">&times;</a>
    11:   </div>
  
app/views/shared/_flash.html.erb:8
app/views/layouts/application.html.slim:44
app/views/layouts/application.html.slim:43
app/controllers/application_controller.rb:68:in `switch_locale'

```

WillNigel23:


app/controllers/article_controller.rb
```diff
@@ -80,7 +80,7 @@ def delete
     if result.success?
       redirect_to collection_subjects_path(@collection.owner, @collection)
     else
-      flash.alert = result.message
+      flash[:alert] = result.message || t('errors.error')
```
WillNigel23, line 83: It seems sometimes it fails and result.message is nil. This can happen if the failure happens in the `destroy!` call instead of the context failure:

```
  def perform
    context.fail!(message: I18n.t('article.delete.only_subject_owner_can_delete')) unless user_can_delete_article

    versions = @article.article_versions.where('created_on > ?', 1.hour.ago)
    @article.destroy!

    Article::RenameJob.perform_later(
      user_id: @user.id,
      article_id: @article.id,
      old_names: versions.pluck(:title) + [@article.title],
      new_name: ''
    )
  end
```

So it means the failure to delete the article is DB/model level. For now I added a generic failure message in case there is nil. Will investigate destroy failure for now

app/views/shared/_flash.html.erb
```diff
@@ -1,12 +1,12 @@
 <% icon = flash_icons[type.to_sym] %>
 <% aria_attrs = flash_aria_attributes(type) %>
 
-<% if icon.present? %>
+<% if icon.present? && message.present? %>
```
WillNigel23, line 4: Let's add nil handling on flash messages to avoid having this occur in the future.

PR #5059, opened 2025-11-10
5033 - Add search in subjects view

### LINKED ISSUE #5033: Add search in subjects view

WillNigel23:


benwbrum:
If I search for an indexed subject that is in the first category listed, we see the subject in the subject screen.  However, if I search for a subject that is in a different category, I see no results.

I'm not sure what the proper behavior should be here -- it seems correct that, if a user has opened the Places category and searched for "Virginia", they see the subjects for Virginia within the Places category only.  But it seems incorrect that someone opening the Subjects tab for the first time and searching for "Virginia" will not see any results, or will only see results in the first category.  

Maybe we could add some kind of indicator for search results being present next to the category selector?

benwbrum:
Sara and I have discussed the UI challenge and we think that we have a solution:
* First, we need an `x` or some other way to clear the search term without needing to delete text and lose focus
* On the categorization/search problem, 
  * a search should search across all categories in the collection, always
  * the category selector tree should be pruned to only categories that have search results, so that users don't have to click on categories that look empty because they had no matching results
  * if a category is currently selected and has search results, leave that category selected
  * in any other case, we select the first category that contains search results
    * that handles the "I just clicked the Subjects tab and searched" use case
    * that also handles the "The category I was looking at has no results" case

saracarl (changes requested):
image

The UI is not very usable.  Can we build the list of search results and only show the category if there is a result in it?  And can we remove the "edit" menu from the category title?

saracarl:
Better, but a bit off:
image

image

Not sure why "military" is always showing up.

Also, whenever I click out of the browser (say to get a screenshot), the entire page reloads.

WillNigel23:
Can I get a dump of the collection in question? @saracarl 

As far as I know, for articles we search these fields:
```
    search_fields = [
      'title^2',
      'title.no_underscores^1.3',
      'content_english' # aka source_text
    ]
```

So could `Aspen Grove` contain `tobacco` in it's source_text?

The military one is baffling to me though. No idea why it shows :thinking: 

> Also, whenever I click out of the browser (say to get a screenshot), the entire page reloads.

This has not yet happened to me, let me see if I can reproduce.

Issue #5124, opened 2025-11-26
Work metadata upload doesn't work


saracarl:
Uploading the following metadata for [this collection](https://fromthepage.com/cca/alvaro-siza-sketchbooks-pilot-project/edit/look#)

[FromthePageMetadataTextV3 (1).csv](https://github.com/user-attachments/files/23780239/FromthePageMetadataTextV3.1.csv)

Doesn't create the work metadata as expected.

Here's what I see in the log (not much -- it goes to SolidQueue):

```
I, [2025-11-26T21:08:41.195461 #1657362]  INFO -- : Started POST "/collection/metadata/create" for 2605:a601:a095:9600:8495:b51a:b3fc:15a4 at 2025-11-26 21:08:41 +0000
I, [2025-11-26T21:08:41.196729 #1657362]  INFO -- : Processing by MetadataController#create as */*
I, [2025-11-26T21:08:41.196785 #1657362]  INFO -- :   Parameters: {"authenticity_token"=>"mEAmzqTP_E2fQ_JHCeOnRo29h6t47qpQpANAlPihugPPbQxZWMhmNGcGfW17oNiRCfeGlpzAkUdlZ3-oAYL_zg", "metadata"=>{"file"=>#<Ac
tionDispatch::Http::UploadedFile:0x00007516451ae1b0 @tempfile=#<Tempfile:/tmp/RackMultipart20251126-1657362-kb82k.csv>, @content_type="text/csv", @original_filename="FromthePageMetadataTextV3 (1).csv", @h
eaders="Content-Disposition: form-data; name=\"metadata[file]\"; filename=\"FromthePageMetadataTextV3 (1).csv\"\r\nContent-Type: text/csv\r\n">, "collection_id"=>"32001725"}, "button"=>""}
I, [2025-11-26T21:08:41.217768 #1657362]  INFO -- : [ActiveJob] Enqueued Work::Metadata::ImportCsvJob (Job ID: e9419fb5-4fb9-4c11-b9cb-69458470a4d5) to SolidQueue(default) with arguments: {:metadata_file_
path=>"/home/fromthepage/deployment/releases/20251126000023/tmp/import_csv/1764191321.csv", :collection_id=>"32001725", :user_id=>32037440}
I, [2025-11-26T21:08:41.223821 #1657362]  INFO -- : Completed 201 Created in 26ms (ActiveRecord: 5.8ms (12 queries, 0 cached) | GC: 0.5ms)
[...]

PR #5129, opened 2025-12-01
5124 - Work Metadata upload better job logging

WillNigel23:


benwbrum:
Note that this addresses but does not fix #5124

PR #5141, opened 2025-12-02
5140 - Suspicious behavior model and large paste flagging

### LINKED ISSUE #5140: Create Suspicious Behavior Table and support for large paste flagging

WillNigel23:


app/javascript/libs/behavior_flagger/core/foundations/constants.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Constants - Type definitions and configuration values
+ *
+ * Responsibilities:
+ * - Provides consistent identifiers and values across the system
+ */
+export const PASTE_CHARACTERS_THRESHOLD = 50;
```
WillNigel23: Feel free to modify this value. This is just arbitrary number I have decided on

app/javascript/libs/behavior_flagger/mixins/usePasteFlagger.js
```diff
@@ -0,0 +1,44 @@
...
+      );
+    }
+  }
+
+  return () => {
+    // Nothing to clean up for now
+  };
+};
+
+export default usePasteFlagger;
```
WillNigel23, line 44: This is a pattern I learned from one of my mentors who's really good with JS. I chose this because:

1. I want an isolated mixin per behavior_type we are flagging. Right now we only  have support for `large_pastes`. 
2. It's easier to maintain and modify. If I want to add new feature, we know where to add it. If I want to update existing feature, we don't break other existing features.

app/javascript/libs/behavior_flagger/core/foundations/constants.js
benwbrum: In some cases we allow/encourage users to paste from the AI Assist into the editor.  These should probably be a maximum of 2-3 lines, so let's change this to 200 characters.

db/migrate/20251202174404_create_suspicious_behaviors_table.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      t.json :metadata, null: true
+
+      # Types: [:large_paste, :high_wpm, :chatgpt_tell, :low_backspace, :no_image_adjustment]
+      t.integer :behavior_type, null: false
+
+      # Types: [:pending, :confirmed, :dismissed]
+      t.integer :status, null: false, default: 0
+
+      t.datetime :flagged_at, null: false
```
benwbrum: Is this the same as `created_at`?  If so, I'd be happy to use that and alias/delegate the `flagged_at` method in the model.

app/models/suspicious_behavior.rb
```diff
@@ -0,0 +1,64 @@
...
+#  fk_rails_...  (resolved_by_user_id => users.id) ON DELETE => nullify
+#  fk_rails_...  (user_id => users.id) ON DELETE => cascade
+#
+class SuspiciousBehavior < ApplicationRecord
+  belongs_to :user
+  belongs_to :page, optional: true
+  belongs_to :collection, optional: true
+  belongs_to :resolved_by_user, optional: true, class_name: 'User'
+
+  enum :behavior_type, {
```
benwbrum, line 37: In the `Page` model, we were able to use string values for our status `enum`.  This gives us the benefit of `enum` behavior while still letting raw DB access look readable.  Could we do that here instead of using integers?

app/controllers/admin/suspicious_behaviors_controller.rb
```diff
@@ -0,0 +1,50 @@
+class Admin::SuspiciousBehaviorsController < AdminController
```
benwbrum, line 1: I see that this is namespaced as `Admin::`.  If we build a separate view to show owners suspicious behaviors within their own collection, will that need to be in a separate namespace?  How will code re-use work for shared view and controller functionality, since the views will differ by a single column (`owner`) and a filter on collections?

app/javascript/libs/behavior_flagger/mixins/usePasteFlagger.js
benwbrum: I like this.  It seems like we'll be able to attach the flagger to either CodeMirror or HandsOnTable pretty cleanly?

app/views/admin/suspicious_behaviors/_table.html.slim
```diff
@@ -0,0 +1,47 @@
+table.datagrid.striped.dataTable.no-footer
+  thead
+    th
+    th =t('admin.suspicious_behaviors.table.user')
+    th =t('admin.suspicious_behaviors.table.collection')
```
benwbrum, line 5: The admin view will need to have the name of the collection owner as well, so that we can know who to contact.

benwbrum (changes requested):
This is very impressive indeed.  Let's add the owner view and start testing.

db/migrate/20251202174404_create_suspicious_behaviors_table.rb
WillNigel23: Actually I'd prefer not adding flagged_at at all. the built-in `created_at` timestamps should be enough here

app/models/suspicious_behavior.rb
WillNigel23: Sure, that is possible. I just went ahead and used integer so it is smaller in DB. But that can be arranged.

app/controllers/admin/suspicious_behaviors_controller.rb
WillNigel23: Reusing controllers in multiple different contexts and views leads to fat controllers and harder separation of privileges. The function we are building is not supposed to be a restful endpoint, and as such I'd rather have different controllers for each view, one for admins and one for collections.

For views, we can easily create shared components so there is less html written. But for controllers it's best we keep them separate, under their own namespace, so redirects/responses are not within 3+ nested if statements.

app/javascript/libs/behavior_flagger/mixins/usePasteFlagger.js
WillNigel23: That is the idea. At the current set up it still requires the `reportBehavior` function from the base controller, but that can easily be arranged later on.

db/migrate/20251202174404_create_suspicious_behaviors_table.rb
benwbrum: Souns good!

app/javascript/libs/behavior_flagger/mixins/usePasteFlagger.js
benwbrum: That sounds good to me.

app/controllers/admin/suspicious_behaviors_controller.rb
benwbrum: It sounds like you've been working in the `transcribe_controller` again!  That sounds fine to me.

app/models/suspicious_behavior.rb
benwbrum: Thanks!  I generally prefer readability over space optimization, and have been burned by integers in schemas in the past.

PR #5145, opened 2025-12-04
5144 - Owner level suspicious behavior dashboard and controls

### LINKED ISSUE #5144: Owner-level suspicious behavior dashboard should have controls
Owners should be able to decide whether to confirm/dismiss suspicious behaviors.

IF dismissed:
- greenlist user so no more new records are created under their name ???

If confirmed:
- block user from transcribing

Dashboard should allow changing decisions later on.

WillNigel23:


benwbrum:
I have tested this and it seems to work well.  The only things that I'd like to see are changes to the search boxes to make sure that project owners can search for user by display name, real name, email, or slug, and that admins can search for users and owners by the same, and for collections by title or slug.

saracarl (changes requested):
Changes:
1)  Is this supposed to show up for owners or staff on a collection?  It doesn't.  (I'm hoping this is the PR where we turn this back on, assuming there is suspicious behavior on a collection.)
2)  Collection search doesn't seem to work at all, using the beginning of a collection title, or a full slug.
3)  The search prompt is "<field name> id or slug".  No one in the UI is going to have (or go to the rails console to get) an id.  And human users don't know the word "slug".  Let's just say "search <field name>"  (and I don't think it needs to search by an ID)
4)  the primary way people are going to use this is by searching the display strings for that object.  So if the user is "Sara Carlstead Brumfield" a search for "Carlstead" should bring up that user's suspicious behaviors.  It doesn't.
5)  I think if the display string searching is robust enough, we won't need to do the "copy slug" icon in the listing.  It's extraneous and confusing.

WillNigel23:
We should probably use ES here as well then.

app/views/shared/_collection_tabs.html.slim
```diff
@@ -96,7 +96,7 @@
     :path => "#{main_app.collection_works_list_path(@collection.owner, @collection)}",
   }]
 
--if false && @collection.owner == current_user
+-if @collection.owner == current_user
```
saracarl: This does not let staff see suspicious behaviors.  (This was explicitly called out as a use case in my last review -- please don't waste my time.)

saracarl (changes requested):
Search by user doesn't work (do we actually ES index *all* users?  I'm not sure we do, just owners.)

Preview review, the dashboard looks like this:

Image

Now it looks like this (note blue highlights):

Image

Can we merge development in?

WillNigel23:
The highlighted text here is a missing `status`, likely due to this branch not being in sync with development where we introduced new suspicious_behaviors.

It should be fixed now after merge.

saracarl:
This looks good, but collection search doesn't seem to work.

saracarl (changes requested):
This looks good, but collection search isn't working (the others are).  (with ElasticSearch, fwiw.)

WillNigel23:
After diving deeper I found the issue.

It was working locally for me because I am searching collections my logged in user `HAVE` access to. This assumes our `es_search` setup does not filter by collections a user has access to (ex: blocked, private, etc)

But this assumption is wrong, and it does not make sense to have those restrictions when this is an `admin` page. I am adding an `admin` param to the es_search so whenever we just want to match the query without any other filters we can use this.



@saracarl

saracarl:
@WillNigel23 that doesn't add up to me -- I'm the person creating the suspicious behavior, and I have (staff!) access to the collections and am an admin (and testing in the admin view, as well as the collection view).  

The most recent iteration of the code takes away my suspicious behavior searches completely. (!)

image

WillNigel23:
That's weird!

image

When in admin page, there is 0 conditionals present on what search inputs are shown.

WillNigel23:
image

This is how it looks in `development` branch, similar to how it is in your screenshot.

Let me check github status again, I think it's glitchy now! Earlier when I pushed changes here I saw the lint for `linked_files_changes` when I haven't merge development HERE yet. Which was very weird (it also happened on another PR before merging from development! https://github.com/benwbrum/fromthepage/pull/5328)

WillNigel23:
GH actions are still experiencing degredation bugs: 
image

saracarl:
@WillNigel23 in this version collection search works, but owner doesn't ("Ben" or "Brumfield") and user search works for "Brumfield" but not "Sara" or "Carlstead".

saracarl (approved):
Works well, as long as the searchable users are actually working in ES!

Image

PR #5146, opened 2025-12-04
5120 - Display gemini reasoning on ai tab

WillNigel23:


db/migrate/20251204202855_create_ai_transcriptions_table.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAiTranscriptionsTable < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ai_transcriptions do |t|
+      t.references :page, null: false, foreign_key: { on_delete: :cascade }, type: :integer
+      t.text :source_text
+      t.string :model, null: false
```
benwbrum, line 7: Let's add a string for prompt or prompt filename here -- possibly in the future.

app/models/page.rb
```diff
@@ -612,24 +615,33 @@ def contributors
   end
 
   def has_ai_plaintext?
-    File.exist?(ai_plaintext_path)
+    self.ai_transcription.present? || File.exist?(self.ai_plaintext_path)
```
benwbrum, line 622: I would prefer us to write a migration that reads plaintext files into `ai_transcription` models (defaulting the model to "Transkribus+OpenAI") then deletes the files, so that we have a single source of truth.

benwbrum (changes requested):
This looks good and works well.  I'd prefer to migrate all the ALTO-generated transcriptions to the new AR model and delete the files, so that our code is a bit shorter and we don't have two possible sources of truth.  I'm not very concerned about how long this migration takes to run, either.

WillNigel23:
We have 2 rake tasks added for migrating to new implementation

`bundle exec rake 'fromthepage:migrate_ai_transcription'`
Run this to migrate old alto and old transcriptions to the new version.


`bundle exec rake 'fromthepage:delete_ai_transcription_files'`
Run this to delete the files once processed.


I decided to separate these two so we can first confirm if the migration is correct before we delete the files.

PR #5152, opened 2025-12-09
5151 - Fix alto migration

WillNigel23:


benwbrum (approved):
This looks reasonable to me.

PR #5155, opened 2025-12-10
5081 - Record when a transcriber uses AI

### LINKED ISSUE #5081: Record when a transcriber uses AI
Currently, if a page has an `ai_plaintext` file, we display an AI Draft button, letting the user copy in the text we have produced using an AI engine, so that they can use it as a first draft to edit.  However, this is a function that has potential for abuse, and we want to track when transcribers have used it and the material on which it has been used.  

As a result, we should:
* Log a new deed when a user presses the AI Draft button.  This should be publicly visible, with a deed type of `ai_draft` and description "Used an AI draft for _page_ in _work_" or something similar.
* Add a new `ai_draft_used` field to `page_versions`, defaulted to `false`, but which is set to `true` if a transcriber has used an AI draft editing a page and then saved that page.  The page version screen should indicate this visually -- next to each version, we should display a small `AI` alongside status, user, etc.  (There is no need to show anything if this is `false`)

WillNigel23:


benwbrum:
@copilot Please merge the `development` branch into this branch.

benwbrum:
This looks pretty good and works well.  I have a couple of UI tweaks I'd like to see before we ship this:

### Column alignment in the version list
I'm not sure how easy it will be to do this, but it would be very nice if the justification of the AI icon, status icon, and names of contributors was in a consistent vertical line, rather than whatever it's doing here:
Screenshot from 2025-12-19 08-39-18

### Order of deed logging
The activity stream shows the "used AI draft" occurring after the "edited page" deed, when in fact it happened before the edit.  I'm not sure how we can change the order in which we log these deeds to the DB, but it would be nice to get them in order.  (I'm aware that this is also a problem when someone edits/transcribes a page and marks it as needing review -- the order is counter-intuitive there.)

Screenshot from 2025-12-19 08-53-15

PR #5157, opened 2025-12-11
Fix broken AI transcriptions again

benwbrum:
The previous migration code was not aware that multiple plaintext and alto files are stored in a folder for a work, so it was grabbing a file from the folder and using it for all pages in the system. This should fix that problem by following our naming conventions to find the files corresponding to the page_id and create a record from them.

This also contains convenience scripts to clear out old `ai_transcription` records and rename the folders, so that we can clean up the bad migration before re-running this.

### DIFF
--- a/lib/tasks/migrate_ai_transcriptions.rake
+++ b/lib/tasks/migrate_ai_transcriptions.rake
@@ -1,5 +1,4 @@
 namespace :fromthepage do
-
   desc 'Cleanup bad migration'
   task cleanup_ai_migration_folders: :environment do
     # first rename all of the folders in public/text that end with _processed back to original name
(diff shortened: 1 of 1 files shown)


lib/tasks/migrate_ai_transcriptions.rake
```diff
@@ -40,12 +67,17 @@ namespace :fromthepage do
...
-        AiTranscription.create!(
-          page_id: page.id,
-          source_text: source_text,
-          model: model,
-          prompt: prompt
-        )
+        begin
+          AiTranscription.create!(
+            page_id: page.id,
+            source_text: source_text.gsub('🤔', '?'),
```
saracarl: get rid of this.

saracarl (changes requested):

WillNigel23 (commented):
LGTM

PR #5240, opened 2026-01-27
5238 - Missing paragraph break fix

### LINKED ISSUE #5238: Newly saved pages lose paragraph breaks in word-wrap display
Users have complained that we are losing the paragraph breaks when we display transcriptions in word-wrap in the read_work view.  At first I thought this was due to changes to how we convert XML to HTML, but older pages do show the paragraph breaks correctly.  It looks like we are doing something wrong when we transform wikitext into XML.

See the version history on these two pages for an example: https://fromthepage.com/hagley-library/wood-family-letters/work-193-08-13

WillNigel23:
https://github.com/benwbrum/fromthepage/pull/5224 introduced a bug where if there is single whitespace between `\r\n`s then it considers it as `<indent space="1">`  so that our parser misses that they are actually paragraph breaks instead.

I propose the solution: if space="1", let's not add the <indent> tag. Anything > 1 will be considered a `tab` and therefore an indent.

### DIFF
--- a/app/models/xml_source_processor.rb
+++ b/app/models/xml_source_processor.rb
@@ -340,18 +340,16 @@ end
   # Process initial whitespace at the beginning of lines
   # Replaces leading spaces with <indent> elements containing a spaces attribute
   def process_initial_whitespace(text)
-    # Process initial whitespace at the very beginning of the text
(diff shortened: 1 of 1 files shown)


saracarl:
Much better.  
Before:
image

After:
image

PR #5249, opened 2026-01-29
5244 optimize page linking

### LINKED ISSUE #5244: Time-outs saving page with lots of tags
Jason at the CHL is running into problems saving this page: https://fromthepage.com/woodruff/george-q-cannon-journal-1890/george-q-cannon-journal-october-1890/transcribe/35382570

Cloudflare is timing out when FromThePage does not respond within 120 seconds.  Based on the log entries, saving may take as long as 12 minutes.  (See attached log file)

[jason_error_1377061.txt](https://github.com/user-attachments/files/24922044/jason_error_1377061.txt)

We suspect that the problem is the number of tags in the text, although it looks like there are only 50.  Other possibilities include the number of subjects (15K) or the HTML mark-up within the page.

benwbrum:
Closes #5244

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -14,7 +14,7 @@
 #  lock_version            :integer          default(0)
 #  metadata                :text(65535)
 #  position                :integer
-#  search_text             :text(65535)
+#  search_text             :text(16777215)
 #  shrink_factor           :integer
 #  source_text             :text(16777215)
 #  source_translation      :text(16777215)
@@ -36,6 +36,7 @@
 #  index_pages_on_status_and_work_id                      (status,work_id)
 #  index_pages_on_status_and_work_id_and_edit_started_at  (status,work_id,edit_started_at)
 #  index_pages_on_work_id                                 (work_id)
+#  pages_search_text_index                                (search_text)
 #
 require 'search_translator'
 require 'transkribus/page_processor'
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -560,7 +560,7 @@ class Page < ApplicationRecord
     if self.page_article_links.present?
       self.clear_article_graphs
       # clear out the existing links to this page
-      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").destroy_all
+      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").delete_all
     end
   end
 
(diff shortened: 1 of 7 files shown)


benwbrum:
On my development branch, this drops a test page save from 
```
164475ms (ActiveRecord: 108513.1ms (7491 queries, 1 cached) | GC: 2987.7ms)
```
to
```
29665ms (ActiveRecord: 2557.7ms (2539 queries, 1 cached) | GC: 2859.8ms)
```

The remaining queries appear to be primarily INSERT operations on the `page_article_links` table.  I'm not sure how we could get rid of those without substantial refactoring, since we need the ID of the created record for the XML element.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_12_10_202813) do
+ActiveRecord::Schema[7.2].define(version: 2026_01_29_102417) do
```
WillNigel23: Where did this come from? I don't see `2026_01_29_102417` in `db/migrate` directory. This probably came from a feature branch you were working on that was not yet merged/comitted?

app/models/xml_source_processor.rb
```diff
@@ -391,7 +391,9 @@ def update_links_and_xml(xml_string, preview_mode = false, text_type)
     # log the count of articles before and after
     clear_links(text_type) unless preview_mode
 
-    candidate_articles = collection.articles.left_joins(:article_versions)
+
+    articles_by_title = collection.articles.index_by(&:title) # optimize for common case
+    candidate_articles = collection.articles.left_joins(:article_versions) # fall-back
```
WillNigel23, line 396: LGTM. Can remove the comment since it's pretty self-explanatory

app/models/page.rb
```diff
@@ -559,7 +560,7 @@ def clear_links(text_type)
     if self.page_article_links.present?
       self.clear_article_graphs
       # clear out the existing links to this page
-      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").destroy_all
+      PageArticleLink.where("page_id = #{self.id} and text_type = '#{text_type}'").delete_all
```
WillNigel23, line 563: I checked and there seems to be no `count` or other callbacks associated with PageArticleLink so I'm good with this one.

WillNigel23 (changes requested):

db/schema.rb
benwbrum: Good catch!

WillNigel23 (approved):

PR #5274, opened 2026-02-05
5273 - Minor quality of life improvements for development

### LINKED ISSUE #5273: Minor Quality of Life improvements for development
Few minor niche things to improve with regards to local development that should not affect default development experience for everyone else:

1. Add `env` var for development and test database names. This is especially useful after introducing our dump import/export tool. Instead of editing database.yml file and manually excluding it in committed files.
```
DEVELOPMENT_DATABASE=******
TEST_DATABASE=******
```

2. Disable support for descending index. This is very specific problem for me since the beginning, so please bear with me here. My database migrations fail due to this migration:
```
class ReviseIndexOnPageWorks < ActiveRecord::Migration[6.0] 
  def change 
    # remove_index :pages, [:status, :work_id] 
    add_index :pages, [:status, :work_id, :edit_started_at], order: { edit_started_at: :desc } 
  end 
end
```
where in order to run specs, I need to replace `:desc` with `:asc` ALL the time. I added an env flag `DESC_INDEX_UNSUPPORTED`, which if your db does not support it, just set this to true. Other devs should not be affected. Very specific for me 🙏 

3. Include `WorkFacet` in dump import/export. Discovered after exporting `Jeremiah Graves` collection that this is needed so that there are no errors when collection has facets_enabled on the `collection#show` page.

WillNigel23:


benwbrum:
I love these!

PR #5282, opened 2026-02-09
5243 - Flag as suspicious behavior when users do not change ai draft text

### LINKED ISSUE #5243: Flag as suspicious when users do not change the AI Draft text
In a conversation with a customer yesterday, we learned that some users are "transcribing" a page by clicking "AI Draft", then hitting Save without changing anything substantial.  I'm not sure whether or not the users are removing the warning text at the top, but even if they are, this is not desirable behavior.

We should look at the difference between an AI draft and the transcribed `source_text`, and if it's identical, flag the version/user/page as suspicious behavior.

WillNigel23:


app/interactors/transcribe/flag_ai_use.rb
```diff
@@ -0,0 +1,67 @@
+class Transcribe::FlagAiUse < ApplicationInteractor
+  DISTANCE_FLAG = 0.05
```
WillNigel23, line 2: Since we are normalizing the calculations, `3` is right about 0.05. But feel free to change the threshold here if it's too little/too much @benwbrum

app/controllers/transcribe_controller.rb
```diff
@@ -102,7 +102,15 @@ def save_transcription
       end
 
       if @page.save
-        record_deed(DeedType::AI_DRAFT) if @page.ai_draft_used
+        # TODO: Implement in save_transcription refactor PR as well
+        if @page.ai_draft_used
+          record_deed(DeedType::AI_DRAFT)
+
+          Transcribe::FlagAiUseJob.perform_later(
```
WillNigel23, line 109: Moved to job since it's going to use Levenshtein, which is slow

benwbrum: Good idea!

saracarl (approved):
Looks great!

Image

PR #5285, opened 2026-02-10
dev to ftp

saracarl:


saracarl:
@copilot can you fix this rubocop Offenses for me?

app/helpers/collection_helper.rb:169:53: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
    return true unless current_user.account_type == "Individual Researcher"
                                                    ^^^^^^^^^^^^^^^^^^^^^^^
app/helpers/collection_helper.rb:170:1: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.

PR #5291, opened 2026-02-12
5290 - Calculate approval delta job fix

### LINKED ISSUE #5290: Only run calculate_approval_delta_job on persisted records
Image

Since `calculate_approval_delta` is a `before_save` callback on page, this also runs on non-persisted pages.

WillNigel23:


app/models/page.rb
```diff
@@ -54,7 +54,7 @@ class Page < ApplicationRecord
   before_update :populate_search
   before_update :update_line_count
   before_save :calculate_last_editor
-  before_save :calculate_approval_delta
+  after_save :calculate_approval_delta
```
WillNigel23, line 57: Need extra set of eyes here @benwbrum 

Initially, I kept it a `before_save` callback and essentially check:
1. Is it persisted?
2. Is the source_text going to change

But that felt and looks like it's more of a `after_save` after all. IF it is after_save, then:
1. No need to check if it is persisted. We can assume it is
2. And then check if there was saved change to source text

I think they are equivalent in terms of logic done, especially in the actual calculate_job we are using update_column calls to avoid callbacks. But would like an extra set of eyes just to be sure I'm not missing anything

benwbrum: This looks much cleaner to me.  I'm approving and will merge.

PR #5301, opened 2026-02-18
5276 - Subjects listing optimizations

### LINKED ISSUE #5276: Subject listing for large document set shows "Content Missing"
While we are able to display subjects for the civil-war-governors-of-kentucky collection, the listing of [People subjects in the Migrated from Mashbill document set](https://fromthepage.com/khs/migrated-from-mashbill/subjects?selected_category_id=1855) attempts to load, then displays "Content missing".  This is probably due to a performance problem, since the calculation for which subjects to display in a document set is costly.

WillNigel23:


benwbrum:
I just tested this and it's blazing fast!

PR #5326, opened 2026-03-03
add gemini 3.1 option

saracarl:


lib/gemini/text_transcriber.rb
```diff
@@ -8,11 +8,13 @@ class TextTranscriber
     # Add custom handling here if the model you are using
     # does not use `v1`
     VERSION_MAP = {
-      'gemini-3-pro-preview' => 'v1beta'
+      'gemini-3-pro-preview' => 'v1beta',
+      'gemini-3.1-pro-preview' => 'vibeta'
```
WillNigel23: Typo at `vibeta` here @saracarl

WillNigel23 (changes requested):

PR #5328, opened 2026-03-03
5315 - Show different gemini results for same page

### LINKED ISSUE #5315: Show different Gemini results for the same page
We are running into scenarios in which we need to experiment running Gemini on the same page with different models or different prompts.  To support this, I suggest this:

After #5189 is merged, I would like to add the following functionality to the AI tab:

If there is more than one `ai_transcription` record for `@page`, we should change the display of the model and run date to a selector, defaulting to the most recent `ai_transcription` (or the one used by the `ai_transcription` scope which will be used when a user clicks _AI Draft_.

When a user selects a different `ai_transcription` from the default, we should switch the AI transcription display, reasoning, and prompt to pull from that record.

If there is only one `ai_transcription` record for a page, the page should display normally without a picker.

WillNigel23:


saracarl (changes requested):
This works great.  One small request:  can we make the dropdown a bit wider so our standard information fits within it without being cut off?

Image

WillNigel23:
It should be responsive now @saracarl

PR #5330, opened 2026-03-04
5134 - Only 1 email on metadata import csv

### LINKED ISSUE #5134: 1 email per retry on metadata upload
If your metadata upload fails, we retry 5 times in the queueing system.  However... that results in 5 emails with failure messages in them.

Can we only send the email once, and only if the 5th one fails?

WillNigel23:


benwbrum:
I don't understand why there are not any checks here.

WillNigel23:
@benwbrum yesterday during gh action outtage it did not run, together with other PRs. The others, however, after a while had their actions started eventually. It seems this oje never recovered 🫠

PR #5331, opened 2026-03-04
5193 - Page caching prototype

### LINKED ISSUE #5193: Investigate page caching
A lot of our bot traffic is hitting read-only, rarely-changing pages like `article#show`.  If we were able to cache these--either through template caching or using a CloudFlare service--it could really drop the load on our servers.

WillNigel23:
Leveraging browser's built-in etag caching and integrating it in our controllers by making sure no relevant record queries are changed so that we don't have to re-render pages unless necessary.

`page_caching` concern computes the etag of records based on their `updated_at` field. Calculating the etag based on that should be optimized in the db by default.

Rails provides a method `stale?` which checks if browsers sent a request header `If-None-Match`, and matches it with the computed cache_key. If there is a match, we can do an early return so that anything below the initial queries will not have to be recomputed, and instead we return. Otherwise, we proceed and recompute the rest and render the html.
```
    return if cache_fresh?(etags: [@user, collections, sets, @tag, @deeds])
   # To test, add a binding.pry here. Whenever cache is fresh, the binding.pry will not run and the browser will return 304
```

So the idea is, our controller logic will now be separated in half by the `return if cache_fresh?` call:
First part:
  - Base queries. These will be scopes and SQL queries that will NOT be loaded into rails object/memory yet. We just want to compute the etag for now

Second part:
  - Loading the queries into memory and then rendering the page. This will only run unless the cache is fresh.
  
```
def show
  # First part
  @users = User.all # this is the base query. This will not load to object yet.
  
  return if cache_fresh?(etag: [@users]) # Inside, a query to compute all updated_at for users is done.
  
  # Second part
  @user_ids = @users.pluck(:id) # This now loads the query to object/memory.
  @user_names = @users.pluck(:display_names) # This now loads the query to object/memory.
  @users.each do |user|
    ... # this will load the query to object/memory
    # this caching strategy will also help detect N+1 issues as usually when that happens, the records are pre-loaded into memories already, which we are purposefully trying to avoid before the return if statement.
  end
end
```
  
  
We are adding (SolidCache)[https://github.com/rails/solid_cache], a db-backed caching with deep integration to modern rails ecosystem. Although this is not used in HTTP caching, I have plans of utilizing this in some caches (say, caching an array of options for a select dropdown, caching names for all contributors, etc).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'interactor-rails', '~> 2.0'
 gem 'friendly_id'
 gem 'thredded', '~> 1.1'
 gem 'solid_queue', '~> 1.2'
+gem 'solid_cache'
 gem 'mission_control-jobs'
 gem 'aws-sdk-s3', require: false
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -470,7 +470,6 @@ GEM
     mini_magick (5.3.1)
       logger
     mini_mime (1.1.5)
-    mini_portile2 (2.8.9)
     minitest (6.0.1)
       prism (~> 1.5)
     mission_control-jobs (1.1.0)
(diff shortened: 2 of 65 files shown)


WillNigel23:
Setting this to draft.

The strategy used here was multi-db setup for setting solid_cache. But #5500 requires us to use solid_cache and after some thought it would be best to do single db setup.

PR #5332, opened 2026-03-04
Warn changed link files

WillNigel23:


WillNigel23:
@benwbrum @saracarl 

I added a newline in database.yml to simulate that this file changed, and we got the `CI / Warn on linked files changes` red in this test. This will at least warn us that we are modifying files that are linked/manually patched on deploys to avoid future unexpected behavior.

PR #5333, opened 2026-03-05
5323 - Numeric slug problem

### LINKED ISSUE #5323: subjects (and many other) tabs links to wrong collection
If you start at [this collection](https://fromthepage.com/fdp/1-omekas-documents-ominibus-collection-private) (need to log in as Frederick Douglass Papers to see it) and click on the "subjects" tab the link is this:
https://fromthepage.com/fdp/1-omekas-documents-ominibus-collection-private/subjects
But it shows you the subjects for the Julia Brumfield Diaries:

Image

The statistics, review, export, metadata fields, sets, facets, and works tabs also take you to the Julia Brumfield Diaries.
Collaborators and Settings link you to the right places.

benwbrum:
Closes #5323

### DIFF
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -769,17 +769,7 @@ class CollectionController < ApplicationController
   end
 
   def set_collection
-    unless @collection
-      if Collection.friendly.exists?(params[:id])
-        @collection = Collection.friendly.find(params[:id])
-      elsif DocumentSet.friendly.exists?(params[:id])
-        @collection = DocumentSet.friendly.find(params[:id])
-      elsif !DocumentSet.find_by(slug: params[:id]).nil?
-        @collection = DocumentSet.find_by(slug: params[:id])
-      elsif !Collection.find_by(slug: params[:id]).nil?
-        @collection = Collection.find_by(slug: params[:id])
-      end
-    end
+    @collection ||= Collection::Lib::SetFriendlyFind.perform(id: params[:id])
   end
 
   def set_collection_for_work(collection, work)
--- a/app/interactors/collection/lib/set_friendly_find.rb
+++ b/app/interactors/collection/lib/set_friendly_find.rb
@@ -1,7 +1,6 @@
 class Collection::Lib::SetFriendlyFind
   def self.perform(id:, work_reference: nil)
-    collection = Collection.friendly.find(id) rescue nil
-    collection ||= DocumentSet.friendly.find(id) rescue nil
+    collection = Collection.friendly.find(id, allow_nil: true) || DocumentSet.friendly.find(id, allow_nil: true)
 
     return collection unless work_reference
 
--- a/app/interactors/collection/lib/set_friendly_find.rb
+++ b/app/interactors/collection/lib/set_friendly_find.rb
@@ -1,10 +1,7 @@
 class Collection::Lib::SetFriendlyFind
   def self.perform(id:, work_reference: nil)
-    collection =
-      Collection.friendly.find_by(id: id) ||
-      DocumentSet.friendly.find_by(id: id) ||
-      DocumentSet.find_by(slug: id) ||
-      Collection.find_by(slug: id)
+    collection = Collection.friendly.find(id) rescue nil
+    collection ||= DocumentSet.friendly.find(id) rescue nil
 
     return collection unless work_reference
 
--- a/lib/tasks/gemini_transcription.rake
+++ b/lib/tasks/gemini_transcription.rake
@@ -129,15 +129,7 @@ namespace :fromthepage do
 
       # Find collection by ID or slug
       id = args.collection_slug
-      if Collection.friendly.exists?(id)
-        collection = Collection.friendly.find(id)
-      elsif DocumentSet.friendly.exists?(id)
-        collection = DocumentSet.friendly.find(id)
-      elsif !DocumentSet.find_by(slug: id).nil?
-        collection = DocumentSet.find_by(slug: id)
-      elsif !Collection.find_by(slug: id).nil?
-        collection = Collection.find_by(slug: id)
-      end
+      collection = Collection::Lib::SetFriendlyFind.perform(id: id)
 
       raise 'Collection does not exist' if collection.nil?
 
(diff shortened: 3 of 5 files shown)


WillNigel23 (approved):
LGTM, but I don't really understand the problem we are trying to duplicate. Can you explain more?

WillNigel23:
Pushed fix.

app/controllers/collection_controller.rb
```diff
@@ -769,17 +769,7 @@ def review_authorized?
...
-        @collection = Collection.friendly.find(params[:id])
-      elsif DocumentSet.friendly.exists?(params[:id])
-        @collection = DocumentSet.friendly.find(params[:id])
-      elsif !DocumentSet.find_by(slug: params[:id]).nil?
-        @collection = DocumentSet.find_by(slug: params[:id])
-      elsif !Collection.find_by(slug: params[:id]).nil?
-        @collection = Collection.find_by(slug: params[:id])
-      end
-    end
+    @collection ||= Collection::Lib::SetFriendlyFind.perform(id: params[:id])
```
benwbrum, line 772: Great!

lib/tasks/gemini_transcription.rake
```diff
@@ -129,15 +129,7 @@ namespace :fromthepage do
...
-      if Collection.friendly.exists?(id)
-        collection = Collection.friendly.find(id)
-      elsif DocumentSet.friendly.exists?(id)
-        collection = DocumentSet.friendly.find(id)
-      elsif !DocumentSet.find_by(slug: id).nil?
-        collection = DocumentSet.find_by(slug: id)
-      elsif !Collection.find_by(slug: id).nil?
-        collection = Collection.find_by(slug: id)
-      end
+      collection = Collection::Lib::SetFriendlyFind.perform(id: id)
```
benwbrum, line 132: Good catch.

app/interactors/collection/lib/set_friendly_find.rb
```diff
@@ -1,10 +1,7 @@
 class Collection::Lib::SetFriendlyFind
   def self.perform(id:, work_reference: nil)
-    collection =
-      Collection.friendly.find_by(id: id) ||
-      DocumentSet.friendly.find_by(id: id) ||
-      DocumentSet.find_by(slug: id) ||
-      Collection.find_by(slug: id)
+    collection = Collection.friendly.find(id) rescue nil
+    collection ||= DocumentSet.friendly.find(id) rescue nil
```
benwbrum, line 4: I'm comparing this against my implementation (which I stupidly did not check in):
```ruby
    # only consult slug if the ID is non-numeric so we don't have conflicts with slugs containing numbers
    if id.to_s.match?(/\D/)
      collection=Collection.find_by(slug: id) || DocumentSet.find_by(slug: id)
    else
      collection =
        Collection.friendly.find_by(id: id) ||
        DocumentSet.friendly.find_by(id: id) ||
        Collection.find_by(slug: id) ||
        DocumentSet.find_by(slug: id)
    end
  ```

**Review by benwbrum — COMMENTED** [review]
This looks really good.  I would like to understand the logic behind `friendly.find` a little better, though -- especially in the case of numeric slugs, integers, and slugs beginning with numbers.

File: app/interactors/collection/lib/set_friendly_find.rb (hunk 3 above)
Comment (WillNigel23) [reply app/interactors/collection/lib/set_friendly_find.rb:4]: This is actually the initial fix I have in mind. But upon further research I think `friendly` gem is already geared to handle these kinds of scenario the `.find` function, so what this does is just repeat the solution without leveraging the gem.

The bug is in fact, introduced by my change due to the wrong assumption that `friendly` gem also patches the `find_by` function, which it does not apparently. I missed this in my initial design for this lib :trollface:

Comment (WillNigel23) [reply app/interactors/collection/lib/set_friendly_find.rb:4]:
```
    # Finds a record using the given id.
    #
    # If the id is "unfriendly", it will call the original find method.
    # If the id is a numeric string like '123' it will first look for a friendly
    # id matching '123' and then fall back to looking for a record with the
    # numeric id '123'.
    #
    # @param [Boolean] allow_nil (default: false)
    # Use allow_nil: true if you'd like the finder to return nil instead of
    # raising ActivRecord::RecordNotFound
    #
    # ### Example
    #
    #     MyModel.friendly.find("bad-slug")
    #     #=> raise ActiveRecord::RecordNotFound
    #
    #     MyModel.friendly.find("bad-slug", allow_nil: true)
    #     #=> nil
    #
    # Since FriendlyId 5.0, if the id is a nonnumeric string like '123-foo' it
    # will *only* search by friendly id and not fall back to the regular find
    # method.
    #
    # If you want to search only by the friendly id, use {#find_by_friendly_id}.
    # @raise ActiveRecord::RecordNotFound
 ```
 
 https://github.com/norman/friendly_id/blob/cf4b81c8024c8b482c032a2d987a8065bab4b585/lib/friendly_id/finder_methods.rb#L22

app/interactors/collection/lib/set_friendly_find.rb
```diff
@@ -1,7 +1,6 @@
 class Collection::Lib::SetFriendlyFind
   def self.perform(id:, work_reference: nil)
-    collection = Collection.friendly.find(id) rescue nil
-    collection ||= DocumentSet.friendly.find(id) rescue nil
```
WillNigel23, line 4: Checking the gem implementation, we can pass `allow_nil: true` to silence errors on find here.

```
    collection = Collection.friendly.find(id, allow_nil: true) || DocumentSet.friendly.find(id, allow_nil: true)
```

The past implementation was this:
```
-    if Collection.friendly.exists?(id)
-      @collection = Collection.friendly.find(id)
-    elsif DocumentSet.friendly.exists?(id)
-      @collection = DocumentSet.friendly.find(id)
-    elsif !DocumentSet.find_by(slug: id).nil?
-      @collection = DocumentSet.find_by(slug: id)
-    elsif !Collection.find_by(slug: id).nil?
-      @collection = Collection.find_by(slug: id)
-    end
-
-    # check to make sure URLs haven't gotten scrambled
-    if @work
-      if @work.collection != @collection
-        # this could be a document set or a bad collection
-        unless @collection.is_a? DocumentSet
-          @collection = @work.collection
-        end
-      end
-    end
-    @collection
```

where we check first if it exists before doing the actual find, which is going to do at least twice the amount of queries made

PR #5343, opened 2026-03-09
5309 - Solid queue health check

### LINKED ISSUE #5309: solidqueue not running
I'm trying to debug a metadata upload problem for a customer, and I've concluded it's because we don't have a solidqueue worker running.
https://fromthepage.com/solid_queue/applications/fromthepage/queues/default?page=29&server_id=solid_queue
and haven't for 5 days:
https://fromthepage.com/solid_queue/applications/fromthepage/queues/default?server_id=solid_queue

So, first question is "how do we restart it"? It's a service, right? -- rather than the test instructions of 
`nohup bundle exec rake solid_queue:start >> log/worker.log 2>&1 &`

Second question is "will it catch up?"

Third question is "how can we monitor things so this doesn't happen in the future?"

WillNigel23:
bundle exec rake 'fromthepage:workers:health_check'

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,7 +169,7 @@ jobs:
   linked_files_warn:
     name: "Warn on linked files changes"
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'push'
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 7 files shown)


saracarl (changes requested):
Can we modify the rake task so it only sends a message when it fails?

otherwise, someone has to read the [#solid-queue-health-check](https://fromthepage.slack.com/archives/C0AKG9M92FL) channel every day/hour/whatever and no one is going to do that.

PR #5346, opened 2026-03-10
5322/5338/5337/5339 - Pdf export issues

### LINKED ISSUE #5322: Facing edition PDF switch image/transcript order part way through
If you open this PDF: [James_C_Nealon_facing_edition.pdf](https://github.com/user-attachments/files/25696291/James_C_Nealon_facing_edition.pdf) you can see that while page 33 and 34 are in the correct order -- the image appears on 33 and the transcription on 34 -- page 35 and 36 are switched -- the transcription comes first and the image comes second.  It swaps spots again on page 41, then again on 43, 51, etc.

Another way to look at it -- it's in the wrong order (transcript than image) in four different spots: Pages 35-40, 42-50, 69-78 and 89-92.

Here's a link to the exported collection:  https://www.fromthepage.com/castatelibrary/early-california/james-c-nealon-letters

### LINKED ISSUE #5337: pdf exports don't maintain empty lines
The postcard transcription on [this page](https://fromthepage.com/https-www-terencemckenna-com/terence-mckenna-notes-and-journals/postcards-from-terence-to-his-mom-june-sep-1970/display/35345118) has a number of blank lines for formatting.  Those blank lines are stripped out in the face edition pdf export.

[Post-cards_facing_edition.pdf](https://github.com/user-attachments/files/25802710/Post-cards_facing_edition.pdf)
Image

### LINKED ISSUE #5338: rotated images aren't preserved in pdf export
A work page like [this one](https://fromthepage.com/https-www-terencemckenna-com/terence-mckenna-notes-and-journals/postcards-from-terence-to-his-mom-june-sep-1970/display/35345118)  shows a postcard image.  I'm pretty sure the image was rotated in the page settings, because the thumbnail still has the wrong rotation:

Image

Image

We show the unrotated in the PDF export instead of showing the corrected rotation:
[Post-cards_facing_edition.pdf](https://github.com/user-attachments/files/25802738/Post-cards_facing_edition.pdf)

### LINKED ISSUE #5339: "Figure 1:" subtitles in pdf exports
Images in PDF exports are subtitled "Figure 1:" (etc) under the image.  That is not any value from the actual uploaded work, but something we're inserting.  We shouldn't.  The page title is all that's needed.  (Although we should ensure that the page title shows up in the right places after fixing the ordering issue.)

[Post-cards_facing_edition.pdf](https://github.com/user-attachments/files/25802801/Post-cards_facing_edition.pdf)

WillNigel23:


app/helpers/export_helper.rb
```diff
@@ -179,8 +179,10 @@ def write_work_exports(works, out, export_user, bulk_export)
         include_metadata = bulk_export.report_arguments['include_metadata'] != '0'
         include_contributors = bulk_export.report_arguments['include_contributors'] != '0'
         include_notes = bulk_export.report_arguments['include_notes'] != '0'
+
         if bulk_export.facing_edition_work
-          export_printable_to_zip(work, 'facing', 'pdf', out, by_work, original_filenames, preserve_lb, include_metadata, include_contributors, include_notes)
+          # NOTE: Facing editions should always preserve_lb
+          export_printable_to_zip(work, 'facing', 'pdf', out, by_work, original_filenames, true, include_metadata, include_contributors, include_notes)
```
WillNigel23, line 185: https://github.com/benwbrum/fromthepage/issues/5337 is fixed by this line

app/views/export/tex.html.erb
```diff
@@ -1,10 +1,8 @@
 <% @pages.each do |page| %>
   <% if @edition == 'facing' %>
-    \begin{figure}[h]
-      \centering
-      \includegraphics[width=\linewidth]{<%= page.image_url_for_download %>}
-      \label{fig:<%= page.title %>}
-    \end{figure}
+    \begin{center}
+      \includegraphics[width=\textwidth,height=\textheight,keepaspectratio]{<%= page.image_url_for_download %>}
+    \end{center}
```
WillNigel23, line 5: We were previously using `\begin{figure}[h]`. And I just discovered that figures are positioned latex by deciding at which position/order it would look better. So the initial assumption of the image size affecting the order is partially true.

That being said, this should now enforce order.

Bonus, it removes the `Figure` annotation that we want removed from https://github.com/benwbrum/fromthepage/issues/5339

benwbrum: That's fascinating!  Sounds like it might be an artifact of our old pandoc MD-to-HTML-to-LaTex code.

saracarl (changes requested):
I tried this on my test server with this document:
http://localhost:3000/benwbrum/jeremiah-white-graves-diaries/jeremiah-white-graves-diary-volume-2-book-10
which has IIIF loaded images from the internet archive (I think IIIF -- maybe double check that).  

The "download" tab direct download of a facing PDF has *no images*!  

So....  needs some changes.

WillNigel23:
@saracarl I think the direct download tab here was never meant to export facing edition? At least checking the git blame on `export_controller`.

I tried downloading the linked work on the bulk export menu selecting `Facing Edition` and the pdf included images.

saracarl (changes requested):
Missing images were user error.

However, while line breaks are being respected, paragraph breaks (double returns) are not.

Read view:

Image

PDF view:

Image

Also, @benwbrum can you confirm if facing PDFs should have index links/articles if they are indexed?  (This doesn't.)

WillNigel23:
> Also, @benwbrum can you confirm if facing PDFs should have index links/articles if they are indexed? (This doesn't.)

I checked the latex generated, it does index it. but for some reason it does not appear as a link in the pdf view. Maybe just a styling issue. I'm checking.

WillNigel23:
I found the setting here that we seem to always be setting `flatten_links` true in pdf export. I think we are able to link it to fromthepage, should we do that in pdf exports?

WillNigel23:
@saracarl I'm investigating more on the links not having proper styling at all, and it led me down another rabbithole. I think we should fix that in a separate [issue](https://github.com/benwbrum/fromthepage/issues/5359).

app/interactors/work/export/lib/utils.rb
```diff
@@ -55,7 +55,7 @@ def self.process_element(page, element, preserve_lb, flatten_links)
 
     case element.name
     when 'p'
-      "#{content}#{LINEBREAK_ELEMENT}"
+      "#{content}#{LINEBREAK_ELEMENT}#{preserve_lb ? LINEBREAK_ELEMENT : ''}"
```
WillNigel23: We are adding 1 linebreak per `<p>` tag by default. If we choose `preserve_lb` true, I believe we should break twice to simulate a `paragraph break`

saracarl (changes requested):
two issues:

1)  I see this encoding in the text:
Image

2)  So there's a lot of blank pages after the text pages.  I'm wondering if there's an extra line at the end that is pushing it to a second page?  (see page 5, for instance).
[facing_edition.pdf](https://github.com/user-attachments/files/26067592/facing_edition.pdf)

WillNigel23:
@saracarl 

> 1. I see this encoding in the text:

I found out that this was a bug. Page content was this:
```
"1855 Augt 23\\textsuperscript{rd}"
```

And our parser escapes it to this, which is wrong.
```
1855 Augt 23\textbackslash{}textsuperscript\{rd\}
```
Fixed this issue in my latest commit.

> So there's a lot of blank pages after the text pages. I'm wondering if there's an extra line at the end that is pushing it to a second page? (see page 5, for instance).

I checked the pages where these blank pages happened. (pages with id: [504360, 504365, 504369] to name a few) All of them had an empty space AFTER the last line (see attached picture).

image

Should we be handling cases like these? I believe if we add empty spaces at the end then it should add those spaces. So although this visually looks wrong, it is functionally correct and could be attributed to user error (`error` for lack of a better term)

saracarl (changes requested):
OK.  This is getting closer.
1)  Checked against prod, we don't do index links in facing edition PDFs, so we're ok there.
2)  Encoding errors look good.
3)  I think we should probably trim any whitespace/extra lines from the bottom of a page of text.  Yes, it might be user error, but we'll get complaints about it so let's just "head it off at the pass".

saracarl (changes requested):
Page 5 and page 20 are still blank...

[facing_edition.pdf](https://github.com/user-attachments/files/26197273/facing_edition.pdf)

PR #5358, opened 2026-03-16
5338 - Fix latex rotated image bug

WillNigel23:


app/models/page.rb
```diff
@@ -652,23 +652,50 @@ def image_url_for_download
...
-        uri.port = port
-      else
-        uri.scheme = 'https'
-      end
-      uri.to_s
+      image_url_from_web_path(self.canonical_facsimile_url)
+    end
+  end
+
+  def normalized_image_url_for_download
```
WillNigel23, line 659: @saracarl @benwbrum 

The idea is we normalize images so that the `exif` metadata is not affecting the pdf export.
Admittedly this will be slower. But given that we are only exporting images when it is `facing` edition, and that all `facing` edition exports happen async, it should be fine.

I'll look into ways of improving this.

saracarl: It didn't seem that long for me, but that was only one rotated image.

saracarl (approved):

PR #5373, opened 2026-03-24
5361 - Error 500 on article view

### LINKED ISSUE #5361: 500 error on migrated mashbill subject
This article URL results in a 500 error:
https://fromthepage.com/khs/migrated-from-mashbill/article/32215211

```
I, [2026-03-18T14:35:27.699772 #424853]  INFO -- : Completed 500 Internal Server Error in 3664ms (ActiveRecord: 483.9ms (36 queries, 9 cached) | GC: 793.8ms)
F, [2026-03-18T14:35:27.701978 #424853] FATAL -- :   
ActionView::Template::Error (undefined method `title' for nil):

Causes:
NoMethodError (undefined method `title' for nil)
    20:     -@article.target_article_links.each do |link|
    21:       li
    22:         =svg_symbol '#icon-page', class: 'icon'
    23:         =link_to link.source_article.title, collection_article_show_path(@collection.owner, @collection, link.source_article.id)
    24:         =="&nbsp;&mdash;&nbsp;#{link.display_text}"
  
app/views/article/_article_links.html.slim:23
app/views/article/_article_links.html.slim:20
app/views/article/show.html.slim:55

```

It feels like there's a nil linked article (in the sidebar?  in the text of the article?).  I suspect the sidebar, since the is a successful "hover" article.

Image

WillNigel23:


WillNigel23:
@benwbrum @saracarl 

I can't replicate this locally with dump since we are not exporting `ArticleArticleLink` table. If we want true replication I suggest we merge this one first: https://github.com/benwbrum/fromthepage/pull/5372

But I think my guess here is right, the issue was an `ArticleArticleLink` record with missing `source_article` record. Artificially creating this in my local dev and I got a similar error:
image

How this record came to be is still a mystery to me. But it could be deletion, renaming, etc. In any case I think this will fix the error in prod.

PR #5394, opened 2026-03-31
Post bulk export worker issue

WillNigel23:


app/interactors/workers/env_check.rb
```diff
@@ -0,0 +1,50 @@
+# :nocov:
+class Workers::EnvCheck < Workers::Base
+  REQUIRED_VARS = [
```
WillNigel23, line 3: This list is sure to grow. But this will notify us whenever the envs for `job` context is out of sync with what we are expecting it to be.

app/views/user_mailer/bulk_export_finished.html.erb
```diff
@@ -1,7 +1,7 @@
 <h2 style="color:#453939; margin:0; font-weight:normal;"><%= t('.your_export_is_ready') %></h2>
 <p>
   <%= t('.your_export_of') %>
-  <%= @bulk_export.collection.title %> 
+  <%= @bulk_export.collection&.title %>
```
WillNigel23, line 4: I saw an error in one of the job logs of an SMTP email saying `title` is missing. This is the only place I could think of that it could be, so just to be safe I am adding this.

config/queue.yml
```diff
@@ -3,7 +3,7 @@ default: &default
     - polling_interval: 1
       batch_size: 500
   workers:
-    - queues: ["default"]
+    - queues: ["default", "solid_queue_recurring"]
```
WillNigel23, line 6: I missed adding this, now our solid_queue_recurring job queue is full and not running. Apologies 🙇 

Nothing crucial in that queue, just clearing of old finished job logs.

WillNigel23 (commented):
@benwbrum @saracarl  minor but severely needed PR to save us some headaches in the future

PR #5395, opened 2026-03-31
5370 - Remediate blank ai drafts

### LINKED ISSUE #5370: Blank AI Drafts
SHSU reports:

_Finding random pages in sets that don't have the AI draft radio button - [A Summary of "Religion and Drink" - page 17](https://www.fromthepage.com/ngldigiprojects/collection-of-materials-on-prohibition/a-summary-of-religion-and-drink/transcribe/35633718)_

When I look at that page, I see the AI tab, but it's blank.  The page image itself is not blank.  That implies to me we *thought* we got a transcript back from the AI call, but we actually didn't.  I wonder if there was an error returned -- are we catching those?  If there wasn't an error, I'm not sure how to account for the difference between a truly blank page and no text returned.  (I'm tempted to go back and try all blank AI drafts after a work is done.)

Thoughts on how to handle this, @benwbrum, @WillNigel23  ?

Image

WillNigel23:
`bundle exec rake 'oneoff:remediate_blank_ai_transcription_records'`


lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,7 @@
+namespace :oneoff do
```
WillNigel23, line 1: Let's use `oneoff` namespace for all `rake` tasks that we are expecting to only run once. Anything that we are tempted to just run inside production console basically.

This is to:
1. Add more `eyes` and `reviewers` to tasks we are running. This will prevent non-reversible bugs which could happen if we directly run it from console.
2. Give us footprints on the steps we did in case we want to `reverse` what we just did to some degree.

saracarl: I like the `oneoff` namespace, but I question if this is a one-off task.  Do we have another strategy in mind for noticing and fixing pages where the API never returned a transcript?  A "post work run" check might be good enough.

benwbrum:
@WillNigel23 How will this script behave if there is an `ai_transcription` record that is still waiting to be run, as when we have a thousand or more jobs in the queue?  Will this re-run the transcription for those records (running them twice) or handle the status gracefully?

WillNigel23:
@benwbrum we are adding `where(status: :finished)` in the query so we only target those that have finished the job process.

lib/tasks/oneoff.rake
WillNigel23: You are right, unless we fix the root source this will not be a `oneoff`. That being said, having a `finished` ai transcription where both reasoning and source_text blank is something that should not occur in the first place.

I am adding a check at the end of `generate` interactor so we raise an error and handle it accordingly when these cases occur, setting the statuses to `:error` so we are informed something went wrong.

This way we keep this task a `oneoff`, and ensure no future records get to this invalid state. 

@saracarl

benwbrum:
@copilot resolve the merge conflicts in this pull request

PR #5402, opened 2026-04-01
Add read timeout on image fetch

### LINKED ISSUE #5401: Handle Net::OpenTimeout in image fetching code for ai transcription jobs
We are running into problems fetching images to post to Gemini because we don't handle timeouts gracefully.  This can be external image hosts, or images that we host ourselves and serve through Apache.  There is a good chance that this morning's outage was caused by this; at least we saw a large spike in these errors during that outage.

WillNigel23:


benwbrum:
We now believe that the original problem was a side-effect of a different issue.

PR #5415, opened 2026-04-07
AI transcription job concurrency fix

WillNigel23:


lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,56 @@
+namespace :oneoff do
```
WillNigel23, line 1: NOTE: This was already ran before. We only keep this code here as to `track` that we did run this oneoff task. No need to re-run this again.

benwbrum (approved):
This look good to me.  Fascinating approach!

benwbrum:
@copilot please resolve merge conflicts.

PR #5421, opened 2026-04-08
Address AI transcription starvation part 1

WillNigel23:


WillNigel23:
Small PR. Still working on the batch_api for gemini as there is no support yet for the gem

config/queue.yml
```diff
@@ -8,11 +8,7 @@ default: &default
       processes: <%= ENV.fetch("JOB_CONCURRENCY", 1) %>
       polling_interval: 0.1
     - queues: ["ai_transcriptions"]
-      threads: 3
-      processes: <%= ENV.fetch("JOB_CONCURRENCY", 1) %>
-      polling_interval: 0.1
-    - queues: ["bulk_exports"]
```
benwbrum, line 14: Why remove the separate queue for bulk exports?

WillNigel23: We reverted back bulk export to use rake tasks. Until we are fully confident it is working on the right context we can remove this for now.

benwbrum: Right!  That makes sense.

PR #5424, opened 2026-04-09
5423 - Image on top transcription panel layout bug

### LINKED ISSUE #5423: "image on top" transcription pane has line through it and page layout is problematic
From Jason at WWP:

I have noticed since yesterday that when I am in the transcribe view and scroll down the page, the image is no longer "contained" in its own box and is now going down into the transcription box. The image is now covering the shortcut icons as well, and I have to scroll up to find them. I'm not sure if this is just something I am experiencing, but I was wondering if you could please take a look? 

Image

And Meredith at ADAH:

Good morning! One of our long-time volunteers has encountered an issue with the “Image at top” page layout (see her message below). The display is wonky for me as well, except that I can’t even see the transcription area at all—just a line where it looks like I should be allowed to expand the viewing area (though nothing happens when I click on it):

 

Image

 

We’ve both tried multiple browsers but can’t figure out how to get around it. Would you mind taking a look and letting us know?

and her transcriber:

Did something change with the transcription pages? I am working on this one: https://fromthepage.com/adah/alabama-supreme-court-case-files/phillips-administrator-v-phillips/transcribe/35159670. I usually work using "Image at top". Now it is not anchoring the transcription section, so I cannot see the strike, sub, sup, etc. buttons once I pass the 15th line or so.

 

I have tried more than one browser, but I see the same behaviour in both. I was wonering if there was a recent change, and whether there might be some setting I need to change on my end.

Image

I'm pretty sure this was caused by this PR:

https://github.com/benwbrum/fromthepage/pull/5202

WillNigel23:


saracarl:
This change regresses the original change that added over/under image in the overview tab.  We should have a slider between the image and the transcription, and we don't:

image

The slider moved to underneath the transcription, but it doesn't do anything:

image

benwbrum:
This does, in fact, fix the regression in #5423

PR #5428, opened 2026-04-13
Remediate Duplicate Jobs

WillNigel23:


lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
+namespace :oneoff do
+  desc 'Remediate duplicate and dead jobs'
+  task remediate_duplicate_jobs: :environment do
+    active_process_pids = [
```
WillNigel23: We can adjust this if it changes.

`ps aux | grep 'solid'` seems to do the trick and show us solid_queue processes.

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+
+      # DISPATCHERS
+      2074922
+    ]
+
+    active_processes = SolidQueue::Process
+      .where(id: active_process_pids)
+
+    stale_processes = SolidQueue::Process
+      .where.not(id: active_processes.pluck(:id))
```
WillNigel23, line 13: I compared in mission_control dashboard and the result of `ps aux | grep 'solid'` and the PIDs of the stale and active workers are accurate.

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+
+    stale_claimed = SolidQueue::ClaimedExecution
+      .where(process_id: stale_processes.pluck(:id))
```
WillNigel23, line 16: Upon reading more, it seems aside from a `Job` object, the state of solid_queue jobs are determined by separate tables

solid_queue_blocked_executions
solid_queue_paused_executions
solid_queue_ready_executions
solid_queue_claimed_executions

All of which references a `job` object `SolidQueue::Job`

The mis-match in count is attributed to our remediation last week, where we just deleted the `Job` records without remediating the state tables. What we should have done instead of `deleting` the records was to `create` the state record for `failed_executions` to keep the counts accurate.

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+        next
+      end
+
+      args = job.arguments
+      payload = args['arguments']&.first || {}
+
+      case job.class_name
+      when 'BulkExport::ProcessJob'
+        bulk_export_id = payload['bulk_export_id']
+        record = BulkExport.find_by(id: bulk_export_id)
```
WillNigel23, line 34: It would be good to at least set the correct status for abandoned jobs

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+        end
+      when 'AiTranscription::GenerateJob'
+        ai_id = payload['ai_transcription_id']
+        record = AiTranscription.find_by(id: ai_id)
+
+        if record&.status.in?(['new', 'processing'])
+          record.update!(status: 'error')
+        end
+      else
+        next
```
WillNigel23, line 47: Aside from AiTranscription and BulkExport, I don't think we run other jobs that holds the state. 

Article::RenameJob I believe can be retried later once records are set. Other than that we can just chalk to `failure` and move on.

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+      SolidQueue::FailedExecution.create!(
+        job_id: job.id,
+        error: 'Reconciled as failed due to stale process cleanup',
+        created_at: Time.current
+      )
+
+      claim.delete
+    end
+
+    stale_processes.delete_all
```
WillNigel23, line 61: Our deletion last week created these stale processes one way or another. Here's to deleting their records on the DB since they don't have matching PID anyway.

lib/tasks/oneoff.rake
benwbrum: I would love to have this parsed from a system call rather than hard-coded.  If you worry about false positives, we could also pass the PIDs in from a parameter (with some good documentation on what to run to find the PIDs)

lib/tasks/oneoff.rake
benwbrum: Awesome!

lib/tasks/oneoff.rake
```diff
@@ -0,0 +1,70 @@
...
+    stale_claimed = SolidQueue::ClaimedExecution
+      .where(process_id: stale_processes.pluck(:id))
+
+    jobs = SolidQueue::Job.where(id: stale_claimed.pluck(:job_id)).index_by(&:id)
+
+    stale_claimed.find_each do |claim|
+      job = jobs[claim.job_id]
+
+      if job.nil?
+        claim.delete
```
benwbrum, line 24: Do we want `delete` or `destroy` here?

lib/tasks/oneoff.rake
benwbrum: I agree completely.

lib/tasks/oneoff.rake
benwbrum: That sounds correct.

lib/tasks/oneoff.rake
benwbrum: Great!

benwbrum (changes requested):
Let's talk about parameterizing the PIDs

lib/tasks/oneoff.rake
WillNigel23: Delete should be fine

PR #5444, opened 2026-04-20
5261 - Wide table grover pdf export

WillNigel23:


WillNigel23:
@benwbrum we will need to have `npm` here to install a js dependency `puppeeteer`.

benwbrum (approved):
This looks really good.  It's ready to test.

saracarl (changes requested):
This cuts off wide tables at the width of the page, losing substantial amounts of data.

[2_box_98_176df7d9_0d2f_4f6d_a956_1247110441e9_20260428174809.pdf](https://github.com/user-attachments/files/27176208/2_box_98_176df7d9_0d2f_4f6d_a956_1247110441e9_20260428174809.pdf)

saracarl:
This looks the same as the current production code, so I don't see that it would be worthwhile to introduce a dependency,  Also, nothing prompted me to install puppeteer.

WillNigel23:
image

Grover pdf export is under the `Web-based` option. I believe I added it as well on bulk_export, but for faster testing we can try in the work `Download` tab right under the `printables` section on top.

It would fail within html-request cycle if we do not have puppeteer in the system.

WillNigel23:
@benwbrum @saracarl I'm lowering the stakes for this one even further.

Enabling only for US admin users. If we can get this deployed and puppeteer set up on prod, I'd love to test this against latex exports that are failing/going to fail and see if this could help solve all our latex issues at hand.

We need node and npm install puppeteer for this PR to work. I added these to use this grover PDF export:
image
image

saracarl:
Tested this on one of our trickier documents.  Received the following error:

```
"html_facing_edition_work"=>true}

	2026-07-29T14:02:13-05:00 Exporting work	25000106	BSB_Cod._Gall._51_Livre_au_Roi

	[MEM] START WORK 25000106 RSS=243.6MB

/home/saracarl/dev/products/fromthepage/fromthepage/app/views/export/grover_pdf.html.slim:23:in `block (2 levels) in _app_views_export_grover_pdf_html_slim__3175673617802973502_38240': undefined local variable or method `note' for an instance of #<Class:0x00007f08ecfe29e0> (ActionView::Template::Error)
	from /home/saracarl/.rvm/gems/ruby-3.3.5/gems/actionview-7.2.3.1/lib/action_view/helpers/translation_helper.rb:108:in `translate'
	from /home/saracarl/dev/products/fromthepage/fromthepage/app/views/export/grover_pdf.html.slim:21:in `block in _app_views_export_grover_pdf_html_slim__3175673617802973502_38240'

```

This is an old Fordham document in the French of Outremeyer collection; it's probably in your test database, @WillNigel23 

Tried a Graves diary and got the same error.  Exporting work	14647	Diary of Jeremiah White Graves, Volume 2, Book 10

saracarl:
This is looking good.  The only problem I found is that if you have subject links, like in the attached, they are not clickable.

[html_facing_edition.pdf](https://github.com/user-attachments/files/30942643/html_facing_edition.pdf)

WillNigel23:
@saracarl for PDF exports we usually pass `flatten_links: true`. I'm also curious, why is that? I actually think for grover exports we can have the url link to fromthepage articles, but  ultimately decided to follow the current convention.

saracarl:
If we're going to flatten links, we need to delete them.  Links to nowhere are a terrible UI.  I was expecting them to link back to the FromThePage article, so let's do that, @WillNigel23.

WillNigel23:
I tested it and it works! Article links are clickable in PDF now!

saracarl:
@WillNigel23 I went to test this and the option to download the web rendered PDF has disappeared.  Maybe the force push?

PR #5445, opened 2026-04-21
Fix #5433 by changing institution list logic

### LINKED ISSUE #5433: organization missing from Find A Project
This account:
https://fromthepage.com/lva250
Isn't showing up on the FindAProject screen under institutions.  It's a newer account.  

I suspect it's because it is not indexed in ES, so it isn't available here:

https://github.com/benwbrum/fromthepage/blob/34ada35203e62a0d597218e098525f9b4864ae2f/app/views/dashboard/_results.html.slim#L25

benwbrum:
Closes #5433

This handles the two use cases in #5433:
* Organization that has uploaded all works as a staff member does not appear in institution list because of the scope on uploaded works
* Organization with all private collections but a public document set does not appear in the institution list because of check for collections.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -208,10 +208,10 @@ class DashboardController < ApplicationController
       end
     end
 
-    users = User.owners.joins(:collections).left_outer_joins(:document_sets)
+    users = User.owners.left_outer_joins(:collections).left_outer_joins(:document_sets)
 
-    org_owners = users.findaproject_orgs.with_owner_works
-    individual_owners = users.findaproject_individuals.with_owner_works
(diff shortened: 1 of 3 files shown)


benwbrum:
This will change production by removing the following institutions from the list, since they have no public collections or document sets:
```
["The Parker Library",
 "Carnegie Hall Archives",
 "Wisconsin Historical Society",
 "Yale University Libraries",
 "Wilford Woodruff Papers Foundation",
 "Middlebury College Special Collections",
 "St. John's College ",
 "University of Texas at San Antonio",
 "Ayn Rand Archives",
 "Salus Populi Project",
 "Industrial Archives",
 "Kansas City Monuments Coalition",
 "Union Mills Homestead"] 

```

PR #5450, opened 2026-04-22
5261 wide tables pdf fix

WillNigel23:


saracarl (approved):
This works reasonably well -- we don't cut off any data, which is an improvement from the current export.

PR #5459, opened 2026-04-27
Move dotenv gem to production env

### LINKED ISSUE #5396: Implement dotenv gem for improved environmental variable and secrets management
At present various environmental variables and secrets are managed in a hodgepodge of locations:

- Apache config files in e.g., `/etc/apache2/sites-available/...`
- Capistrano deployment shared files in `/home/fromthepage/deployment/shared`
- the fromthepage user's `.bashrc`

etc.

These include remote storage credentials, environmental variables related to RVM/Passenger settings, and application specific settings.

Adding the [`dotenv` gem](https://github.com/bkeepers/dotenv) to centralize and simplify as many of these variables and secrets into easily manageable, environment based `.env` files would have the additional benefit of improving our ability to troubleshoot problems quicker and identify potentially conflicting variables.

WillNigel23:


Gemfile
```diff
@@ -20,6 +20,8 @@ gem 'thredded', '~> 1.1'
 gem 'solid_queue', '~> 1.2'
 gem 'mission_control-jobs'
 gem 'aws-sdk-s3', require: false
+gem 'dotenv-rails'
+gem 'dotenv', require: 'dotenv/load'
```
WillNigel23, line 24: We need to add a symlinked `.env` inside our `current` deployment dir.
Then we will have to point both apache and the solid_queue service to it.

@benwbrum @akierig

benwbrum: This should be symlinked the same way that we do other config files, right?  In that case we should add it to https://github.com/benwbrum/fromthepage/blob/development/config/deploy.rb#L26

benwbrum (changes requested):
In order for capistrano to work, we also need to check in a `.env` file with dummy or commented values.

WillNigel23:
I pushed `.env.example`, added the `.env` in the deploy linked files, and I created a .env with correct values in the server's `shared` dir.

@benwbrum

PR #5466, opened 2026-04-29
5375 - Work-level AI settings

WillNigel23:


benwbrum:
Notes from testing:

- [ ] There is no modal "are you sure" dialog to confirm moving a work from one collection to another.  I tested this with both a work containing subject links and one that did not
- [ ] Changing the document set of a work generates this error and does not save: `app/controllers/work_controller.rb:227:in 'update' Unpermitted parameter: :document_set_ids`
- [ ] We need more consistent section/visual organization.  We have large, visible titles for the first few fields, then alternate between no header (Pages are Meaningful, OCR Correction) and headers (Transcription Conventions, Translation Instructions).  We need to do some mock-up work on this, and consider separating this screen into Look & Feel / Task Configuration
- [ ] We should disable the Translation Instructions textarea unless Enable Translation is checked
- [ ] Privacy & Access: "Allow Collaborators to Edit Page Titles" does not belong here -- it should be with other task configuration features
- [ ] Privacy & Access: Uploaded By is pretty strange, since it's just informational.  Perhaps we should get rid of it altogether, or put it someplace global for the work settings like underneath the sidebar tabs
- [ ] Privacy & Access: If no collaborators are added, for some reason the message "No Users Scribes" appears.  We don't want to use the term "scribes" in the UI anywhere, and the wording here should be more consistent with the privacy & access screens on the Collection Settings
- [ ] Privacy and Access -- we should probably change the "Allow"/"Restrict" buttons to be checkboxes.  The only reason they were form submits in the old UI was because we did not have an easy way (at the time) to conditionally enable/disable the Add Collaborators control

The AI tab and Danger Zone both look great.

benwbrum:
These problems with the UI still remain to be fixed:
* There is no modal "are you sure" dialog to confirm moving a work from one collection to another. I tested this with both a work containing subject links and one that did not
* Privacy & Access: "Allow Collaborators to Edit Page Titles" does not belong here -- it should be with Task Configuration


New suggestions:
* Let's move the Transcription Conventions to the Task Configuration tab

WillNigel23:
@benwbrum 

image

It does work for me. Can you try restarting server once you switch to this branch? This PR includes changes in stimulus js file and perhaps it needs a server restart to work.

As per the other concern, I moved the setting in the task configurations tab.

benwbrum:
The settings look good, but I still don't see an alert when I change collections, even after restating my server and reloading the page.


https://github.com/user-attachments/assets/526b11ef-2a84-4ad1-a29e-0f2ade61c5a8

PR #5485, opened 2026-05-04
Add full response to gemini handler

WillNigel23:


benwbrum (approved):
This looks really good

PR #5489, opened 2026-05-05
5488 - Exports better logging

### LINKED ISSUE #5488: Add better logging to Exports
STDERR and STDOUT are not always being added to logfiles, making diagnostics hard

WillNigel23:


app/interactors/bulk_export/process.rb
```diff
@@ -51,34 +53,4 @@ def logger
 
     @logger
   end
-
-  def export_to_zip
```
benwbrum, line 55: Was this dead code?

WillNigel23: It is, I thought this was removed in one of my PRs to revert it but either I forgot or there was bad merge.

PR #5494, opened 2026-05-06
5490 - Catch failures properly

### LINKED ISSUE #5490: Catch export failures properly and ensure email happens on successful export only

WillNigel23:


app/jobs/bulk_export/process_job.rb
```diff
@@ -10,6 +10,35 @@ def perform(user_id:, bulk_export_id:)
 
     result = BulkExport::Process.new(bulk_export: bulk_export).call
 
-    raise result.full_errors unless result.success?
+    if result.success? && SMTP_ENABLED
+      begin
+        UserMailer.bulk_export_finished(result.bulk_export).deliver!
+      rescue StandardError => e
```
benwbrum, line 16: Should this catch `StandardError` or absolutely everything via `rescue => e`?  (I don't quite remember the difference here.)

app/jobs/bulk_export/process_job.rb
```diff
@@ -10,6 +10,35 @@ def perform(user_id:, bulk_export_id:)
...
+      error_message =
+        if error.respond_to?(:backtrace)
+          "#{error.message}\n#{error.backtrace.join("\n")}"
+        else
+          error.to_s
+        end
+
+      path = result.bulk_export.log_file
+
+      if path.present?
```
benwbrum, line 32: Will this always return true, since the we're testing the path string's existence rather than the file's existence?

benwbrum (commented):
I have a couple of questions -- possibly silly ones.

app/jobs/bulk_export/process_job.rb
WillNigel23: Usually mailer can go just catch standard error

We generally want to avoid catching everything, as it can prevent the process from dying. That is why linters flag a catchall without standarderror, since it also tries to catch other lower level ruby errors

app/jobs/bulk_export/process_job.rb
WillNigel23: Yes, only testing the string's existence here. File shluld be created already at that point

PR #5505, opened 2026-05-11
5504 - Process bulk export artifacts individually

### LINKED ISSUE #5504: Process bulk export artifacts individually
https://fromthepage.com/solid_queue/applications/fromthepage/jobs/6c808b52-43f4-4e0e-911d-37f16bc66b58?clean_backtrace=false&server_id=solid_queue
```
Type 	SolidQueue::Processes::ProcessPrunedError
Message 	Process was found dead and pruned (last heartbeat at: 2026-05-07 23:38:26 UTC) 
```

This is consistent with behavior with running the bulk export rake task. It just dies , which likely means this error is not a code error.

I'm thinking the export is too large we are running into memory issues here (not storage). 

Probably zipping all those work exports in-memory is causing the process to die. There are no ruby exception happening which supports my theory here

I think the next step is to slightly change the process.
1. Generate export artifacts individually.
2. Zip them in one go

WillNigel23:


app/models/bulk_export.rb
```diff
@@ -107,8 +107,13 @@ def downloadable?
 
   def clean_zip_file
     output.purge
+
     File.unlink(zip_file_name) if File.exist?(zip_file_name)
     File.unlink(log_file) if File.exist?(log_file)
+
+    artifacts_dir = Rails.root.join("tmp/fromthepage_exports/export_#{id}")
```
benwbrum: I'd prefer to have this in a helper method so that we use the same logic here and in line 156 where we set `artifacts_path`

app/models/bulk_export.rb
```diff
@@ -137,33 +142,46 @@ def zip_file_name
   end
 
   def export_to_zip
-    works_scope = Work.includes(pages: [:notes, { page_versions: :user }])
```
benwbrum, line 140: I worry about losing the `includes`, since we probably added that for performance reasons.  How does this perform on a work with lots of activity, compared to the same code in `development`?

app/models/bulk_export.rb
```diff
@@ -137,33 +142,46 @@ def zip_file_name
...
-      Zip::OutputStream.open(tmpfile.path) do |out|
-        write_work_exports(works, out, self.user, self)
-      end
+    write_work_exports(works, artifacts_path, self)
+
+    zip_path = "tmp/fromthepage_exports/export_#{self.id}.zip"
+    FileUtils.rm_f(zip_path)
 
-      tmpfile.rewind
+    Zip::File.open(zip_path, create: true) do |zip|
```
benwbrum: I would be interested in comparing the memory footprint of this code against the old code in `development`

benwbrum (changes requested):
One minor change to remove duplication, plus a lot of questions.

app/models/bulk_export.rb
WillNigel23: I lost the includes here to lower memory cost. Includes preloads everything. I think in this case we are fine with n+1 since this is technically async anyway

benwbrum: Oh, that's an interesting point.  Yes, optimizing for memory is more important than optimizing for speed on these bulk exports.

benwbrum:
I've tested this with a handful of configurations (excluding PDF or MS-Word) and ran into this error, which seems related to the refactor:
```
Collection: alt-text test

Zip file: tmp/fromthepage_exports/export_6879.zip

Status: error

Configuration:
Plaintext Verbatim Page
Plaintext Verbatim Work
Plaintext Emended Page
Plaintext Emended Work
Plaintext Searchable Page
Plaintext Searchable Work
Tei Work
Html Page
Html Work
Subject Csv Collection
Table Csv Collection
Table Csv Work
Created At
Updated At
Work Metadata Csv
Static
Subject Details Csv Collection
Organization
Use Uploaded Filename
Report Arguments
Notes Csv
Page Details Csv Work
Page Details Csv Collection

Zip File
[Download](http://localhost:3000/bulk_export/6879/download)

Processing Log
# Logfile created on 2026-05-13 13:23:18 -0500 by logger.rb/v1.7.0
Found bulk_export for 
	user=benwbrum
	from collection=alt-text test
{"status"=>"queued",
 "id"=>6879,
 "user_id"=>2,
 "collection_id"=>32000342,
 "plaintext_verbatim_page"=>true,
 "plaintext_verbatim_work"=>true,
 "plaintext_emended_page"=>true,
 "plaintext_emended_work"=>true,
 "plaintext_searchable_page"=>true,
 "plaintext_searchable_work"=>true,
 "tei_work"=>true,
 "html_page"=>true,
 "html_work"=>true,
 "subject_csv_collection"=>true,
 "table_csv_collection"=>true,
 "table_csv_work"=>true,
 "created_at"=>Wed, 13 May 2026 18:23:11.000000000 UTC +00:00,
 "updated_at"=>Wed, 13 May 2026 18:23:11.000000000 UTC +00:00,
 "work_metadata_csv"=>true,
 "work_id"=>nil,
 "facing_edition_work"=>false,
 "text_pdf_work"=>false,
 "text_docx_work"=>false,
 "static"=>true,
 "document_set_id"=>nil,
 "subject_details_csv_collection"=>true,
 "text_only_pdf_work"=>false,
 "organization"=>"by_work",
 "use_uploaded_filename"=>true,
 "plaintext_verbatim_zero_index_page"=>false,
 "owner_mailing_list"=>nil,
 "owner_detailed_activity"=>nil,
 "collection_activity"=>nil,
 "collection_contributors"=>nil,
 "report_arguments"=>{"preserve_linebreaks"=>"0"},
 "notes_csv"=>true,
 "admin_searches"=>nil,
 "page_details_csv_work"=>true,
 "page_details_csv_collection"=>true}

	2026-05-13T13:23:20-05:00 Exporting work	32033032	archives

	[MEM] START WORK 32033032 RSS=225.5MB

/home/benwbrum/dev/products/fromthepage/fromthepage/app/controllers/concerns/export_service.rb:488:in `export_plaintext_translation_pages': undefined method `supports_translation?' for nil (NoMethodError)

    if work.supports_translation?
           ^^^^^^^^^^^^^^^^^^^^^^
	from /home/benwbrum/dev/products/fromthepage/fromthepage/app/helpers/export_helper.rb:210:in `block (2 levels) in write_work_exports'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:82:in `each'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:82:in `block in find_each'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:158:in `block in find_in_batches'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:364:in `block in batch_on_loaded_relation'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:360:in `each'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:360:in `each_slice'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:360:in `batch_on_loaded_relation'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:260:in `in_batches'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:157:in `find_in_batches'
	from /home/benwbrum/.rvm/gems/ruby-3.3.5/gems/activerecord-7.2.3.1/lib/active_record/relation/batches.rb:81:in `find_each'
	from /home/benwbrum/dev/products/fromthepage/fromthepage/app/helpers/export_helper.rb:206:in `with_index'
	from /home/benwbrum/dev/products/fromthepage/fromthepage/app/helpers/export_helper.rb:206:in `block in write_work_exports'
```

WillNigel23:
Nice catch! One fix I did was remove the re-declaration of `@work` variable so that it can be accessed by the sub functions called. We already should be passing either the `work` or `page` variable, this is unnecessary. It will only cause memory leak in iterations. I just missed this one place for accessing work thru the page.

@benwbrum

PR #5525, opened 2026-05-18
5522 - New page image uploads should upload to active storage

### LINKED ISSUE #5522: New page image uploads should upload to active_storage
This is the first step.

All new uploads should go thru active_storage.

WillNigel23:


app/models/page.rb
```diff
@@ -770,6 +785,7 @@ def modernize_absolute(filename)
     end
   end
 
+  # TODO: Deprecate once move to active storage is completed
```
benwbrum, line 836: How does this implementation handle thumbnails?  Are they still stored locally and generated from uploaded images?

app/models/page.rb
```diff
@@ -312,34 +315,6 @@ def canonical_facsimile_url
...
-        self.ia_leaf.page_w
-      else
-        nil
-      end
-    else
-      self[:base_width]
-    end
-  end
-
   def base_image
```
benwbrum, line 320: Do we still need this if we're getting rid of `base_image` as an attribute?

app/models/page.rb
```diff
@@ -357,14 +332,17 @@ def scaled_image(factor = 2)
...
-      return nil
-    end
-    if self.base_image.blank?
-      return nil
+    return if self.ia_leaf
+    return if self.base_image.blank?
+
+    if image.attached?
+      return Rails.application.routes.url_helpers.url_for(
+        image.variant(resize_to_limit: [nil, 400])
```
benwbrum: Can you point me to some docs on `variant`?  I'm thinking about other things which we might be able to use it for.

app/interactors/page/lib/common.rb
```diff
@@ -6,29 +6,20 @@ def process_uploaded_file(image_file)
...
     context.errors = e.message
     context.fail!
   end
 
+  # TODO: Deprecate this once move to active_storage is complete
   def assign_dimensions
-    File.unlink @page.thumbnail_filename if File.exist?(@page.thumbnail_filename)
-
     image = Magick::ImageList.new(@page.base_image)
     @page.base_width = image.columns
```
benwbrum: Is this going to mess up the new calculations for image height and width, or will this not be called for attached images?

app/interactors/page/rotate.rb
```diff
@@ -12,9 +12,65 @@ def initialize(page:, orientation:)
...
   def perform
-    0.upto(@page.shrink_factor) do |i|
-      rotate_file(@page.scaled_image(i), @orientation)
+    ensure_active_storage_image!
+
+    filename = @page.image.filename.to_s
+    content_type = @page.image.content_type
+
+    downloaded = @page.image.download
+    image = Magick::Image.from_blob(downloaded).first
```
benwbrum, line 21: Is there any cleanup code after the rotation finishes, or will the download stay on the filesystem forever?

app/models/page.rb
WillNigel23: For now I'm keeping this since we are still back-supporting base_image. We can probably remove this after part 3 of the PR where we fully deprecate the implementation.

app/models/page.rb
WillNigel23: Here is the link:
https://api.rubyonrails.org/classes/ActiveStorage/Variant.html

After re-reading this part I am adding `processed` to this variant call so that it checks that the variant has been processed and uploaded to the service.

> When you do want to actually produce the variant needed, call processed. This will check that the variant has already been processed and uploaded to the service, and, if so, just return that. Otherwise it will perform the transformations, upload the variant to the service, and return itself again. Example:

> avatar.variant(resize_to_limit: [100, 100]).processed.url

app/models/page.rb
WillNigel23: Based on docs, whenever we call `variant(...).processed` it will upload the generated variant back to storage_service and cache it for future uses. It will not be stored locally

app/interactors/page/rotate.rb
WillNigel23: `download` here is the binary string format of a file, so it lives in-memory.

app/interactors/page/lib/common.rb
WillNigel23: Actually, I'm removing this function now.

I originally kept it because I did not touch `rotate` logic at the start. But I realized, if people wanted to rotate their image when using the old `base_image` implementation, then we might as well use that step to also upload to active_storage.

I added that to rotate, and forgot to remove this old assign_dimensions function

PR #5531, opened 2026-05-19
5523 - Migrate old page image implementation to use active storage

### LINKED ISSUE #5523: Migrate old page image implementation to use ActiveStorage
Second step.

We will run a script that will migrate old implementation to ActiveStorage.

We should do this as a oneoff rake task, we log which page id we are in and the last id we stopped at so we know where to continue in case the task is suspended.

WillNigel23:


benwbrum (changes requested):
Two comments:
* It would be a bit easier to test this if we had rake tasks to migrate a single page, work or colleciton.
* I'm not sure that this handles the pages that don't have uploaded images, like the ones resulting from a IIIF import/CONTENTdm or Internet Archive import.

PR #5539, opened 2026-05-21
DRAFT: Add field-based AI transcription support

saracarl:
For field-based collections, the Gemini rake task now builds a structured JSON prompt listing each transcription field by ID, label, and data type (including select options and spreadsheet column layouts). The response is validated for correct JSON, required field keys, and select option values before being stored in a new `transcription_json` column on ai_transcriptions. Validation failures store the raw response in `source_text` and mark status as error so the job error path handles them normally.

The AI tab display renders field values using the same label/value HTML format as the overview tab, with the human transcription (xml_text) shown alongside AI values when the comparison toggle is used. CER/WER statistics now work for field-based pages by deriving AI plain text from transcription_json when source_text is blank.


benwbrum (changes requested):
In the case of field-based projects, the AI accuracy calculations should not include the field labels.  Currently, [the code](https://github.com/benwbrum/fromthepage/blob/development/app/models/concerns/ai_accuracy_calculator.rb#L32) just uses the plaintext version of the page transcription and the plaintext version of the the ai_transcription which is generated from the same label/value pairs that we render.  We should instead try to only pull the values from the fields transcribed.  

It would be a reasonable implementation to have new methods like `Page#ground_truth_for_comparison` and `AiTranscription#text_for_comparison` which use the current plaintext code for document-based projects, but--for field-based projects--only take the field values and concatenate them with whitespace.

benwbrum (approved):

PR #5557, opened 2026-05-26
5500 error redirecting to assign categories for huge number of subjects

WillNigel23:
The initial idea is to use `session` to store the article_ids. But I think this is going to be flaky and unreliable, since there is still a 4kb limit to what we can store in a session key.

I added support for solid_cache, a db-backed caching. I know we are having storage issues recently but this should not take much, especially since we can also add `expiration` to what we cache.

So far this will only be explicitly used for assign_categories. We should be fine, I added 30m expiration to each key.

@benwbrum @saracarl 

Specific test cases to cover these will follow, preferably here https://github.com/benwbrum/fromthepage/pull/4978 after I catch it up to this PR.

### DIFF
--- a/app/controllers/transcribe_controller.rb
+++ b/app/controllers/transcribe_controller.rb
@@ -180,8 +180,23 @@ class TranscribeController  < ApplicationController
 
           next_page_id = @page.last? || save_button_clicked ? @page.id : @page.lower_item.id
           flash[:notice] = t('.saved_and_next_notice') if next_page_id != @page.id
+
+          cache_key = [
+            'assign_categories',
+            'save_transcription',
+            current_user.id,
+            @page.id,
+            SecureRandom.hex(8)
+          ].join(':')
+
+          Rails.cache.write(
+            cache_key,
+            old_article_ids,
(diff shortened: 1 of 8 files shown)


app/controllers/transcribe_controller.rb
```diff
@@ -321,8 +337,22 @@ def save_translation
...
 
+        cache_key = [
+          'assign_categories',
+          'save_translation',
+          current_user.id,
+          @page.id,
+          SecureRandom.hex(8)
+        ].join(':')
+
+        Rails.cache.write(
```
benwbrum, line 348: This is fascinating.  Are you using caches instead of sessionshere?

benwbrum (commented):
This looks OK, but seems like a big infrastructure change for this bug.  Will we be using the same cache code for other things too?

app/controllers/transcribe_controller.rb
WillNigel23: Yes. I would prefer this over `sessions`, although it is an architecture change. I don't think it is a huge one though

PR #5563, opened 2026-06-01
5405 Batch AI generation

### LINKED ISSUE #5405: Address starvation and user experience for project owners running AI Drafts
Currently, when a project owner requests that FromThePage create AI Drafts using Google Gemini, we create `ai_transcription` records for every page in the work or collection, and submit jobs to SolidQueue to process.  Since there is a single queue, which runs in a FIFO manner, this leads to starvation problems.  One project owner may run AI Draft on a collection containing 20,000 pages, for example, which prevents anyone else from seeing results until the whole list is done.

How can we address the following issues?
* We want our onboarding experience for trial owners to be a good one, and we want them to see quick results.  Having their AI Transcription jobs queued behind tens of thousands of others means that they may need to wait for days or weeks before they see results.
* We want our existing users to see progress on the AI Draft jobs they submit.  Right now, some of them are emailing us frustrated because they aren't seeing any results after waiting hours or days.
* We're not sure that we need to restrict our use of the Gemini API to one request (or a few requests) at any time, which is our current architecture.  If we could parallelize processing and run multiple requests at a time, that might improve our system-wide throughput running AI Drafts.

I posted this to Codex, and while [its suggestions](https://chatgpt.com/s/cd_69cfc1d74970819192e6a4addcf11838) were not very compelling, it did point out that we set `ai_transcription.status` to `processing` before submitting a page to Gemini, which may lead project owners to see items marked as "processing" in the UI that are, in fact, waiting to be sent to Gemini.

WillNigel23:
Closes #5405


benwbrum:
I'm not sure whether this code supports non-batch Gemini calls for some pages as well as batched Gemini calls for others.  Eventually we will want to run Gemini quickly for things we want to be responsive, but batch requests for large situations like "Institution runs Gemini over their whole collection, even though much of it is already transcribed".

PR #5593, opened 2026-06-18
5357 store wer cer on ai transcriptions

### LINKED ISSUE #5357: AI tab rendering slowly
In some cases, the AI tab is rendering very slowly.  See [this example](https://fromthepage.com/ohiouniversitylibraries/marie-deffenbaugh-scrapbook-1921-1923/marie-deffenbaugh-scrapbook-1921-1923/ai_text/35464017?pp=enable)

WillNigel23:


benwbrum:
Reviewing the model and other calculations, it looks like this PR stores the *final* CER/WER ratios on the `ai_transcription` model.  This solves the performance problem, but will prevent us from being able to calculate aggregated CER/WER statistics across multiple pages.

For example, if Page 1 has 1000 words of ground truth and the AI transcription is 100% accurate, the WER we store will be `100%`.  If Page 2 has 10 words of ground truth and the AI transcription is 0% accurate, we'll store `0%`.  If I want to calculate the performance of an engine on the work containing both pages, all I can do is take the mean of the two, yielding an incorrect `50%`.

I believe that we should be storing the inputs to these ratios, rather than the ratios themselves.  If we stored `word_length` and `levenshtein distance` for both verbatim and text variants, we'd be able to get a true statistic across the pages of the work: `(levenshtein_page1+levenshtein_page2)/(word_count_page1+word_count_page2)` => `(0+10)/(1000+10)`=>`1%` which is correct.

benwbrum:
We also need to retrospectively calculate the numbers to populate existing pages.  Even though it'll take a long time, I'm fine doing this in the migration script.

benwbrum:
I've tested this and am a little confused.  When I deploy the branch and run the migration, all cer/wer and distance/length fields are `nil`.  Once I visit an AI tab, the cer/wer fields are calculated and cached but the length/distance fields are still `nil`.  (And the speed of rendering is excellent!)

If I run the rake task, the cer/wer and length/distance fields are populated.

If I edit a page, it nils out the CER/WER and length/distance fields.  

After visiting the AI tab or waiting for a job, I see CER/WER but not distance/length:
```
3.3.5 :010 > page.ai_transcription
 => 
#<AiTranscription:0x0000728d99893d40
 id: 230095,
 page_id: 35464017,
 source_text: "American League Baseball.\n                        ...",
 prompt: nil,
 model: "gemini-3-pro-preview",
 reasoning: "**Defining the Scope**\n\nI'm zeroing in on the tran...",
 metadata: {},
 created_at: "2026-01-09 16:18:55.421226000 +0000",
 updated_at: "2026-07-08 11:13:27.840304000 +0000",
 status: "finished",
 transcription_json: nil,
 verbatim_cer: 77,
 verbatim_cer_distance: nil,
 verbatim_cer_length: nil,
 verbatim_wer: 62,
 verbatim_wer_distance: nil,
 verbatim_wer_length: nil,
 text_cer: 56,
 text_cer_distance: nil,
 text_cer_length: nil,
 text_wer: 58,
 text_wer_distance: nil,
 text_wer_length: nil,
 verbatim_non_stopword_accuracy: nil> 
```

WillNigel23:
Okay I understand now what's happening and why! A few things here @benwbrum 

1. This feature hands off the calculation async, so we will need a running solid_queue worker to process the jobs. To make it easier you can just turn off solid_queue in `development.rb` and hand it off to the default in-memory job processor like before.

2. I added a fallback, in case job processing is slow and users want to see the calculations as soon as possible. I implemented a lazy-load fetching of the calculations which caches it as well. So when users go to ai tab and the calculations are not yet ready, we compute and cache it lazily so it gets displayed asap. This is why you saw the calculations cached after viewing the page, but `nil` when the page was edited (since we have to nullify the values before launching the async job)

3. As per the distance/length values being nil, this is a bug since I forgot to add the calculations in the `on-demand` version of calculation, which I now patched.

4. For the rake task, I'm not sure what's the problem. I ran it now and it worked.

PR #5618, opened 2026-06-25
Redirect Riiif image-service root requests to IIIF info endpoint and add spec

benwbrum:
### Motivation
- Ensure bare image service requests (e.g. `/image-service/:id`) are forwarded to the IIIF `info.json` endpoint for compatibility with consumers.
- Apply the redirect at the controller layer so the existing Riiif service routes resolve to the canonical IIIF info URL.

### Description
- Added a patch module `RiiifImagesControllerRedirectPatch` that implements `redirect` to `redirect_to "/image-service/#{ERB::Util.url_encode(params[:id].to_s)}/info.json"`.
- Prepend the patch into `Riiif::ImagesController` inside `ActiveSupport::Reloader.to_prepare` with a guard `next if Riiif::ImagesController < RiiifImagesControllerRedirectPatch` to avoid double-prepending.
- The initializer continues to require the custom file resolver and set `Riiif::Image.file_resolver = Riiif::FromThePageFileResolver.new` on reloads.
- Added a request spec at `spec/requests/riiif_redirect_spec.rb` that asserts a GET to `/image-service/32407848` redirects to `/image-service/32407848/info.json`.

### Testing
- Added and ran the new request spec `spec/requests/riiif_redirect_spec.rb` using `bundle exec rspec spec/requests/riiif_redirect_spec.rb`. The test passed.
- No other automated tests were modified.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bf3f85a108322af19471b1e9621c9)

### DIFF
--- a/config/initializers/riiif.rb
+++ b/config/initializers/riiif.rb
@@ -4,3 +4,15 @@ require 'iiif/from_the_page_file_resolver'
 ActiveSupport::Reloader.to_prepare do
   Riiif::Image.file_resolver = Riiif::FromThePageFileResolver.new
 end
+
+module RiiifImagesControllerRedirectPatch
+  def redirect
+    redirect_to "/image-service/#{ERB::Util.url_encode(params[:id].to_s)}/info.json"
+  end
+end
+
+ActiveSupport::Reloader.to_prepare do
+  next if Riiif::ImagesController < RiiifImagesControllerRedirectPatch
+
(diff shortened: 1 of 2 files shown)


benwbrum:
To reproduce, dereference the canvas IDs in the manifest (like curl -i http://localhost:3000/image-service/32407848).  This isn't something that IIIF servers normally support, nor that IIIF clients do, but crawlers hitting every URL they see probably are causing this.

PR #5646, opened 2026-06-30
Parallel tests

WillNigel23:


WillNigel23:
This is ready @benwbrum @saracarl 

There is a combined coverage down below the tests, so we should ignore the individual coveralls per matrix

PR #5658, opened 2026-07-05
Fix LaTeX table footer for PDF export

benwbrum:
### Motivation
- LuaLaTeX was failing when compiling generated tables because an empty `\noalign{}` was emitted after `\bottomrule`, causing a `Missing number, treated as zero` fatal error at `\end{xltabular}`.

### Description
- Remove the stray `\noalign{}` emitted after `\bottomrule` in `app/interactors/work/export/lib/utils.rb` so table footers end cleanly before `\end{xltabular}`.
- Add a regression spec `spec/interactors/work/export/lib/utils_spec.rb` that asserts table output contains `\bottomrule\n\end{xltabular}` and does not include `\bottomrule\noalign{}` for the problematic table case.

### Testing
- Run syntax checks with `ruby -c` on both `app/interactors/work/export/lib/utils.rb` and the modified spec, which returned `Syntax OK` for both files.
- Run `git diff --check` which reported no whitespace or diff errors for the change.
- Attempted to run the spec runner via `bundle exec`, but execution was blocked by a Ruby version mismatch in this environment (active Ruby `3.4.4` vs Gemfile-specified `3.3.5`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4a6afee61c8322a280a302fde9abd6)

### DIFF
--- a/app/interactors/work/export/lib/utils.rb
+++ b/app/interactors/work/export/lib/utils.rb
@@ -251,7 +251,7 @@ class Work::Export::Lib::Utils
       end
     end
 
-    latex += "\\bottomrule\\noalign{}\n"
+    latex += "\\bottomrule\n"
     latex += "\\end{xltabular}\n"
 
     latex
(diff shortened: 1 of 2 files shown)


benwbrum:
This caused enough failures that we are moving this to @WillNigel23

PR #5668, opened 2026-07-06
5607 - OSD viewer expands to cover transcription screen in image-on-top view

### LINKED ISSUE #5607: OSD viewer expands to cover transcription screen in image-on-top view
Doug at CWF showed us a page in a transcription view in which the image viewer had expanded to several screen-heights tall, completely covering the CodeMirror editor.   He seemed unable to grab the slider control to adjust it as well.

Here is a screenshot of what he saw, scrolled down on the page:

Image

WillNigel23:


app/assets/javascripts/ftp_interface.js
```diff
@@ -246,8 +246,10 @@ const ResizableSplitter = {
         ? parseFloat(position) / 100 * totalHeight
         : parseFloat(position);
 
-      const heightPanel1 = initialHeight;
-      const heightPanel2 = totalHeight - initialHeight;
+      const maxInitialHeight = window.innerHeight * 0.8;
+
+      const heightPanel1 = Math.min(initialHeight, maxInitialHeight);;
+      const heightPanel2 = Math.min((totalHeight - initialHeight), maxInitialHeight);
```
WillNigel23, line 252: The gist is, we seem to decide the splitter height/offset  via the height of the text area. This means when we have an extremely tall codemirror height then the splitter becomes unmanageable. As per my test this happens both to top/bottom layouts.

What I did is cap the initial height calculation to 80% of max screen height. Whichever is smaller is what we will use. As per my testing, this is more usable than the current where the page splitter controller is stuck way below and completely blocks the editor.

benwbrum:
This fixes the problem in both regular transcribe mode and fullscreen mode.  However, it introduces a problem in fullscreen, that the transcription conventions are no longer visible under the CodeMirror editor.

### Regular
image


### Fullscreen
image

PR #5678, opened 2026-07-09
Add AI draft button for field-based transcription

saracarl:
Shows an "AI Draft" button on the transcribe tab when a page has a finished AI transcription and its field-based collection has no spreadsheet-type field. Clicking it populates the field inputs from finished_ai_transcription.transcription_json, mirroring the existing plain-text AI Draft flow (reuses the ai_draft_used hidden field so deed recording and the flag-ai-use job keep working unchanged).

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -272,6 +272,12 @@ class Page < ApplicationRecord
     self.collection.field_based
   end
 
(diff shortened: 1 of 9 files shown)


benwbrum (commented):
This has been reviewed.

PR #5688, opened 2026-07-13
Search collection users by display name

### LINKED ISSUE #5687: collection#search_users does not search display_name field
Eric at LVA reports that a user wanted to be added to a document set, but that he couldn't find them in the system.  He only had access to their user profile, listing their handle `jsmith1776` and that user did not appear in the Add Collaborator options.  On further research, the user account had the `display_name` of `jsmith1776`, the email address of `john.smith@google.com` and the real name of `John Smith`.  

The search users method does not look at display name, so project owners may miss the only thing that they can see.

We should add display_name to the criteria in https://github.com/benwbrum/fromthepage/blob/9206e242fd704c5edf09a4600e59fce11bbda8c4/app/controllers/collection_controller.rb#L72-L75

benwbrum:
### Motivation
- Collection owners could not find users when only the `display_name` (profile handle) matched the search term, so include `display_name` in the search criteria to match what owners see on user profiles.

### Description
- Update `CollectionController#search_users` to include `LOWER(display_name) LIKE :search` in the query (`app/controllers/collection_controller.rb`).
- Add a request spec that verifies a user with only a matching `display_name` (`jsmith1776`) is returned by the search (`spec/requests/collection_controller_spec.rb`).

### Testing
- Static syntax checks passed with `ruby -c` for the modified files.
- Running `rspec spec/requests/collection_controller_spec.rb` was attempted but failed due to the test environment lacking a MySQL connection (`Can't connect to MySQL server on '127.0.0.1:3306'`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a55561bf18c83228b973cd45a289e3c)

### DIFF
--- a/app/controllers/collection_controller.rb
+++ b/app/controllers/collection_controller.rb
@@ -69,7 +69,10 @@ class CollectionController < ApplicationController
                      collaborator_ids
     end
 
-    users = User.where('LOWER(real_name) LIKE :search OR LOWER(email) LIKE :search', search: query)
+    users = User.where(
+      'LOWER(real_name) LIKE :search OR LOWER(email) LIKE :search OR LOWER(display_name) LIKE :search',
+      search: query
+    )
                 .where.not(id: excluded_ids)
                 .where.not(id: @collection.owner.id)
                 .limit(100)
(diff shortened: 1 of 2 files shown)


benwbrum:
This fixes the reported problem and may be merged once tests pass.

Issue #3620, opened 2023-04-24
marking pages inactive doesn't affect page counts


saracarl:
Sonya at LVA is worried that when she marks collections as inactive, that doesn't change her page counts (I'm guessing on her dashboard or owner summary). 

Let's compare this to the count in the admin dashboard.

@anabeabas could you look at the admin dashboard and screenshot it, then log in as LVA and screenshot their dashboard counts and see if they are different.  Include the URLs in the screenshots.

anabeabas:
@saracarl 

Total page count from admin dashboard:
![image](https://user-images.githubusercontent.com/98731543/234376765-3b9e593f-d3e2-45ff-b2a1-d2877f82a875.png)

Total page count from LVA's dashboard:
![image](https://user-images.githubusercontent.com/98731543/234376907-ff11c6db-17fb-492a-ad16-fc0b2c084f1e.png)

saracarl:
We need to watch the summary line (right under the "Owner Dashboard" in the first screenshot above) and the "Summary" tab statistics.  Both should include read only or inactive page counts.  I think the label changes depending on whether it's a trial acount -- "inactive" or an active account "read only".

benwbrum:
@saracarl to define more crisply and possibly assign to copilot.

PR #5697, opened 2026-07-19
Solution (#3620): marking pages inactive doesn't affect page counts

TFGSUMIT:
This pull request fixes the issue of marking pages as inactive not affecting page counts. To test this change, run `rails db:migrate` and then `rails spec` to ensure all specs pass. Additionally, run `rake update_page_counts:update` to update page counts for all pages in the database. This change is essential for maintaining accurate page counts, which is crucial for various use cases in the application.


benwbrum:
This does not actually fix #3620, in no small part due to a misleading title on that issue.

PR #5718, opened 2026-07-22
5703 - Off button for AI drafts

### LINKED ISSUE #5703: "Off Button" for AI Drafts
Currently, if a document has AI Drafts created, they will be available to users, with no way to turn them off.  We should add a collection-level configuration option that disables or suppresses the AI Draft functionality for transcribers, hiding the AI Draft button and the AI tab, even if a finished `ai_transcription` is available.  The AI Draft button and AI tab should remain visible to project owners, even if the draft functionality is disabled.

Configuration for this boolean should be in the collection settings, on the AI tab.  It should only be visible if at least one ai_transcription has been created for the collection.

WillNigel23:


benwbrum:
The "Disable AI Drafts" option should only appear on collections in which there actually are AI Drafts to disable; we should hide this if there are no `ai_transcription` records in a `finished` state.

PR #5743, opened 2026-07-28
Sanitize credentials in AI provider errors

saracarl:
### Motivation

- Prevent AI-provider credentials from being persisted to metadata, written to logs, or displayed in historical records by redacting common credential query params. 
- Ensure sanitization is hostname- and position-independent so secrets in any URL/query position are removed while preserving useful context like HTTP status and redacted URL. 
- Apply defense-in-depth so both runtime logging and persisted/presented error text are protected.

### Description

- Add a centralized sanitizer `AiTranscription::Lib::ErrorMessageSanitizer` that redacts `key`, `api_key`, `token`, and `access_token` query parameters wherever they appear. 
- Use the sanitizer in `AiTranscription::GenerateJob#store_error!` to sanitize messages before assigning `metadata['error_message']`. 
- Sanitize error messages before logging in the Gemini handler `AiTranscription::Lib::Gemini::TranscribeHandler` so credentials are not written to application logs. 
- Add defense-in-depth in `AiTranscription#error_message` to sanitize historical error metadata before presentation and add/modify specs covering job storage, handler logging, and model presentation to assert secrets are removed but status and redacted URL remain useful.

### Testing

- Ran `ruby -c` syntax checks on the changed Ruby files and specs which all reported `Syntax OK`.
- Executed a standalone Ruby check that called `AiTranscription::Lib::ErrorMessageSanitizer.sanitize(...)` to verify `key`, `api_key`, `token`, and `access_token` are redacted and it passed. 
- Ran `git diff --check` which passed without issues. 
- Attempted to run `bundle exec rspec ...` for the new/updated specs but the environment does not have the `rspec` executable available from the bundle, so the full RSpec suite could not be executed in this run (environment warning).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a68c26a5ad48328949bc33ca4cb3f3d)

### DIFF
--- a/app/interactors/ai_transcription/lib/error_message_sanitizer.rb
+++ b/app/interactors/ai_transcription/lib/error_message_sanitizer.rb
@@ -1,0 +1,9 @@
+class AiTranscription::Lib::ErrorMessageSanitizer
+  CREDENTIAL_PARAMETER = /([?&](?:key|api_key|token|access_token)=)[^&#\s]+/i
+
+  def self.sanitize(message)
+    return if message.nil?
+
+    message.to_s.gsub(CREDENTIAL_PARAMETER, '\\1[FILTERED]')
+  end
+end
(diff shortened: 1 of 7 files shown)


saracarl:
Code reviewed.  Looks good to me.

PR #5750, opened 2026-07-30
Lazy load deeds

WillNigel23:
### Places deeds lazy_loading is implemented

collection#show (This covers both collection and document_set)
dashboard#watchlist
dashboard#owner
dashboard#landing_page
dashboard#collection_list

### DIFF
--- a/app/views/collection/show.html.slim
+++ b/app/views/collection/show.html.slim
@@ -185,8 +185,8 @@
 
     -if notes_visible?(@collection, current_user)
       h3 =t('.recent_edits')
-      -cache [@collection.id, @collection.most_recent_deed_created_at, :collection_show] do
-        =deeds_for({ collection: @collection, limit: 20, types: DeedType.collection_edits, long_view: true })
+      turbo-frame id="lazy_deeds" src=collection_deeds_path(@collection.owner, @collection, types: DeedType.collection_edits, limit: 20, long_view: true) loading="lazy"
+        = render partial: 'shared/components/loading'
 
 =link_to(collection_new_mobile_user_path(collection_id: @collection.slug), data: { controller: 'litebox', litebox: { hash: 'new-mobile-user' }}, id: 'new_mobile_user_link', style: "display: none")
   span =""
--- a/app/views/dashboard/owner.html.slim
+++ b/app/views/dashboard/owner.html.slim
@@ -20,7 +20,7 @@
             -unless c.works.empty?
               -collection_stats(c)
               .collection-work_stats
-                span 
+                span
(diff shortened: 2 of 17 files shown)


app/views/collection/show.html.slim
```diff
@@ -185,8 +185,8 @@
 
     -if notes_visible?(@collection, current_user)
       h3 =t('.recent_edits')
-      -cache [@collection.id, @collection.most_recent_deed_created_at, :collection_show] do
```
benwbrum, line 188: I am concerned that we are removing deed caching from the `collection#show` view and replacing it with lazy loading.  The individual pages should at the same speed  (or even slightly faster) under this refactoring, but each request to `collection#show` will fire off a SQL hit to the `deeds` table.  For slow-running queries (which I think are often in older collections), this increases the overall load on the DB, as well as adding long-running requests to the Passenger queue.  Both of these can slow down the overall performance of the site and produce `Queue Full` errors.

app/views/dashboard/owner.html.slim
```diff
@@ -48,4 +48,5 @@
 
   aside.sidecol
     h3 =t('.your_activity')
-    =deeds_for({ :user_id => current_user.id, :limit => 10 })
\ No newline at end of file
+    turbo-frame id="lazy_deeds" src=dashboard_deeds_path(limit: 10) loading="lazy"
```
benwbrum, line 51: Will we also be lazy-loading deeds for non-owner user profile pages?  Routes like https://fromthepage.com/TypingTilly are not very heavily used by humans, but I'm 100% sure that they're hit by bots and search engines.  That's a legitimate use, so adding lazy loading might be appropriate?  That said, it's not a sidebar, and I'm not sure that there is any useful information on those pages except the deed list, so perhaps it makes more sense for this to be live.

benwbrum (changes requested):
This looks really good!  I have two big requests:
* Add lazy loading to `document_set#show` on the model of `collection#show`
* Make sure that we're caching the deeds for collections, or at least addressing my comments on `collection#show:188`.

app/views/collection/show.html.slim
WillNigel23: We are not removing deed caching. We moved the deed caching in the shared partial.

app/views/dashboard/owner.html.slim
WillNigel23: Perhaps, but I think there's still merit to lazy loading this. I think both SEO and users will find it more tolerable to see the page loaded with spinner than white blank page if in case loading takes long?

PR #5755, opened 2026-07-31
Fix field labels rendering as dashed slugs in read view and exports

saracarl:
## Summary
- Field-based transcription projects were showing field labels with dashes between every word in the "read work" view and all downstream exports (plaintext/PDF/DOCX), instead of the label as configured by the project owner.
- Root cause: `TranscriptionField::Lib::Utils.parse_fields` used the submitted form's hash key (`field.label.parameterize`, generated by the transcribe form for the input `name` attribute) as the display label written into `page.source_text`, instead of the actual `field.label`.
- Fix: use `field.label` (already resolved in scope) for the displayed label, matching the pattern already used correctly in `DisplayHelper#ai_field_html`.
- Bug was present since the feature's introduction (PR #4578), deployed to production 2025-11-19 13:06:59 CST -- so it existed from day one, not a later regression.
- Added `fromthepage:fix_field_label_dashes` rake task to backfill `xml_text` for already-transcribed pages saved on/after that deploy, so existing completed projects (like the customer who reported this) don't need every page resaved by hand. It rebuilds the label text from each field's current label plus the page's (already-correct) `transcription_json`, and writes only `xml_text` via `update_column` -- no `page_version` or other save callback is triggered. Defaults to a dry run; supports `work_id=`/`page_id=` filters for spot-checking before a full run.

Reported by a customer via email -- example label `"1. Please provide a listing of activities and programs which your organization sponsored during the past academic year. Be as specific as possible:"` was rendered as `"1-please-provide-a-listing-of-activities-and-programs-which-your-organization-sponsored-during-the-past-academic-year-be-as-specific-as-possible:"` in their exported transcript.

## Test plan
- [x] Added `spec/interactors/transcription_field/lib/utils_spec.rb` covering `parse_fields`, asserting the rendered `field__label` span uses the configured label rather than the parameterized form key.
- [x] Verified the new spec fails against the pre-fix code (reproducing the exact customer-reported dashed output) and passes with the fix applied.
- [x] Added `spec/tasks/fix_field_label_dashes_spec.rb` covering the backfill rake task: dry-run reports without writing, real run rewrites `xml_text` without touching `source_text` or creating a `page_version`, and pages updated before the bug's deploy date are left untouched.
- [x] Found and fixed `spec/features/field_based_spec.rb`, which had asserted the parameterized/dashed label (`"first-field: Field one"`) as the expected preview output -- that assertion was inadvertently locking in the bug; updated to expect the real configured label (`"First field: Field one"`). Full file (13 examples) passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


app/interactors/transcription_field/lib/utils.rb
```diff
@@ -62,12 +62,12 @@ def self.parse_fields(page:, field_cells:)
...
       else
-        field_data.each do |cell_key, cell_value|
+        field_data.each_value do |cell_value|
           # broken tags or actual < / > signs
           cell_value = ERB::Util.html_escape(cell_value) if cell_value.scan('<').count != cell_value.scan('>').count
 
-          cell_key = "#{cell_key}#{field.input_type == 'description' ? ' ' : ': '}"
-          source_text << "<span class=\"field__label\">#{cell_key}</span>#{cell_value}\n\n"
+          label = "#{ERB::Util.html_escape(field.label)}#{field.input_type == 'description' ? ' ' : ': '}"
+          source_text << "<span class=\"field__label\">#{label}</span>#{cell_value}\n\n"
```
benwbrum, line 70: This looks good to me, though we do need to make sure that this still works in spreadsheets..

benwbrum (approved):
This looks reasonable to me.

benwbrum:
I just tested running the rake task against a spreadsheet project on [this page](http://localhost:3000/riksarkivet/i-demokratins-namn/30002894a/display/32819792) using this command:
```
page_id=32819792 rake fromthepage:fix_field_label_dashes
```

This claimed to migrate this page, but the xml_text still has dashes in the field labels.

saracarl:
@benwbrum you left off the dry-run=false in your rake task.  It's working for me.

saracarl:
@copilot resolve the merge conflicts in this pull request

PR #5797, opened 2026-08-14
Harden unzip_file against path traversal/symlinks and add tests

benwbrum:
### Motivation
- Prevent unsafe ZIP extraction (path traversal, absolute paths, Windows paths, null bytes, and symlink attacks) when unzipping files for the new zoom feature and general ingestion.
- Ensure extraction is atomic per-archive validation so a malicious entry cannot leave a partially-extracted archive.

### Description
- Rewrote `ImageHelper.unzip_file` to `File.expand_path` the destination, validate entry names (null bytes, backslashes normalized to `/`, absolute paths, drive letters, and `..`) and ensure each entry resolves inside the destination before writing.
- Added pre-validation of all entries and a `prepare_zip_directory` private helper to create and validate directories without following symlinks, rejecting existing symlinks or non-directory paths.
- Write files using `File.open` with `File::WRONLY | File::CREAT | File::EXCL` and `File::NOFOLLOW` (when available) and set mode `0600`, and reject unsupported entry types.
- Added comprehensive specs in `spec/helpers/image_helper_spec.rb` to cover valid extraction, unsafe entries, symlink entries, and existing symlink-in-path behavior, and added a nested-ZIP integration spec in `spec/tasks/ingestor_spec.rb` to exercise recursive ZIP processing.

### Testing
- Ran the new helper specs with `bundle exec rspec spec/helpers/image_helper_spec.rb` and they passed.
- Ran the ingestor task spec with `bundle exec rspec spec/tasks/ingestor_spec.rb` and it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7f3dd1ea848322b389c61d40c0cb58)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -93,7 +93,4 @@ Rails.application.configure do
   # Add a custom hostname for uploads
   # config.upload_host = 'uploads'
   # config.hosts << "#{config.upload_host}.localhost"
-
-  config.active_job.queue_adapter = :solid_queue
-  config.solid_queue.logger = ActiveSupport::Logger.new(Rails.root.join('log/queue.log'))
 end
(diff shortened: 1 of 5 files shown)


benwbrum:
Wrong base branch.

PR #5870, opened 2026-09-09
5869 - Bots crawling iiif manifests causes queue full error

### LINKED ISSUE #5869: Bots crawling IIIF manifests cause queue full errors
The bots have discovered our IIIF manifests, and that is causing our users to run into Passenger Queue Full errors.  Part of the problem with this is the bots, but part of the problem is that we have never done any caching or performance tuning to our IIIF manifest generation.  If I try to fetch http://localhost:3000/iiif/32073305/manifest (a manifest for the inactive account Maryland State Archives), it hits the DB almost thirty thousand times:

`Completed 200 OK in 29820ms (Views: 17.9ms | ActiveRecord: 2837.6ms (28097 queries, 11234 cached) | GC: 4112.1ms)`

Sara is looking at ways to try to block bots from our iiif endpoint as a temporary stopgap.  I think that we should look at ways to cache IIIF manifests, since most of them change very rarely.

WillNigel23:


app/controllers/iiif_controller.rb
```diff
@@ -167,110 +175,17 @@ def manifest
...
-    if work.supports_translation?
-      seed = {
-              '@id' => url_for({ controller: 'iiif', id: work_id, action: 'layer', type: 'translation', only_path: false }),
-              'label' => 'translation layer'
-            }
-      layer = IIIF::Presentation::Layer.new(seed)
-      manifest['otherContent']  << layer
-    end
+    # TODO: Refactor to use cache concern when https://github.com/benwbrum/fromthepage/pull/5331 is merged
+    expires_in MANIFEST_EXPIRES_AT, public: true
```
WillNigel23, line 200: We haven't merged https://github.com/benwbrum/fromthepage/pull/5331, but we have bootstrapped the cache concern there so we don't have to write it this way. It's not ready yet, but this is just to note we have to change it later.

app/controllers/iiif_controller.rb
```diff
@@ -787,22 +699,19 @@ def iiif_collection_from_collection(collection, depth)
...
-      value << manifest_uri_from_ia(work.ia_work) if work.ia_work
-
-      if value.length == 1
-        dc_source['value'] = value.first
-      else
-        dc_source['value'] = value
-      end
-    end
-    dc_source
+    return unless work.identifier.present? || work.sc_manifest || work.ia_work
```
WillNigel23, line 722: Just some housekeeping refactor.

app/controllers/iiif_controller.rb
```diff
@@ -1082,33 +991,15 @@ def add_seeAlso_to_canvas(canvas, page)
...
-      annotation_list['label'] = 'Translation'
-      annotation_list['sc:forCanvas'] = canvas_id_from_page(page)
-    end
-  when 'notes'
-    unless page.notes.blank?   # no comments
-      annotation_list = IIIF::Presentation::AnnotationList.new
-      annotation_list['@id'] = url_for({ controller: 'iiif', action: 'list', page_id: page.id, annotation_type: type, only_path: false })
-      annotation_list['label'] = 'Notes'
-      annotation_list['sc:forCanvas'] = canvas_id_from_page(page)
+  def annotationlist_from_page(page, type)
```
WillNigel23, line 1014: Checking the diff, I believe there is a bug here (probably unreachable) where we return annotation_list even if the switch case does not satisfy anything. I said probably unreachable because the switch case seems to cover all cases (even if there is no catch-all else block). 

Even still, I think it's worth handling that just in case. Also simplified how this is written.

app/controllers/iiif_controller.rb
```diff
@@ -9,6 +9,14 @@ class IiifController < ApplicationController
   before_action :load_objects_from_ids
   before_action :check_api_access, except: [:collections, :contributions, :for, :collection_for_domain]
 
+  ANNOTATION_FIELDS = {
+    'transcription' => :source_text,
+    'translation'   => :source_translation,
+    'notes'         => :notes
+  }.freeze
+
+  MANIFEST_EXPIRES_AT = 1.day
```
WillNigel23, line 18: We can adjust this if we want.

benwbrum: Looks accurate to me.

app/controllers/iiif_controller.rb
```diff
@@ -1153,6 +1045,108 @@ def status_service_for_page(page)
     service
   end
 
+  def build_iiif_manifest(work)
```
benwbrum, line 1067: This looks so much cleaner!

app/controllers/iiif_controller.rb
benwbrum: Nice -- it's much cleaner!

app/controllers/iiif_controller.rb
```diff
@@ -167,110 +175,18 @@ def manifest
...
-
-    if work.supports_translation?
-      seed = {
-              '@id' => url_for({ controller: 'iiif', id: work_id, action: 'layer', type: 'translation', only_path: false }),
-              'label' => 'translation layer'
-            }
-      layer = IIIF::Presentation::Layer.new(seed)
-      manifest['otherContent']  << layer
-    end
+    if stale?(work)
```
benwbrum: I'd like to know more about how `stale?` is calculated.  This seems like the thing that checks cache invalidation?

benwbrum (commented):
Looks fantastic.  I just have some questions about cache invalidation.

app/controllers/iiif_controller.rb
```diff
@@ -178,10 +178,18 @@ def manifest
     return render status: :not_found, json: { error: 'Not found' } unless work
 
     # TODO: Refactor to use cache concern when https://github.com/benwbrum/fromthepage/pull/5331 is merged
+    latest_page_update = work.pages.map(&:updated_at).compact.max
```
benwbrum: I suspect that our performance would improve substantially if we use SQL to do this; probably by something like `work.pages.maximum(:updated_at)`

app/controllers/iiif_controller.rb
```diff
@@ -178,10 +178,18 @@ def manifest
...
+    cache_key = [
+      'iiif-manifest-v1',
```
benwbrum: At some point we will want to generate manifests for the IIIF Presentation API standard v3 and v4.  (Currently we only support v2).  I'd prefer to relabel this to something like `iiif-v2-manifest-v1` to clarify the difference between implementation version and API/standard version

app/controllers/iiif_controller.rb
```diff
@@ -178,10 +178,18 @@ def manifest
...
+      work.cache_key_with_version,
+      latest_page_update&.to_i,
+      work.pages.size
```
benwbrum: This is a very good catch.  I hadn't thought about what to do when a page is added or deleted from the work.

benwbrum (changes requested):
Just a minor performance suggestion and naming change.

benwbrum:
I'm running into this error in development:
```


ActiveRecord::UnknownAttributeReference - Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "COUNT(*)".This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql().:
  app/controllers/iiif_controller.rb:181:in `manifest'

```

It seems to be called by `COUNT(*)`

app/controllers/iiif_controller.rb
benwbrum: We need to add a count of `ai_transcriptions` for pages in the work.  It's fine with me if we modify the model to facilitate this via a `has_many :ai_transcriptions, through: :page` or something similar.

PR #5875, opened 2026-09-10
Restrict metadata editing in read-only projects

### LINKED ISSUE #5871: Non-owners can still edit/create metadata in read-only collections or document sets
GRI reports that a transcriber has been editing metadata in one of their collections which has been set to read-only.  It looks like when we added the read-only document sets and collections, we only restricted users from transcribing, not from accessing the metadata description flows.  We need to restrict users from creating or editing metadata in the UI and in routes/controllers for any access point if the work is in a read-only collection or document set, using the same logic we do for transcription.

benwbrum:
### Motivation

- Prevent non-owner/non-collaborator users from creating or editing work metadata when the work is located in a read-only `Collection` or `DocumentSet`, using the same authorization semantics as transcription.  

### Description

- Add a controller-level guard `authorized_to_describe?` and wire it into `WorkController` with `before_action` for `:describe` and `:save_description`, redirecting unauthorized users to the metadata overview with `metadata_overview_collection_work_path`.  
- Hide or conditionally render metadata entry links and actions in views when the current user cannot transcribe, replacing calls like `work.user_can_transcribe?(current_user)` with the transcription check `current_user&.can_transcribe?(work, collection)` where applicable (changes in `app/views/*` and `app/views/shared/_description_tabs.html.slim`).  
- Protect the post-segmentation "edit metadata" action so it only shows when the user can transcribe the new work.  
- Add request specs covering read-only `Collection` and read-only `DocumentSet` access attempts, including attempts to open the description editor and to `PATCH` `save_description`, and verify owner access remains allowed (`spec/requests/work_controller_spec.rb`).

### Testing

- Ran `git diff --check` to ensure no whitespace/merge issues and it passed.  
- Ran Ruby syntax checks with `ruby -c` on modified files: `app/controllers/work_controller.rb` and `spec/requests/work_controller_spec.rb`, both returned `Syntax OK`.  
- Attempted to run the new request specs with `bundle exec rspec spec/requests/work_controller_spec.rb` but the environment could not run them because the container Ruby is `3.4.4` while the project `Gemfile` requires `3.3.5` and the `rspec` executable is not available (bundler / Ruby mismatch), so the full spec run was not executed.  
- Manual inspection and local view logic checks were performed to ensure describe/save flows are redirected or hidden for unauthorized users.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aa2c543c3848322a114a5f3c6a36ee6)

### DIFF
--- a/app/controllers/work_controller.rb
+++ b/app/controllers/work_controller.rb
@@ -28,6 +28,7 @@ class WorkController < ApplicationController
     :remove_scribe,
     :search_scribes
   ]
+  before_action :authorized_to_describe?, only: [:describe, :save_description]
 
   # no layout if xhr request
   layout :dynamic_layout, only: [:new, :create, :configurable_printout, :edit_scribes, :remove_scribe]
@@ -399,6 +400,12 @@ class WorkController < ApplicationController
   end
 
   private
+
+  def authorized_to_describe?
+    return if user_signed_in? && current_user.can_transcribe?(@work, @collection)
+
+    redirect_to metadata_overview_collection_work_path(@collection.owner, @collection, @work)
+  end
 
   def require_segmentation_feature
     return if @collection&.segmentation_feature_enabled?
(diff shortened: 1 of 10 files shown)


benwbrum (commented):
This looks good.

PR #5881, opened 2026-09-11
Fix empty categories in TEI taxonomy exports

benwbrum:
### Motivation

- Prevent empty category shells from being emitted in TEI taxonomy output when a category only contains organization subjects that are emitted separately in `<listOrg>`.
- Ensure a parent category is retained only when it actually contains an eligible subject or when a child recursion returns non-empty TEI.
- Add request coverage for multiple organization-enabled branches and for retaining ancestors when a deeper descendant contains a non-organization subject.

### Description

- Change `ExportHelper#category_to_tei` to build each child’s serialized TEI first and append it only when non-empty, and set `has_content` based on that result instead of unconditionally when children exist (file: `app/helpers/export_helper.rb`).
- Extend `spec/requests/export_controller_spec.rb` to configure additional organization-enabled branches (Businesses/Companies and Military Units/Regiments), include an unrelated deep non-organization subject so `<encodingDesc>` is rendered, and wire those articles into the test page.
- Add assertions that organization records remain in `<listOrg>`, that organization-only category branches are omitted from the `<taxonomy>`, and that parent categories are retained when a deeper descendant contains a non-organization subject (file: `spec/requests/export_controller_spec.rb`).

### Testing

- Ran `ruby -c app/helpers/export_helper.rb` and it returned syntax OK.
- Ran `ruby -c spec/requests/export_controller_spec.rb` and it returned syntax OK.
- Ran `git diff --check` and no issues were reported.
- Attempted `bundle exec rspec spec/requests/export_controller_spec.rb` but the test runner could not be executed in this environment because the container Ruby version differs from the Gemfile requirement and the `rspec` executable is unavailable, so full specs were not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aa4616982c08322bc2f362528e12be0)

### DIFF
--- a/app/helpers/export_helper.rb
+++ b/app/helpers/export_helper.rb
@@ -464,8 +464,11 @@ module ExportHelper
       end
     end
     category.children.each do |child|
-      has_content = true
-      tei << category_to_tei(child, subjects, seen_subjects)
+      child_tei = category_to_tei(child, subjects, seen_subjects)
+      unless child_tei.empty?
+        has_content = true
+        tei << child_tei
+      end
     end
     tei << "</category>\n"
 
--- a/spec/requests/export_controller_spec.rb
+++ b/spec/requests/export_controller_spec.rb
@@ -370,9 +370,21 @@ describe ExportController do
     end
 
     context 'with organization articles in descendant categories' do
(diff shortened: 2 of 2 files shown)


benwbrum:
Obviated by #5885

